PR #143, opened 2012-10-02
Multi-core/cpu compiling

### LINKED ISSUE #139: Compile using multiple cores
It would be nice if node-gyp read something like a JOBS environment variable and if available, sets the "-j" (or Windows/OSX/etc equivalent) value to this variable value?

mscdex:
Set JOBS=n where n is a number or set JOBS=max and all available cores/cpus will be used.

Closes #139


Preen:
Why isnt `JOBS=max` default?

PR #366, opened 2013-12-12
Quick and dirty fix to allow WriteXmlIfChanged to write xml in utf-8, co...

jfhenriques:
This is a quick and dirty fix to allow node-gyp to generate the vcxproj when the are latin-1 characters in the paths.

I'm not a python programmer, this just solved the problem I had with latin-1 characters..

WriteXmlIfChanged was not actually writing in utf-8, but in ascii. Converting directly to utf-8 would fail if unicode(string, native_enconding) is not called first.


eTorAken:
Seems it is not fixed yet `node-gyp@1.0.3`

Sogl:
Not fixed yet =(

PR #368, opened 2013-12-14
run libtool command more explicitly

yorkie:
See below trace:
```
libtool: unrecognized option `-static'
libtool: Try `libtool --help' for more information.
```
I think we need a more explicit path for `libtool`, do you think so?
Indeed it should be a stuff in `gyp`, but what I do maybe helps some people like me do less confusing, why not?

### DIFF
--- a/gyp/pylib/gyp/generator/make.py
+++ b/gyp/pylib/gyp/generator/make.py
@@ -161,7 +161,7 @@ cmd_solink_module = $(LINK.$(TOOLSET)) -shared $(GYP_LDFLAGS) $(LDFLAGS.$(TOOLSE
 
 LINK_COMMANDS_MAC = """\
 quiet_cmd_alink = LIBTOOL-STATIC $@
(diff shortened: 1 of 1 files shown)


bnoordhuis:
@yorkie Maybe you can upload this to codereview.chromium.org?  I'll be happy to walk you through it.  The upstream repo is https://chromium.googlesource.com/external/gyp and you'll need to upload it with git-cl from https://chromium.googlesource.com/chromium/tools/depot_tools.git.

yorkie:
Sorry for the delay response, Ben, I will try it tomorrow, thank you :-)

saper:
@yorkie the problem here is not about wrong path to libtool (by the way, my libtool lives in `/usr/local/bin/libtool`). Are you still having this issue? The source of your problem is something else...

yorkie:
I do not have this problem by adding the libtool to PATH. Sorry and closing.

PR #379, opened 2014-01-02
Show better error message in case of bad network settings

robertkowalski:
Instead of just showing Error: getaddrinfo ENOTFOUND
this will show an error message to the user

Turns out the message is quite irritating to users, e.g:
https://github.com/voodootikigod/node-serialport/issues/288


robertkowalski:
yay, thank you! :)

PR #397, opened 2014-02-07
Import local gyp.py, then fallback to system

caruccio:
Some hosts have an old gyp version, which causes an ImportError.
It is better to import newer gyp.py first, then fallback to system-wide version.


TooTallNate:
This was fixed via PYTHONPATH in 52e8d9f7.

Issue #425, opened 2014-04-09
Migrate to Ninja


denzp:
[Ninja](http://martine.github.io/ninja/) is widely used in Chromium and associated projects. 
You can read some info at https://code.google.com/p/chromium/wiki/NinjaBuild

So, why not? :)
Only problem - is to build ninja if it isn't exists on host. I could investigate details and try to implement this

tk120404:
@denzp Can we close this issue as there is no activity on this?

denzp:
@tk120404 yes, sure. Unfortunately, there were no activity for years.

PR #429, opened 2014-04-15
Ninja

denzp:
Implementation of #425
On non-Windows Ninja is used by default. On Windows it can be used with flag `--ninja`

I have successfully used it by default in npm and builded popular modules on Linux64, Linux32, Windows.
Ninja is setting up in `install` command. For Windows we download a prebuild binary, for others we build it from sources.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -507,14 +507,32 @@ function install (gyp, argv, callback) {
       return done();
     }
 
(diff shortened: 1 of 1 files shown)


Sogl:
+1
good if that solves problems with modules installation on Windows

denzp:
Closing because of no activity here for years.

PR #450, opened 2014-06-03
Made it easier to build multiple platform binaries in Visual Studio

LaurentGoderre:


TooTallNate:
How does this make things _easier_ exactly?

LaurentGoderre:
It's been a while but those are Visual Studio target platforms. This creates one folder for x86 and one folder for x64 based on the compile target.

PR #473, opened 2014-07-10
Adds support for custom GYP include

anprogrammer:
When developers have libraries in non-standard locations (or don't wish
to dump them into System32 on Windows), they need to run "node-gyp
configure", edit the generated files to add include paths, and then run
"node-gyp install"

This is doubly painful when you wanted to "npm install" something, but
it depends on a library being in a default path on your system.

I've added the ability for users to include an additional ".gyp" file in
all node-gyp builds by setting an environment variable
(NODE_GYP_ADDITIONAL_CONFIG), which they can use to configure
include/library paths.  No longer must I manually install packages or
dump files into global include directories.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -9,6 +9,7 @@ var fs = require('graceful-fs')
   , glob = require('glob')
   , log = require('npmlog')
   , osenv = require('osenv')
+  , process = require('process')
   , which = require('which')
   , semver = require('semver')
   , mkdirp = require('mkdirp')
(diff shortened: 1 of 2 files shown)


bdunlay:
Has this been resolved in subsequent releases? I'm currently stuck trying to install packages that require libraries I've installed in custom locations.

TooTallNate:
The _preferred_ way to do this is to add a `common.gypi` file to the root of your module. For example: https://github.com/TooTallNate/node-vorbis/blob/master/common.gypi

Issue #501, opened 2014-09-11
error while node-gyp rebuild


yongjinkwon:
```
pi@raspberrypi ~ $ sudo node-gyp rebuild
gyp info it worked if it ends with ok
gyp info using node-gyp@1.0.2
gyp info using node@0.10.31 | linux | arm
gyp info spawn python
gyp info spawn args [ '/usr/local/lib/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'make',
gyp info spawn args   '-I',
gyp info spawn args   '/home/pi/build/config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/usr/local/lib/node_modules/node-gyp/addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/root/.node-gyp/0.10.31/common.gypi',
gyp info spawn args   '-Dlibrary=shared_library',
gyp info spawn args   '-Dvisibility=default',
gyp info spawn args   '-Dnode_root_dir=/root/.node-gyp/0.10.31',
gyp info spawn args   '-Dmodule_root_dir=/home/pi',
gyp info spawn args   '--depth=.',
gyp info spawn args   '--no-parallel',
gyp info spawn args   '--generator-output',
gyp info spawn args   'build',
gyp info spawn args   '-Goutput_dir=.' ]
gyp info spawn make
gyp info spawn args [ 'BUILDTYPE=Release', '-C', 'build' ]
make: Entering directory '/home/pi/build'
make: *** No rule to make target 'Release/obj.target/binding/src/binding.o', needed by 'Release/obj.target/binding.node'.  Stop.
make: Leaving directory '/home/pi/build'
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/node-gyp/lib/build.js:267:23)
gyp ERR! stack     at ChildProcess.emit (events.js:98:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:810:12)
gyp ERR! System Linux 3.12.22+
gyp ERR! command "node" "/usr/local/bin/node-gyp" "rebuild"
gyp ERR! cwd /home/pi
gyp ERR! node -v v0.10.31
gyp ERR! node-gyp -v v1.0.2
gyp ERR! not ok 
```

this is my error code.

[...]

Issue #485, opened 2014-08-03
PYTHONPATH makes rebuild fail


mstram:
action@nodeh-131169:~/git/dnode(master)$      pwd                                                                                         
     /home/action/git/dnode                 

     npm install weak

Why is node-gyp trying to build stuff in the PYTHONPATH?

I.e.   File "/home/action/git/SATK/tools/lang/parser.py"     has nothing to do with 'npm install weak'


> weak@0.3.3 install /home/action/git/dnode/node_modules/weak                                                                        
> node-gyp rebuild                                                                                                                   
                                                                                                                                     
Traceback (most recent call last):                                                                                                   
  File "/home/action/.parts/packages/nodejs/0.10.28/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py", line 15, in <module>
    import gyp                                                                                                                       
  File "/home/action/.parts/packages/nodejs/0.10.28/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 8, in
 <module>                                                                                                                            
    import gyp.input                                                                                                                 
  File "/home/action/.parts/packages/nodejs/0.10.28/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 5, in <m
odule>                                                                                                                               
    from compiler.ast import Const                                                                                                   
  File "/usr/lib/python2.7/compiler/__init__.py", line 29, in <module>                                                               
    from compiler.transformer import parse, parseFile                                                                                
  File "/usr/lib/python2.7/compiler/transformer.py", line 29, in <module>                                                            
    import parser                                    
                                                                                
  File "/home/action/git/SATK/tools/lang/parser.py", line 195                                                                        
    def print(self,*args,**kwds):                                                                                                    
            ^                                                                                                                        
SyntaxError: invalid syntax

Issue #424, opened 2014-04-08
ELIFECYCLE error


EnzoMartin:
I've installed all the dependencies, I've tried running it from VS2010 and VS2012 CMD, Python2.7 is in my path, .NET framework is in my path, I just can't seem to get this working on Windows 7 x64.

Not sure what more to try at this point.. Any help would be greatly appreciated

Path if it helps..

```
C:\Ruby200-x64\bin;
C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common;
C:\Program Files (x86)\ImageMagick-6.8.6-Q16;
C:\Users\emartin\AppData\Roaming\npm;
C:\Program Files\Java\jdk1.7.0_25\bin;
C:\Program Files (x86)\Microsoft SQL Server\100\Tools\Binn\;
C:\Program Files\Microsoft SQL Server\100\Tools\Binn\;
C:\Program Files\Microsoft SQL Server\100\DTS\Binn\;
D:\Android\sdk\platform-tools;
C:\HashiCorp\Vagrant\bin;
C:\Program Files (x86)\Git\cmd;
C:\Program Files\nodejs\;
C:\Program Files (x86)\MySQL\MySQL Utilities 1.3.6\;
C:\Python27;
C:\Python27\Lib;
C:\Python27\DLLs;
C:\Python27\Lib\lib-tk;
C:\Windows\System32;
C:\Program Files\Microsoft SQL Server\110\Tools\Binn\;
C:\Windows\Microsoft.NET\Framework\v4.0.30319;
C:\Program Files\TortoiseGit\bin;
C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\x64
```

```
D:\Repos\my-project>npm install passport-ldapauth
npm http GET https://registry.npmjs.org/passport-ldapauth
npm http 304 https://registry.npmjs.org/passport-ldapauth
npm http GET https://registry.npmjs.org/ldapauth-fork
npm http 304 https://registry.npmjs.org/ldapauth-fork
npm http GET https://registry.npmjs.org/ldapjs/0.7.0
npm http GET https://registry.npmjs.org/lru-cache/2.5.0
npm http GET https://registry.npmjs.org/bcrypt/0.7.7
npm http 304 https://registry.npmjs.org/lru-cache/2.5.0
npm http 304 https://registry.npmjs.org/ldapjs/0.7.0
npm http 304 https://registry.npmjs.org/bcrypt/0.7.7
[...]

Issue #418, opened 2014-03-21
`npm install -g node-gyp` fails


siddhion:
I've been having been trying to get MongoDB to start up properly with my Express.js app but when I run `node app.js` I get the `Failed to load c++ bson extension, using pure JS version` error. I should mention that when I ran `npm install` to install all the modules listed in `package.json` I got these two errors.
```
> kerberos@0.0.3 install /home/max/dev/3lf2/node_modules/mongodb/node_modules/kerberos
> (node-gyp rebuild 2> builderror.log) || (exit 0)

> bson@0.2.5 install /home/max/dev/3lf2/node_modules/mongodb/node_modules/bson
> (node-gyp rebuild 2> builderror.log) || (exit 0)
```
So I figured it was a problem with `node-gyp`. This is holding me back from moving forward with learning about Express and MongoDB. I'm not sure how to proceed so any help is super appreciated.

TooTallNate:
Should be fixed by #527.

PR #502, opened 2014-09-12
Use python2 binary as per PEP 394

### LINKED ISSUE #384: Python search should prefer "python2" executable
On systems where Python 3 is the default, node-gyp complains about an incorrect Python version, even though Python 2 is often available as `python2`.

Per [PEP 394](http://www.python.org/dev/peps/pep-0394/):

> Python 2 only scripts should either be updated to be source compatible with Python 3 or else to use python2 in the shebang line

There are a few distributions that don't provide a `python2` executable. But searching PATH for both `python2` and `python` (and preferring `python2`) should catch cases where Python 2 is installed but node-gyp selects the wrong executable, while still supporting said distributions that don't ship a `python2`.

Is this something you guys'd endorse? If so, happy to provide a pull request.

flying-sheep:
fixes #384

### DIFF
--- a/gyp/gyp
+++ b/gyp/gyp
@@ -5,4 +5,4 @@
 
 set -e
 base=$(dirname "$0")
-exec python "${base}/gyp_main.py" "$@"
+exec python2 "${base}/gyp_main.py" "$@"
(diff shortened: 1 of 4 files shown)


flying-sheep:
hey, this is old and fixes #581 #570 #526 #524 #501 #485 #424 #418 #390 #384

do you want your users to be confused and helpless?

ryneeverett:
+1

rgbkrk:
:+1:

TooTallNate:
Fixed by #527.

PR #527, opened 2014-11-05
Fixed #526. Check for python2 executable by default, fallback to python

### LINKED ISSUE #526: Gyp should check for python2 executable if default python is 3
It's fairly standard these days (at least in *nix flavors) to have "python2" and "python3" binaries installed in an environment where both versions are used. Ideally gyp should check for a python2 executable in the path.  

Gyp currently allows explicitly setting --python, however  one should only need to use this in special cases, not every time npm is called. You can also set the PYTHON environment variable in your .bashrc, but this variable is fairly generic and might affect other programs that depend on it to mean "the default python on this system".

bendavis78:
Fixes #526

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -22,20 +22,23 @@ exports.usage = 'Generates ' + (win ? 'MSVC project files' : 'a Makefile') + ' f
 
 function configure (gyp, argv, callback) {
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Thank you!

PR #534, opened 2014-11-13
rebuild: forward arguments to configure phase

agnat:
The rebuild command currently ignores additional command line arguments. I think it makes sense to pass these arguments to the configure phase:

````
> node-gyp rebuild -- -I some.gypi
````


TooTallNate:
Thanks!

Issue #443, opened 2014-05-07
Build 32 and 64 bits under Windows


LaurentGoderre:
In order to successfully build for the two configuration under Windows I had to do the following tweaks and I'm wondering if there is value in adding those to the project.
1. Changed the Additional Dependencies from
   `kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;DelayImp.lib;"c:\Users\Laurent\.node-gyp\0.10.28\$(Configuration)\node.lib"`
   to
   `kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;DelayImp.lib;"c:\Users\Laurent\.node-gyp\0.10.28\$(Plateform)\node.lib"`
2. Copy the file `.node-gyp\0.10.28\ia32\node.lib` to `.node-gyp\0.10.28\Win32\node.lib

rvagg:
closing due to staleness, I think this might be resolved now anyway?

PR #537, opened 2014-11-20
Avoid copying around node.lib on Windows

corsix:
Imagine a Jenkins build server, where node-gyp chooses a dev dir of say `C:\Users\jenkins\.node-gyp\0.10.29\`. The build server runs multiple build jobs in parallel, each with disjoint working directories, but sharing global state such as the node-gyp dev dir. Some of these build jobs are using node-gyp to build 32-bit binaries, and some are using node-gyp to build 64-bit binaries. Sometimes the following scenario could previously occur:
  1. Job A step 1: Copy `C:\Users\jenkins\.node-gyp\0.10.29\ia32\node.lib` to `C:\Users\jenkins\.node-gyp\0.10.29\Release\node.lib`
  2. Job B step 1: Copy `C:\Users\jenkins\.node-gyp\0.10.29\x64\node.lib` to `C:\Users\jenkins\.node-gyp\0.10.29\Release\node.lib`
  3. Job B step 2: Successfully build and link a 64-bit binary.
  4. Job A step 2: Attempt to build and link a 32-bit binary, but fail because `C:\Users\jenkins\.node-gyp\0.10.29\Release\node.lib` is of the wrong architecture.

This commit avoids the above scenario by making the dev dir immutable, and making the generated vcxproj files refer directly to the node.lib file for the appropriate architecture.
<hr>
This is rather similar to #443 / #450.

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -42,7 +42,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/$(ConfigurationName)/node.lib"'
+          '-l"<(node_root_dir)/$(Platform)/node.lib"'
         ],
         # warning C4251: 'node::ObjectWrap::handle_' : class 'v8::Persistent<T>'
         # needs to have dll-interface to be used by clients of class 'node::ObjectWrap'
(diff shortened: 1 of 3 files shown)


saper:
How would that work when compiling for both `Debug` and `Release` configurations?

Would adding `--node-dir` to the node-gyp invocation help your Jenkins case?

bnoordhuis:
I'll close, there are fresher pull requests that address this.

Issue #21, opened 2012-02-12
Make the dev installation directory configurable (~/.node-gyp)


TooTallNate:
Not a real high priority to me, but :+1: if you want this feature.

rafi:
Until you guys decide, it'll be great if you offer an environment variable or npm style config, _e.g._
```ini
cache  = ${XDG_CACHE_HOME}/node-gyp
```
:)

dantran:
where are we with this issue??. Dont meant to brag, but the discusion started in 2012 :-)

jmesmon:
Having an env-var as @mindon suggested would be useful for me as well. Not clear how to adjust the way npm calls node-gyp.

terinjokes:
@TooTallNate echoing @jmesmon I don't know of a good way to change how npm calls node-gyp for me to add the `--nodedir` header. Can we get this as an environment option?

bnoordhuis:
@terinjokes You can pass `--nodedir` to npm and npm will pass it on to node-gyp.  Alternatively, you can set `npm_config_nodedir=/path/to/sources` in the environment.

terinjokes:
@bnoordhuis I had just edited my comment to point that out.

jcrben:
Does npm support `npm_config_devdir`? I tried to test but not really sure how all this stuff is tied together.

bnoordhuis:
It should.  npm passes on options it doesn't consume itself.

develar:
Since electron-builder 18.4.0 ugly solution HOME env will be not used anymore. Because in this case cache dir is doubled (yarn and npm have to download and cache deps twice). Thanks! `--devdir` and `npm_config_devdir` works in my tests. I will risk to roll out this update to users. Thanks again that now node-gyp allows to set devdir.

jcrben:
never ended up figuring out how this works - I suppose I might dig into it someday... these were the steps, expecting `npm install -g node-gyp` to put ~/.node-gyp into the devdir (which I'm setting with `export npm_config_devdir="$XDG_CACHE_HOME"/.node-gyp` in my `.bashrc` on OSX
![image](https://user-images.githubusercontent.com/5614134/36059216-6949d8d0-0de8-11e8-9101-ce6a1b57989e.png)

@jasonkarns fyi, for me, it doesn't seem like ~/.node-gyp is cache data - it's got include files and stuff to run applications, I suppose in XDG it would end up in XDG_DATA_HOME or something

mighty9245:
[...]

PR #540, opened 2014-11-26
Ability to configure devDir with flag [#21]

glosee:
Potential fix for issue #21 that adds ability to use a flag called `--dev-dir`, and a check for `XDG_CACHE_DIR` before `HOME`.

The `--dev-dir` flag is the be-all-end-all result. If this flag is not used, then the application looks for `XDG_CACHE_DIR` first, then `HOME`, then `USERPROFILE` as it used to.


lib/node-gyp.js
```diff
@@ -98,6 +86,7 @@ proto.configDefs = {
   , 'tarball': String // 'install'
   , jobs: String      // 'build'
   , thin: String      // 'configure'
+  , 'dev-dir': String // everywhere
```
TooTallNate: Flag name should be `--devdir` to parallel the `--nodedir` flag.

dantran:
@TooTallNate, please merge this PR with your suggested change

glosee:
@TooTallNate I completely forgot about this PR until @dantran commented about making your suggested change. I have updated the code to use `devdir` instead of `dev-dir`. Let me know if there is anything else that should change. Thanks!

evaryont:
A reminder/vote for this PR. Is it OK to merge?

saper:
Very useful concept, but can we name it more meaningfully (we also have `nodedir`) ? This is really a cache for the node configuration and include files, so something with `node` and `cache` would be better: `cachenodeinc` maybe?

jasonkarns:
The default for `XDG_CACHE_HOME` is `$HOME/.cache`, so to truly follow the XDG basedir spec, that should be used if `XDG_CACHE_HOME` isn't set (and before `HOME`)


I'll also point out that under the XDG dir, the node-gyp dir ought to be `node-gyp`, not `.node-gyp`.

So I think the directory priority ought to be:

1. `$XDG_CACHE_HOME/node-gyp`
2. `$HOME/.cache/node-gyp`
3. `$HOME/.node-gyp`
4. `$USERPROFILE/.node-gyp`

bnoordhuis:
Making devDir configurable seems reasonable to me but I'm less convinced that honoring XDG_CACHE_HOME is a good thing.  For one, the way it's done in this PR would force everyone to download tarballs again (because the cache directory changes) when it's released.

I agree with @saper that the name is not great.  Something like `--cachedir` better describes what it does.

Also, it would be good to have a regression test or two.

jasonkarns:
> For one, the way it's done in this PR would force everyone to download tarballs again (because the cache directory changes) when it's released.

That's a valid reason against the way this PR is written, but I don't think it's necessarily a strike against XDG_CACHE_HOME in principle.

An option is to choose the directory based on which one *exists* and not by which env var is set (in the order I [listed](https://github.com/nodejs/node-gyp/pull/540#issuecomment-180057762)). This would prevent re-downloading everything. And it would allow a manual (one-time) opt-in for users to just move their `~/.node-gyp` dir into the XDG location if desired.

If none of the directories already exist, then choose (and create) the directory based on the env vars.

bnoordhuis:
What I missed in the discussion in #21 is why following XDG_CACHE_HOME and XDG in general is a good thing.

XDG is not a thing on Windows - people will instead use `set npm_config_cachedir=c:\temp` or whatever we end up naming the flag - so I'm leaning towards simply not bothering and being consistent across platforms.

jasonkarns:
There may not be XDG_* env vars on windows, but the default XDG cache dir is `$HOME/.cache` so using `$HOME/.cache/node-gyp` would be perfectly consistent across platforms. And leveraging the XDG env vars puts node-gyp in line with a host of other tools that let users manage their directory structure through the XDG env vars.

Put simply, if we agree that allowing users to configure where the node-gyp dir is (and I think that's evident with the assumption that `cachedir` is a useful option), then it logically follows that adhering to a spec that makes it possible for users to configure directory locations *across a large and growing number of apps* with just a couple env vars is vastly superior to app-specific flags/config.

bnoordhuis:
Okay, I would suggest that @glosee starts with just the `--cachedir` option and then you or @glosee can follow up with a XDG compatibility PR.  I haven't made up my mind yet about whether said compatibility is preferable but that doesn't need to hold up the core functionality.

jasonkarns:
sounds good

bnoordhuis:
Closing, superseded by #916 which landed earlier today.

jcrben:
Is there a reason why it seems that https://github.com/nodejs/node-gyp/pull/916 declined to go with:

1. dev-dir flag
2. XDG_CACHE_HOME 
3. fall back to $HOME?

I saw some discussion about Windows but that doesn't seem like a good reason to ignore the Unix platform's preferences...

PR #541, opened 2014-11-30
Add target name preprocessor define

agnat:
Adds the name of the current target as a preprocessor define. Uses the automatic variable `_target_name` in combination with late variable expansion.

It originates in joyent/node#8802 but has other applications, too. Think logging. It's a little service to downstream add-ons to keep things DRY.


TooTallNate:
Good idea. Thanks!

PR #560, opened 2015-01-07
Allow for 1.x npmlog

iarna:
Which doesn't break the interface and is only going to 1.x to get
normal semver semantics going forward.


TooTallNate:
Sounds good. Thanks!

PR #564, opened 2015-01-17
Add io.js support

### LINKED ISSUE #563: node-gyp and iojs
Hi,

I'm trying to rebuild `node-leveldown` which calls `install`. The following line assumes `node` is used:

https://github.com/TooTallNate/node-gyp/blob/4ceed113a800378e78435a42632fd035413c99b2/lib/install.js#L188

It's possible to configure with `--disturl=https://iojs.org/dist` so that part is cool.

rvagg:
fixes #563

Added a Makefile and a couple of devdeps to form a functional test that can be used to test compatibility on both Node and io.js, also in the form of a Docker image so you can run the tests on Linux without having install both runtimes locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,23 @@
+test: build-buffertools build-leveldown
+
+build-buffertools:
+	cd node_modules/buffertools/ && \
+		../../bin/node-gyp.js rebuild
+
+build-leveldown:
+	cd node_modules/leveldown/ && \
+		../../bin/node-gyp.js rebuild
+
+test-docker-node:
+	docker run \
+		--rm=true -i -v `pwd`:/opt/node-gyp/ \
+		nodesource/node:wheezy \
+		bash -c 'rm -rf /root/.node-gyp/ && cd /opt/node-gyp && make test'
+
+test-docker-iojs:
+	docker run \
+		--rm=true -i -v `pwd`:/opt/node-gyp/ \
+		iojs/iojs:1.0 \
+		bash -c 'rm -rf /root/.node-gyp/ && cd /opt/node-gyp && make test'
+
+test-docker: test-docker-node test-docker-iojs
--- a/lib/build.js
+++ b/lib/build.js
@@ -14,6 +14,8 @@ var fs = require('graceful-fs')
   , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , win = process.platform == 'win32'
+  , semver = require('semver')
+  , runtime = semver.parse(process.version).major < 1 ? 'node' : 'iojs'
 
 exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
 
@@ -181,15 +183,15 @@ function build (gyp, argv, callback) {
     if (!win || !copyDevLib) return doBuild()
 
     var buildDir = path.resolve(nodeDir, buildType)
-      , archNodeLibPath = path.resolve(nodeDir, arch, 'node.lib')
-      , buildNodeLibPath = path.resolve(buildDir, 'node.lib')
+      , archNodeLibPath = path.resolve(nodeDir, arch, runtime + '.lib')
+      , buildNodeLibPath = path.resolve(buildDir, runtime + '.lib')
 
     mkdirp(buildDir, function (err, isNew) {
       if (err) return callback(err)
       log.verbose('"' + buildType + '" dir needed to be created?', isNew)
       var rs = fs.createReadStream(archNodeLibPath)
         , ws = fs.createWriteStream(buildNodeLibPath)
-      log.verbose('copying "node.lib" for ' + arch, buildNodeLibPath)
+      log.verbose('copying "' + runtime + '.lib" for ' + arch, buildNodeLibPath)
       rs.pipe(ws)
       rs.on('error', callback)
       ws.on('error', callback)
--- a/lib/install.js
+++ b/lib/install.js
@@ -32,6 +32,8 @@ function install (gyp, argv, callback) {
     return callback(e)
   }
 
+console.log(release)
+
   log.verbose('install', 'input version string %j', release.versionStr)
 
   // ensure no double-callbacks happen
@@ -170,8 +172,8 @@ function install (gyp, argv, callback) {
       }
 
       // now download the node tarball
-      var tarPath = gyp.opts['tarball']
-        ,  badDownload = false
+      var tarPath = gyp.opts.tarball
+        , badDownload = false
         , extractCount = 0
         , gunzip = zlib.createGunzip()
         , extracter = tar.Extract({ path: release.devDir, strip: 1, filter: isValid })
@@ -183,19 +185,19 @@ function install (gyp, argv, callback) {
       // only .h header files and the gyp files get extracted
       function isValid () {
         var name = this.path.substring(release.devDir.length + 1)
-        var isValid = valid(name)
+        var validName = valid(name)
         if (name === '' && this.type === 'Directory') {
           // the first directory entry is ok
           return true
         }
-        if (isValid) {
+        if (validName) {
           log.verbose('extracted file from tarball', name)
           extractCount++
         } else {
           // invalid
           log.silly('ignoring from tarball', name)
         }
-        return isValid
+        return validName
       }
 
       gunzip.on('error', cb)
--- a/lib/install.js
+++ b/lib/install.js
@@ -21,6 +21,8 @@ var fs = require('graceful-fs')
   , minimatch = require('minimatch')
   , mkdir = require('mkdirp')
   , win = process.platform == 'win32'
+  , runtime = semver.parse(process.version).major < 1 ? 'node' : 'iojs'
+  , defaultDisturl = runtime == 'node' ? 'http://nodejs.org/dist' : 'https://iojs.org/dist'
 
 function install (gyp, argv, callback) {
 
@@ -39,7 +41,7 @@ function install (gyp, argv, callback) {
     }
   }
 
-  var distUrl = gyp.opts['dist-url'] || gyp.opts.disturl || 'http://nodejs.org/dist'
+  var distUrl = gyp.opts['dist-url'] || gyp.opts.disturl || defaultDisturl
 
 
   // Determine which node dev files version we are installing
@@ -185,7 +187,7 @@ function install (gyp, argv, callback) {
 
       // now download the node tarball
       var tarPath = gyp.opts['tarball']
-      var tarballUrl = tarPath ? tarPath : distUrl + '/v' + version + '/node-v' + version + '.tar.gz'
+      var tarballUrl = tarPath ? tarPath : distUrl + '/v' + version + '/' + runtime + '-v' + version + '.tar.gz'
         , badDownload = false
         , extractCount = 0
         , gunzip = zlib.createGunzip()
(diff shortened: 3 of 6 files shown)


lib/install.js
```diff
@@ -21,6 +21,8 @@ var fs = require('graceful-fs')
   , minimatch = require('minimatch')
   , mkdir = require('mkdirp')
   , win = process.platform == 'win32'
+  , runtime = semver.parse(process.version).major < 1 ? 'node' : 'iojs'
```
othiym23: Longer-term, I'd really like to see iojs add an `"iojs"` property to `process.versions` so this test, which feels like a flaky heuristic that somebody, somewhere is going to forget about at some point, with unpredictable, but probably dire, consequences.

package.json
```diff
@@ -15,7 +15,7 @@
   "author": "Nathan Rajlich <nathan@tootallnate.net> (http://tootallnate.net)",
   "repository": {
     "type": "git",
-    "url": "git://github.com/TooTallNate/node-gyp.git"
+    "url": "git://github.com/rvagg/pan-gyp.git"
```
othiym23: This is a strange change to include in this PR.

Makefile
```diff
@@ -0,0 +1,23 @@
+test: build-buffertools build-leveldown
+
+build-buffertools:
```
othiym23: This and `build-leveldown` should probably depend on an `npm install` task if you want these tests to be self-contained for CI.

lib/install.js
TooTallNate: Agreed. This is pretty lame.

TooTallNate:
General comment: this seems like a departure from being "node-compatible". Why is changing the build file names so necessary?

othiym23:
I proposed iojs/io.js#491 to improve the tests for whether `node-gyp` is being run against `iojs` or `node`.

package.json
isaacs: Yeah, seems like this is a typo or something.

isaacs:
Just so I'm clear, changing the build filenames is because io.js is shipping as a binary named "iojs" and only symlinked to "node", right?

It seems like, rather than key on the version or rely on a `process.versions.iojs`, we might want to add some kind of API in iojs for "What is the build filename target?", which we can also add to Node.js, or any other forks/distros in the future.

That might also actually bring peace to any remaining debiants who still call it "/usr/bin/nodejs".

TooTallNate:
@isaacs Possibly something like this?

``` js
path.basename(fs.realpathSync(process.execPath))
```

isaacs:
In the very short term, it seems like a version number switch is the only option before us, and in the medium term, adding a one-off flag for iojs is probably easier than getting something into Node, so +1 from me for pragmatism on both counts, as long as we don't lose sight of cleaner solutions to the core meta-problem.

Tests and everything else about this pull req are awesome.

isaacs:
@TooTallNate Actually you don't even need the realpath, since execPath is already realpathed.

```
$ which node
/usr/local/bin/node

$ node
> process.execPath
'/usr/local/bin/iojs'
```

Makefile
rvagg: howso? that's why they are in devDependencies, doesn't that suffice?

lib/install.js
rvagg: discussion here https://github.com/iojs/io.js/pull/491

package.json
rvagg: derp, I temporarily released `pangyp` while we wait for this merge, this line shouldn't have made it here and is about to disappear as soon as I stop typing this

rvagg:
`execPath` is problematic on Windows, we currently have a link but a stub `node.exe` was planned; I don't know if that's still on the table, @piscisaureus

rvagg:
re build filenames - we're not going to get people to download node-v1.0.2.tar.gz to get io.js, it's a different product and is Node-compatible (ish) but there's a limit to what we can do. Remember that Joyent were the ones that forced us to change from "node-forward" to io.js because of tradmark threats so we need some differentiation.

isaacs:
If process.execPath isn't an option, then I'd strongly prefer a more full-featured `process.release` option which could answer these questions explicitly, rather than inferring from the version number.

rvagg:
`process.release`:

```
{
  source: "https://iojs.org/dist/v1.0.2/iojs-v1.0.2.tar.gz"
  headers: "https://iojs.org/dist/v1.0.2/iojs-v1.0.2-headers.tar.gz" # fictional but easy
  lib: 'https://iojs.org/dist/v1.0.2/win-x64/iojs.lib' # only on Windows
}
```

Then node-gyp could even support nightlies.

TooTallNate:
Very nice @rvagg :)

piscisaureus:
> execPath is problematic on Windows, we currently have a link but a stub node.exe was planned; I don't know if that's still on the table, @piscisaureus

The stub trick I wanted to use turned out to be a bit problematic (issues with the stack protection cookie; ask me on irc for details.) The hard link solution we have now seems to work fine, so I want to stick to that.

> {
>  source: "https://iojs.org/dist/v1.0.2/iojs-v1.0.2.tar.gz"
> headers: "https://iojs.org/dist/v1.0.2/iojs-v1.0.2-headers.tar.gz" # fictional but easy
>  lib: "https://iojs.org/dist/v1.0.2/win-x64/iojs.lib" # only on Windows
> }

That seems like a nice solution. 
I'd still like to ship the headers plus lib file in the installer so no download would be necessary at all.

rvagg:
headers are shipped with all distros but Windows currently, they have been from somewhere in Node 0.10, but nobody has put in the work to make node-gyp aware of them.

However, even if we do use installed headers there's no guarantee that they are actually going to be available on the system (plain binary install, stripped-down installer of some kind, manual install), so there needs to be a facility to fetch the headers anyway.

XadillaX:
```
➜  thmclrx git:(master) node --help
Usage: node [options] [ -e script | script.js ] [arguments]
       node debug script.js [arguments]

Options:
  -v, --version        print node's version
  -e, --eval script    evaluate script
  -p, --print          evaluate script and print result
  -i, --interactive    always enter the REPL even if stdin
                       does not appear to be a terminal
  --no-deprecation     silence deprecation warnings
  --throw-deprecation  throw an exception anytime a deprecated function is used
  --trace-deprecation  show stack traces on deprecations
  --v8-options         print v8 command line options
  --max-stack-size=val set max v8 stack size (bytes)

Environment variables:
NODE_PATH              ':'-separated list of directories
                       prefixed to the module search path.
NODE_MODULE_CONTEXTS   Set to 1 to load modules in their own
                       global contexts.
NODE_DISABLE_COLORS    Set to 1 to disable colors in the REPL

Documentation can be found at https://iojs.org/
```

here you can see `iojs`.

lib/install.js
```diff
@@ -21,6 +21,8 @@ var fs = require('graceful-fs')
...
+  , defaultDisturl = runtime == 'node' ? 'http://nodejs.org/dist' : 'https://iojs.org/dist'
```
fengmk2: Any idea to support nvm custom dist url env by default? https://github.com/creationix/nvm/blob/master/nvm.sh#L73

rvagg: @fengmk2 would this be to support mirrors, like in China? I'm tinkering around in there currently and it wouldn't be hard to bring in custom URLs from where my local version is at.

XadillaX: Not only for mirrors, but also can be used in private NPM services.

fengmk2: @rvagg yes, it will be very helpful for us.

rvagg:
WIP but I've pushed my latest. Introduces a `getRelease()` that inspects `process.release` and does some other magic in lieu of that not being available.

The Windows bits haven't been done yet and I'm not sure `process.release` will be the final way to get this information either.

fengmk2:
Any progress?

rvagg:
yes, I spent a whole afternoon shaving yak to get my Windows machine working properly to compile and test for this .. so there's that.

I'm just working on some io.js release automation stuff so I can hand off some release responsibility but other than that this is top of my list and I am chipping away at it. It'd be nice to push `process.release` (or similar) into 1.1.0 to support this but we'll see what timing ends up being like.

toddself:
This causes updating NPM when using io.js to break when trying to use node-gyp, since the non-included version of npm still calls out to nodejs.org to download the tarball.

Would be great if `npm install -g npm` works correctly on io.js without having to do some `npm link` trickery.

Qard:
:+1: I want this so [traceview](https://github.com/appneta/node-traceview/) can support both iojs and node 0.8 on TravisCI, which requires a `npm install -g npm` to get around many caret dependencies leaking into to various packages and breaking node 0.8 with old npm.

mzgol:
> I want this so traceview can support both iojs and node 0.8 on TravisCI, which requires a npm install -g npm

@Qard You can do it conditionally in your `.travis.yml`:
```
before_install:
  - if [ "`node --version | cut -f2- -dv | cut -f1-2 -d.`" == "0.8" ]; then npm -g install npm; fi
```

Qard:
True, I simplified it a bit though. Don't need all those cuts in there. :)

Globegitter:
Would really love to see this so projects don't have to depend in pangyp as well as node-gyp to support both iojs as well as nodejs. Has slowed down my switch to iojs quite a bit.

am11:
@rvagg, for io.js process name detection, would it make sense to use `execPath` as follow:

```powershell
node -p 'process.execPath.split(/[\\/]+/).pop()'
# returns node.exe on Windows

iojs -p 'process.execPath.split(/[\\/]+/).pop()'
# returns iojs.exe on Windows
```

Alternatively, we can shell out `'tasklist /FI "PID eq' + process.pid + '" /NH /FO CSV'` (returns comma-separated value) and read the first chunk to get `iojs.exe` or `node.exe`.

lib/build.js
```diff
@@ -14,6 +14,8 @@ var fs = require('graceful-fs')
   , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , win = process.platform == 'win32'
+  , semver = require('semver')
+  , runtime = semver.parse(process.version).major < 1 ? 'node' : 'iojs'
```
pksunkara: Won't this be a problem when node v1 is released?

mikeal:
What's the status on this? It's hard to put together just from the comments and all the references from other GitHub issues.

TooTallNate:
@mikeal I think https://github.com/iojs/io.js/pull/493 is the major hold-up at this point.

lib/build.js
tcr: Can there be an override for this as an env variable? e.g. https://github.com/tcr/pangyp/commit/b9209b874ef8ac74111a389e16a8ea90978c70be

If I want to enable `gyp configure --target=1.3.0` *and* `gyp configure --target=0.12.0`, I'd rather not have to install both `iojs` and `node` to create this effect.

XadillaX:
I think https://github.com/XadillaX/gyp-io-service will temporarily resolve this problem before the official version of node-gyp is published.

You just need to start this service and you can run `node-gyp rebuild` under io.js.

leecade:
:+1:

leecade:
@XadillaX thanks https://github.com/XadillaX/gyp-io-service works for me.

fengmk2:
I temporary use pangyp to build io.js and nodejs cpp module now.
Need npm@>2.7.0

```bash
$ npm i -g pangyp
$ npm --node-gyp=$your-npm-global-dir/pangyp/bin/node-gyp.js
```

leecade:
@fengmk2 出手 PR 啊, 哈哈

XadillaX:
@leecade siwujidan yong zhongwen zhende haome?

trevorlinton:
Curious question... Why is the node native module system requiring a CRT//DllMain type initialization?

Couldn't NODE_MODULE macro be changed to define a stub "#modulename#_register" that returns the initialization function (rather than trying to run the node_module_register from within #modulename#_register_" that then gets ran and passed the options exports?

The current method relies on a very smart dynamic link loader thats transitive (and OS X is very much so) but with Windows and statically linking the node.lib its creating a lot of sticky situations where the proc addresses are being hard coded for, i'm not sure why.

Granted, i may not fully understand this but it seems like theres a simple non-breaking change that could be made if both node.js and io.js agreed to mutually change the definition of NODE_MODULE macro.

Or just have the NODE_MODULE macro kick out symbols for the old way and the new way, and node.js can continue doing whatever it wants, and others like io.js or tint can successfully register modules on windows. 

Thoughts?

magicae:
Why this pr isn't merged so far?

lib/install.js
```diff
@@ -15,15 +15,27 @@ var fs = require('graceful-fs')
...
 
+  // Determine which node dev files version we are installing
+  var release
+  try {
+    release = getRelease(gyp, argv[0] || gyp.opts.target)
+  } catch (e) {
+    return callback(e)
+  }
+
+console.log(release)
```
mhart: Remove this (I assume?)

deepsweet:
does pangyp solution work for io.js@3?

lib/build.js
micky2be: in iojs you can check for `process.release.name`.
in nodejs `process.release` doesn't even exist, yet.

To set as a target I would recommend to use same notation as nvm: `gyp configure --target=iojs-v3.0.0`

JCMais:
@deepsweet yes

rvagg:
new impl, using process.release and taking convergence into consideration https://github.com/nodejs/node-gyp/pull/711

PR #593, opened 2015-03-18
Avoid downloading shasums if using tarPath

chapel:
If using a manually provided `tarPath` the [content shasums](https://github.com/TooTallNate/node-gyp/blob/master/lib/install.js#L250-L254) aren't [actually computed](https://github.com/TooTallNate/node-gyp/blob/master/lib/install.js#L225) and downloading the remote shasums is a bit of a waste.

This avoids the call, and makes it so you can provide a local tarball with the necessary source files without any external network calls.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -277,9 +277,12 @@ function install (gyp, argv, callback) {
         var installVersionPath = path.resolve(devDir, 'installVersion')
         fs.writeFile(installVersionPath, gyp.package.installVersion + '\n', deref)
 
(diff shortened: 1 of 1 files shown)


chapel:
@TooTallNate Can I get some eyes on this, it is a simple change and will help immensely. I currently have to run an internal fork of node-gyp to prevent the shasum download on our internal servers (which don't have network access to the internet).

TooTallNate:
Thanks!

chapel:
:smile:

PR #599, opened 2015-03-25
win: allow node.exe/iojs.exe to be renamed

piscisaureus:
On Windows, when node or io.js attempts to dynamically load a compiled
addon, the compiled addon tries to load node.exe or iojs.exe again -
depending on which import library the module used when it was linked.
This makes it impossible to rename node.exe or iojs.exe, because when
that happens the module can't find its dependencies.

With this patch, a delay-load hook is added to all modules that are
compiled with node-gyp. The delay-load hook ensures that whenever a
module tries to load imports from node.exe/iojs.exe, it'll just refer
back to the process image, thus making it possible to rename the
iojs/node binary.

Bug: iojs/io.js#751
Bug: iojs/io.js#965
Downstream PR: iojs/io.js#1251

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -1,7 +1,9 @@
 {
   'target_defaults': {
     'type': 'loadable_module',
+    'win_delay_load_hook': 'false',
     'product_prefix': '',
+
     'include_dirs': [
       '<(node_root_dir)/src',
       '<(node_root_dir)/deps/uv/include',
@@ -13,10 +15,33 @@
         'product_extension': 'node',
         'defines': [ 'BUILDING_NODE_EXTENSION' ],
       }],
+
       ['_type=="static_library"', {
         # set to `1` to *disable* the -T thin archive 'ld' flag.
         # older linkers don't support this flag.
         'standalone_static_library': '<(standalone_static_library)'
+      }],
+
+      ['_win_delay_load_hook=="true"', {
+        # If the addon specifies `'win_delay_load_hook': 'true'` in its
+        # binding.gyp, link a delay-load hook into the DLL. This hook ensures
+        # that the addon will work regardless of whether the node/iojs binary
+        # is named node.exe, iojs.exe, or something else.
+        'conditions': [
+          [ 'OS=="win"', {
+            'sources': [
+              'src/win_delay_load_hook.c',
+            ],
+            'msvs_settings': {
+              'VCLinkerTool': {
+                'DelayLoadDLLs': [ 'iojs.exe', 'node.exe' ],
+                # Don't print a linker warning when no imports from either .exe
+                # are used.
+                'AdditionalOptions': [ '/ignore:4199' ],
+              },
+            },
+          }],
+        ],
       }],
     ],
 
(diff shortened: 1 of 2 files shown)


TooTallNate:
This seems fine to me (though I admit I don't really understand what the code itself is doing).

othiym23:
I need to know when this is merged / when a new version of node-gyp is released that incorporates this so I can pull it down and include it in npm, so I'm adding myself to this issue. :angel:

addon.gypi
```diff
@@ -47,6 +50,15 @@
         # warning C4251: 'node::ObjectWrap::handle_' : class 'v8::Persistent<T>'
         # needs to have dll-interface to be used by clients of class 'node::ObjectWrap'
         'msvs_disabled_warnings': [ 4251 ],
+        # Set up delay-loading for node.exe/iojs.exe so the loadable module
+        # will still be able to find it's imports if the binary is renamed.
```
bnoordhuis: s/it's/its/

bnoordhuis:
Clever.  LGTM sans grammar nit.

justinmchase:
I'm not sure this is going to work... it would be better to optionally link to `node.lib` or `iojs.lib`. I could be misunderstanding but, for example, we are actually running via http://nwjs.io which embeds io.js into the app but does not run the actual process iojs.exe, nor is that process even available at runtime.

I made a private fork of node-gyp and basically just went in and replaced node.lib with iojs.lib and fixed up a few other urls and then it all started working for me when run with this command:

```
node-gyp --msvs_version=2013 --target=1.2.0 configure build
```

Note that if you just use `--dist-url=https://iojs.org/dist` it gets you pretty close, but again iojs has renamed their main lib to iojs.lib and they also moved a few files to new folders `win-x86` and `win-x64`. A few changes to strings and then all of a sudden it starts working.

piscisaureus:
@justinmchase
> I'm not sure this is going to work... it would be better to optionally link to node.lib or iojs.lib. I could be misunderstanding but, for example, we are actually running via http://nwjs.io which embeds io.js into the app but does not run the actual process iojs.exe, nor is that process even available at runtime.

Have you tried it? I think that this patch at best (or at worst) would have no effect.

justinmchase:
@piscisaureus Sorry, I deleted the comment. I realized that you were actually probably talking about something else. There are a lot of threads on the forums here that I needed to read through. It seems like you guys have it under control and I look forward to getting the fix. I'm not blocked in the meantime.

piscisaureus:
@TooTallNate 
FYI I'm looking at enabling the delay-load hook by default (making it opt-out instead of opt-in) in io.js 2.x (https://github.com/iojs/io.js/pull/1433/files).

justinmchase:
Can I still opt-out? The reason I ask is that it may break native modules in http://nwjs.io. I had to write my own delay load function that included `nw.exe` to get it to work and if you include it by default then it may be a problem. 

If I can turn it off, or if it's possible to append entries to the delay load list in the gyp file then that would be fine.

piscisaureus:
@justinmchase
With this patch, module authors have to opt-in (disabled by default). However in io.js 2.0 we may make it opt-out (iow, enabled by default).

The `win_delay_load_hook` gyp setting controls the behavior. See the commit message for more info: https://github.com/piscisaureus/node-gyp/commit/52ceec3a6d15de3a8f385f43dbe5ecf5456ad07a

I think it would be fine to replace the ".exe whitelist" by logic that enables the delay-load hook for any exe.

Issue #571, opened 2015-02-02
use standard cp arguments


bradain:
The cp command takes an argument -a as a shortcut for -pPR.
While this works on most Linux distributions, it is not available on certain other operating systems such as OpenBSD, while the standard -pPR always works. 
Would you consider replacing cp -af with the functionally equivalent cp -fpPR to improve OS support?

drook:
Same stuff on Solaris.

Fishrock123:
Huh. I thought I commented on this.

`cp` resides in gyp itself here, so I'm not sure if we can do that? cc @TooTallNate, do we still ever get upstream updates from gyp?

Should we just have a nodejs/gyp that both node and node-gyp can use?

alexijs:
Same stuff on Oracle Solaris 11.2

TooTallNate:
@Fishrock123 Historically I've just updated gyp here when node.js/io.js updated its own copy of gyp (probably behind at this point!)

alexijs:
Is there a way to resolve this issue by changing cp flags? Compilation of all packages via node-gyp fails on Oracle Solaris 11.2 because Oracle doesn't support GNU flag -a.

I found a workaround by compiling package locally, creating a package structure, and moving *.node, package.json, and corresponding js files there.
However, it defeats a purpose of autobuilder and too cumbersome because I have to compile package after package.

Also, it looks like running Demeteorizer is not enough and I have to write custom builder that would replace packages from my dev (OS X Maverics) with those locally compiled on Solaris.

Im trying to deploy my Meteor app on Solaris via Demeteorizer and have problems with fibers and bcrypt.

Any help would be appreciated.

drook:
You could also play with Makefile (or, if I don't remember correctly, with any other template issuing "cp -r").

bnoordhuis:
Little known fact but you can override the cp command: `make cmd_copy='ln -f $< $@ || cp -PRfp $< $@'`

If someone sends a patch to the upstream gyp project, I'll be happy to land it here (assuming it's accepted upstream, of course.)  Grep for `copy_archive_args` in pylib/gyp/generator/make.py.

mhdawson:
As an FYI we just did a similar change for AIX here https://codereview.chromium.org/1368133002/

bnoordhuis:
@mhdawson You can make that part of #753 if you want.

GeorgeBailey:
> @GeorgeBailey: I wonder if you experience the same problems on your system.

[...]

PR #603, opened 2015-03-31
cp -a flag is not portable

kaspars:
at least it's not on OpenBSD. probably somewhere else as well.


bnoordhuis:
I'm going to close this, subsumed by https://codereview.chromium.org/1483473002/.

Until it's fixed in upstream gyp, see https://github.com/nodejs/node-gyp/issues/571#issuecomment-149000420 for a workaround.

PR #605, opened 2015-04-01
Changed clean to only delete node-gyp object files

marksweiss:
With this change, node-gyp only deletes object files it created. Resolves a conflict if node-gyp shares a build directory with anything else. This is the simplest possible way to solve this, though it relies on the convention that node-gyp and only node-gyp object files will have the suffix *.node.

### DIFF
--- a/lib/clean.js
+++ b/lib/clean.js
@@ -14,7 +14,7 @@ var log = require('npmlog')
 function clean (gyp, argv, callback) {
 
   // Remove the 'build' dir
-  var buildDir = 'build'
+  var buildDir = 'build/*.node'
 
   log.verbose('clean', 'removing "%s" directory', buildDir)
   rm(buildDir, callback)


TooTallNate:
This isn't really a safe assumption. gyp generates lots of other files other than the `*.node` files including a Makefile and intermediary build files.

marksweiss:
Thanks for following up!

I understand your point.  There is a mistake in my pull request, which would be quite easy to fix. The actual path should be "build/Release/*.node"

In the project where I am using node-gyp, I have a build/Release directory I see only a .lock file and *.node files.

I have a local shell script in my project root which I use. It has the following three commands:
  rm build/Release/*.node
  node-gyp configure
  node-gyp build

This has never led to any issues.

If this makes sense I could fix the path and send a new pull request.

bnoordhuis:
I'm going to close this.  As @TooTallNate says, the assumption that only `*.node` files need to be removed is unsound.

PR #612, opened 2015-04-17
Only add "-undefined dynamic_lookup" to loadable_module targets

kkaefer:
Adding `-undefined dynamic_lookup` to all targets makes it impossible to create Xcode projects with node-gyp for gyp files that also include `none` targets. Since `none` targets don't have a framework phase, gyp errors when trying to add the link flag.

In addition, `-undefined dynamic_lookup` should be added as a linker flag, not a link library, so it's now passed in the `OTHER_LDFLAGS` setting.

This also fixes #536.


TooTallNate:
Thanks!

PR #616, opened 2015-04-27
move HMODULE m declaration to top

cosmosgenius:
The declaration of the m variable not on top causes the build to fail on VS2010(C89), Window 7 x64. 
Providing solution based on [stackoverflow answer](http://stackoverflow.com/questions/9903582/error-c2275-illegal-use-of-this-type-as-an-expression).


piscisaureus:
lgtm, ship it

kkoopa:
Do note that io.js 3.0 will not build with VS 2010 at all due to V8 4.3.

kaycoinsUSA (commented):
Complete

Issue #636, opened 2015-05-26
Node-gyp have problems for long path.


lygstate:
```
c1 : fatal error C1083: Cannot open source file: '..\..\..\..\..\..\..\..\..\..\..\..\Users\lygstate\AppData\Roaming\npm\node_modules\npm\node_module
s\node-gyp\src\win_delay_load_hook.c': No such file or directory [C:\CI-Cor\node_modules\karma\node_modules\socket.io\node_modules\socket.io-client\n
ode_modules\engine.io-client\node_modules\ws\build\bufferutil.vcxproj]
  validation.cc
..\src\validation.cc(112): warning C4996: 'NanSymbol': was declared deprecated [C:\CI-Cor\node_modules\karma\node_modules\socket.io\node_modules\sock
et.io-client\node_modules\engine.io-client\node_modules\ws\build\validation.vcxproj]
          C:\CI-Cor\node_modules\karma\node_modules\socket.io\node_modules\socket.io-client\node_modules\engine.io-client\node_modules\ws\node_module
  s\nan\nan.h(616) : see declaration of 'NanSymbol'
  win_delay_load_hook.c
c1 : fatal error C1083: Cannot open source file: '..\..\..\..\..\..\..\..\..\..\..\..\Users\lygstate\AppData\Roaming\npm\node_modules\npm\node_module
s\node-gyp\src\win_delay_load_hook.c': No such file or directory [C:\CI-Cor\node_modules\karma\node_modules\socket.io\node_modules\socket.io-client\n
ode_modules\engine.io-client\node_modules\ws\build\validation.vcxproj]
```

lygstate:
What's the current file dir for gypi

lygstate:
Temp fix for this issue:
modify addon.gypi
With:
          [ 'OS=="win"', {
            'sources': [
              'C:/Users/lygstate/AppData/Roaming/npm/node_modules/npm/node_modules/node-gyp/src/win_delay_load_hook.c',
            ],

So maybe we need to dynamically generate addon.gypi or some other ways.

PR #637, opened 2015-05-27
Setting NODE_GYP_DIR for addon.gypi to setting absolute path for src/…

lygstate:
…win_delay_load_hook.c,

and fixes of the long relative path issue on Win32.
Fixes of #636

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -40,7 +40,7 @@
         'conditions': [
           [ 'OS=="win"', {
             'sources': [
-              'src/win_delay_load_hook.c',
+              '<(node_gyp_dir)/src/win_delay_load_hook.c',
             ],
             'msvs_settings': {
               'VCLinkerTool': {
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -316,12 +316,14 @@ function configure (gyp, argv, callback) {
       // Windows expects an absolute path
       output_dir = buildDir
     }
+    var nodeGypDir = path.resolve(__dirname, '..')
 
     argv.push('-I', addon_gypi)
     argv.push('-I', common_gypi)
     argv.push('-Dlibrary=shared_library')
     argv.push('-Dvisibility=default')
     argv.push('-Dnode_root_dir=' + nodeDir)
+    argv.push('-Dnode_gyp_dir=' + nodeGypDir)
     argv.push('-Dmodule_root_dir=' + process.cwd())
     argv.push('--depth=.')
     argv.push('--no-parallel')


bin/node-gyp.js
```diff
@@ -20,6 +21,7 @@ var log = require('npmlog')
 var prog = gyp()
 var completed = false
 prog.parseArgv(process.argv)
+process.env['NODE_GYP_DIR'] = path.resolve(__dirname, '..');
```
TooTallNate: We should set this as a variable in `config.gypi`, rather than as an env variable.

addon.gypi
```diff
@@ -40,7 +40,7 @@
         'conditions': [
           [ 'OS=="win"', {
             'sources': [
-              'src/win_delay_load_hook.c',
+              '$(NODE_GYP_DIR)/src/win_delay_load_hook.c',
```
TooTallNate: Remember to use `<(node_gyp_dir)` syntax when switching to a gyp variable

bin/node-gyp.js
```diff
@@ -12,6 +12,7 @@ process.title = 'node-gyp'
 
 var gyp = require('../')
 var log = require('npmlog')
+var path = require("path");
```
TooTallNate: styling: remove semicolons please

TooTallNate: and single quotation marks for the require call

bin/node-gyp.js
lygstate: That's not possible, because NODE_GYP_DIR is calculated.
cause gypi can not retrieve the current directory.

addon.gypi
lygstate: Already have a node_gyp_dir variable? I can not found this.

bin/node-gyp.js
lygstate: I'll fix the style problem, I care more about if the environment solution is acceptable

bin/node-gyp.js
lygstate: hope for fast response...

addon.gypi
lygstate: You didn't tell how to use gyp variable yet.
I don't know how to setting NODE_GYP_DIR in with correct semantics.


2015-05-30 1:50 GMT+08:00 Nathan Rajlich <notifications@github.com>:

> In addon.gypi
> <https://github.com/TooTallNate/node-gyp/pull/637#discussion_r31347695>:
>
> > @@ -40,7 +40,7 @@
> >          'conditions': [
> >            [ 'OS=="win"', {
> >              'sources': [
> > -              'src/win_delay_load_hook.c',
> > +              '$(NODE_GYP_DIR)/src/win_delay_load_hook.c',
>
> Remember to use <(node_gyp_dir) syntax when switching to a gyp variable
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TooTallNate/node-gyp/pull/637/files#r31347695>.
>



-- 
         此致
礼
罗勇刚
Yours
    sincerely,
Yonggang Luo

lygstate:
Updated.

TooTallNate:
Much better now! Thanks!

saper:
Unfortunately this breaks MSBuild `%(RelativeDir)%` variable, which suddenly becomes absolute. See #732

PR #664, opened 2015-07-17
allow semver@5

othiym23:
@isaacs merged a PR that removed the built artifacts for browsers, which is a breaking version, so everything using semver needs to be updated to allow the new major version. It's a low-risk change, because there are no other changes to the package.


TooTallNate:
Merged. Sorry for the delay!

PR #668, opened 2015-07-19
Fix #667 - use `which`'s result

### LINKED ISSUE #667: Inconsistent python2 exe running
My "weird" system has two executables named `python2`:
```
[12:59:24.60] c:\Users\refael>where python2
C:\Tools\res\scripts\python2.bat
C:\bin\msys64\usr\bin\python2.exe

[13:02:04.47] c:\Users\refael>C:\Tools\res\scripts\python2.bat -V
C:\bin\dev\python27\python.exe -V
Python 2.7.8

[13:02:06.15] c:\Users\refael>C:\bin\msys64\usr\bin\python2.exe -V
Python 2.7.9
```
When running `node-gyp configure` the wrong one is run:
```
...
gyp verb check python checking for Python executable "python2" in the PATH
gyp verb `which` succeeded python2 C:\Tools\res\scripts\python2.BAT
gyp verb check python version `python2 -c "import platform; print(platform.python_version());"` returned: "2.7.9\n"
...
```

refack:


refack:
Fixes #750

lib/configure.js
```diff
@@ -48,6 +48,7 @@ function configure (gyp, argv, callback) {
         }
       } else {
         log.verbose('`which` succeeded', python, execPath)
+        python = execPath
```
bnoordhuis: What does this fix?  If `which('python')` succeeds, then by definition python is on the PATH, right?

refack: the `which` module is more similar to windows PATH resolution semantics then the following require('child_process').execFile in line 85.
`execFile` fails to resolve *.bat and *.cmd files from the PATH, but will run them if explicitly specified to.

bnoordhuis: Can you add a comment explaining that (Properly capitalized and punctuated, please.)?

It would be nice to have some kind of regression test for this.  Maybe you can `module.exports.test.checkPython = checkPython` and verify in the test that executing the result with `-V` prints the python version banner?

bnoordhuis: I went ahead and filed a PR that adds a regression test: https://github.com/nodejs/node-gyp/pull/756

refack:
Added comment.
As for adding a test, I'm hesitant since it will need to manipulate the environment, while beeing windows specific...

bnoordhuis:
Thanks, landed in 268f1ca.

Issue #297, opened 2013-07-26
Problem with non-acsii userprofile name on Windows (7)


danyaPostfactum:
Problem with `.vcxproj` file. node-gyp writes it in system codepage (non-utf8), while xml prolog specifies `utf-8` in `encoding` attribute. This causes error while trying to build the project.
My userprofile contains cyrillic characters, so I get non-ACSII characters in paths.

knyazhe1:
@MattouFP
your "*Fix encoding problems in easy_xml.py*" (https://github.com/TooTallNate/node-gyp/pull/554):
```python
+ # Fix encoding
+ xml_string = unicode(xml_string, 'latin-1').encode(encoding)
```
is not working with cyrillic symbols :disappointed:
for correct work in this case  :ru: need another patch:
```python
+ # Fix encoding for cyrillc symbols
+ xml_string = unicode(xml_string, 'Windows-1251').encode(encoding)
```
This is not good solution in whole. Need a change solution provider maybe :interrobang:

Sogl:
@knyazhe1, It not helps me...=(
Windows 8.1 x64, VS 2013 sp4

Can't install loopback:
http://loopback.io/getting-started/

Output:
```
C:\Users\Артем\AppData\Roaming\npm\node_modules\strongloop\node_modules\strong-s
upervisor\node_modules\heapdump>if not defined npm_config_node_gyp (node "C:\nod
ejs\node_modules\npm\bin\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.
js" rebuild )  else (rebuild)
Сборка проектов в этом решении по одному. Чтобы включить параллельную сборку, до
бавьте параметр "/m".
C:\Users\Артем\AppData\Roaming\npm\node_modules\strongloop\node_modules\strong-
supervisor\node_modules\heapdump\build\addon.vcxproj(47,46): error MSB4025: не
удалось загрузить файл проекта. Недопустимый знак для указанной кодировки., стр
ока 47, позиция 46.
gyp ERR! build error
gyp ERR! stack Error: `C:\Program Files (x86)\MSBuild\12.0\bin\msbuild.exe` fail
ed with exit code: 1
gyp ERR! stack     at ChildProcess.onExit (C:\nodejs\node_modules\npm\node_modul
es\node-gyp\lib\build.js:269:23)
gyp ERR! stack     at ChildProcess.emit (events.js:110:17)
[...]

PR #672, opened 2015-07-25
fixing ansi path related issues

Chudesnov:
As I've already mentioned, paths passed to vcbuild/msbuild are still encoded incorrectly on non-Latin editions of Windows. Here's a more flexible solution that, hopefully, will fix ANSI→UTF encoding issues on Windows 2000 and higher.

Fixes: #297

### DIFF
--- a/gyp/pylib/gyp/easy_xml.py
+++ b/gyp/pylib/gyp/easy_xml.py
@@ -4,6 +4,7 @@
 
 import re
 import os
+import locale
 
 
 def XmlToString(content, encoding='utf-8', pretty=False):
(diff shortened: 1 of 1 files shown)


Sogl:
:+1:

khanhoi:
Gvgg

bnoordhuis:
Is this still an issue?  The patch should be sent to the upstream gyp project but I'll be happy to cherry-pick it once it lands.

Chudesnov:
Yes, it is.

No problem, I'll submit a pull request to bnoordhuis/gyp when I get home.

bnoordhuis:
I'm not the upstream, https://code.google.com/p/gyp is. :-)

Sending patches to Google projects is kind of hassle (need to check out depot_tools, upload patch with git-cl) but I'll be happy to walk you through it if you get stuck.  Please post the link to the CL in the issue when you're done so we can track it.

Chudesnov:
It seems that depot_tools and gclient in particular suffer from the very same bug - incorrectly re-encoding ansi paths into unicode. Gonna try to fix that as well :)

roman-yerin:
@Chudesnov any chance to fix it in upstream?

bnoordhuis:
Seeing how there has been no follow-up, I'll go ahead and close this.

PR #686, opened 2015-08-09
Update Install.js, add iojs support

camsong:
This PR is much like https://github.com/TooTallNate/node-gyp/pull/669, but this one really works.

Fixes
>  gyp ERR! stack Error: 404 response downloading https://nodejs.org/dist/v3.0.0/node-v3.0.0.tar.gz


gabeio:
when I run `((process.release || {}) || {}).name` I get undefined using iojs 2.5.0.

camsong:
Yep, this API works since v3.0.0, close this PR unless this is reliable way to check iojs.

PR #708, opened 2015-08-31
win: enable delay-load hook by default

Fishrock123:
Enables the following commit by default:
52ceec3a6d15de3a8f385f43dbe5ecf5456ad07a

See also:
https://github.com/nodejs/node/commit/3bda6cbfa4a9bb073790d53bc14e85b6e575bbe5

Seeing if we can get this in, even into a major, to land in npm, so that we don't need to keep patching it in io.js/node.

I talked with @othiym23 a bit last Friday and he seemed to think it wasn't major-worthy. In io.js we wated to do it in a major.

@TooTallNate what are your thoughts on this?

cc @nodejs/tsc?

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -1,7 +1,7 @@
 {
   'target_defaults': {
     'type': 'loadable_module',
-    'win_delay_load_hook': 'false',
+    'win_delay_load_hook': 'true',
     'product_prefix': '',
 
     'include_dirs': [


rvagg:
this is one for @nodejs/platform-windows, specifically @piscisaureus who introduced it. I think this is more for io.js than node so maybe not as urgent?

Fishrock123:
> I think this is more for io.js than node so maybe not as urgent?

@othiym23 seemed to think it also applied to electron and nw.js. If it needs to be in a major I'd prefer to get it into npm@3.

rvagg:
@Fishrock123 let's merge this and float on v4

rvagg:
so, I know we're in a transition phase here and still wanting to defer to @TooTallNate somewhat but I'm also assuming that he's happy for us to take the incentive to get stuff done, so I've gone ahead and merged this and used the core style of merging including adding metadata, I've also added a `semver-major` here (so we could use changelog-maker here if we want to as well). It's hard for me to assess whether this really is strictly a semver-major but given that binaries compiled with this change will operate differently on Windows I think we can afford to be cautious and make this bump to v3.0.0.

TooTallNate:
:+1: Feel free to push to master. I've also added @rvagg and @Fishrock123 as owners on the npm package. Cheers!

Issue #701, opened 2015-08-23
The node-gyp and io.js meta-issue


Fishrock123:
node-gyp does not work on io.js. This has been a known issue since io.js 1.0.0-ish.

The reason is that node-gyp is hardcoded to do certain things in respect to how node is distributed on nodejs.org. This issue will not into the reasons on why node-gyp needs to download tarballs.

As such, io.js floats a number of floating patches on npm/node-gyp to make node-gyp work, [as can be seen here](https://github.com/nodejs/node/commits/master/deps/npm).

If you are running io.js and attempt to build native modules using a different version of npm than shipped with your version of io.js, your native module builds **will fail**.

https://github.com/nodejs/node-gyp/pull/564 and https://github.com/nodejs/node-gyp/pull/669 attempt to fix this, but as of the time of this issue, both are out of date and neither uses io.js 3.0.0's `process.release`, which was introduced to alleviate this. Version checking could be acceptible for io.js 1.0.0> <3.0.0, as those will only ever be io.js releases.

TooTallNate:
> Version checking could be acceptible for io.js 1.0.0> <3.0.0, as those will only ever be io.js releases.

I do want to add this backwards-compat logic for sure, and of course proper `process.release` checking.

sbrl:
This module seems to be used all over the place. The `canvas` and `ttystudio` packages are just two example. Since it doesn't work with io.js, this prevents me from using a whole bunch of modules, even though my system meets all the other requirements :(

In addition, I think that once io.js and Node.js merge, the io.js version system is going to be used. Does  that mean that `node-gyp` is going to be broken on the next major release on Node.js too?

rvagg:
node-gyp v3, released yesterday, supports both io.js and Node.js, historical versions and near future versions, it'll be bundled with Node.js v4 but isn't bundled with npm just yet—that's coming soon hopefully.

node-canvas has problems of its on in getting up to speed with the current releases, it's been a little painful making the code work with the new V8 but they are nearly ready last I looked. I don't know about ttystudio but they may just need to upgrade to NAN v2 and you'll have to file an issue with them about it—there's nothing node-gyp, nor Node core can do about this, it's all on V8.

Issue #114, opened 2012-07-30
node-gyp Error - UNCAUGHT EXCEPTION 


kumarldh:
Hi,

I was trying to install websocket and got this error

```
# npm install websocket
npm WARN websocket-server@1.4.04 package.json: bugs['web'] should probably be bugs['url']

> websocket@1.0.5 preinstall /home/kumar/node_modules/websocket
> node-gyp configure

gyp ERR! UNCAUGHT EXCEPTION 
gyp ERR! stack Error: Invalid protocol
gyp ERR! stack     at Request.init (/home/kumar/node_modules/node-gyp/node_modules/request/main.js:291:31)
gyp ERR! stack     at new Request (/home/kumar/node_modules/node-gyp/node_modules/request/main.js:103:8)
gyp ERR! stack     at request (/home/kumar/node_modules/node-gyp/node_modules/request/main.js:858:11)
gyp ERR! stack     at download (/home/kumar/node_modules/node-gyp/lib/install.js:132:15)
gyp ERR! stack     at /home/kumar/node_modules/node-gyp/lib/install.js:203:17
gyp ERR! stack     at Object.oncomplete (/home/kumar/node_modules/node-gyp/node_modules/mkdirp/index.js:46:26)
gyp ERR! System Linux 2.6.32-5-686
gyp ERR! command "node" "/home/kumar/node_modules/.bin/node-gyp" "configure"
gyp ERR! cwd /home/kumar/node_modules/websocket
gyp ERR! node -v v0.6.4
gyp ERR! node-gyp -v v0.6.2
gyp ERR! This is a bug in `node-gyp`.
gyp ERR! Please file an Issue:
gyp ERR!     <https://github.com/TooTallNate/node-gyp/issues>
npm ERR! error installing websocket@1.0.5 Error: websocket@1.0.5 preinstall: `node-gyp configure`
npm ERR! error installing websocket@1.0.5 `sh "-c" "node-gyp configure"` failed with 7
npm ERR! error installing websocket@1.0.5     at ChildProcess.<anonymous> (/usr/local/lib/node_modules/npm/lib/utils/exec.js:49:20)
npm ERR! error installing websocket@1.0.5     at ChildProcess.emit (events.js:70:17)
npm ERR! error installing websocket@1.0.5     at maybeExit (child_process.js:359:16)
npm ERR! error installing websocket@1.0.5     at Process.onexit (child_process.js:395:5)
```

Thanks!

PR #711, opened 2015-09-04
use process.release, make aware of io.js & node v4 differences

rvagg:
Looking for patient reviewers, ASAP please. I'd like to be able to ship v4 with this version patched over the top of npm and upstream it to npm as soon as they are happy with it as well.

* introduce a lib/process-release.js module that encapsulates all of the logic around downloads, names, versions and directories
* introduce a test for process-release.js that hits it with various configurations and asserts that it comes up with the right answers
* make node-gyp aware of process.release where it exists (`>=3.0.0`) and use it to determine download locations and names
* make node-gyp aware of "iojs" where process.release doesn't exist, thereby making it work for _all versions of Node and io.js_ from 0.8 to current so an `npm install npm@latest -g` even on one of the older versions of io.js will _just work_ (magic).
* make node-gyp able to compile against header-only tarballs, see the one-line change in addon.gypi and the search in two locations for common.gypi, for `>=3.0.0` we are shipping header-only tarballs which are super quick to download, but it also needs to be able to compile against full source as well
* keep the `--target` stuff intact (although I'd love to know if anybody actually uses that)
* make node-gyp aware of the change in location for node.lib (& iojs.lib) for `>=1.0.0`. In `<1.0.0` you find 32-bit node.lib in the root of the version's dist directory (e.g. http://nodejs.org/dist/v0.12.7/node.lib) while the 64-bit node.lib is in an x64 subdirectory (e.g. http://nodejs.org/dist/v0.12.7/x64/node.lib), for io.js we took the opportunity to clean this up and they are now in win-x86 and win-x64 subdirectories (e.g. https://iojs.org/dist/v3.3.0/win-x86/iojs.lib & https://iojs.org/dist/v3.3.0/win-x64/iojs.lib) _(and the .msi's have been renamed and are in the parent directory, but that's not relevant here)_. Note that it still downloads both 32-bit and 64-bit .lib files, I haven't changed that logic.
* introduced a test/docker.sh script that runs it forreal, but only on Linux obviously:
  1. set up a base image with compile tooling
  2. set up a clones image on top of the base image with git clones of some native addons to test
  3. set up an image on top of the clones image for each of the versions of node and io.js to test _plus_ the latest version of npm over the top
  4. start a new container for each version of node & io.js, copy `pwd` into npm's node-gyp dependency folder so it uses the current code for testing when you use `npm install` etc., then do an `npm install` of the projects to see how it behaves.
  * you can see it downloading and using the source tarballs for all but io.js v3.3.0 for which it grabs the headers file because it's only in v3 that we started shipping that and introduced `process.release`

What does this buy us?
 * The ability to ship **actually useful** RC's, nightlies and master/unstable/canary builds which don't have to have proper release version numbers and sit in the release downloads directories, node-gyp will _just work_
 * The additional ability for sophisticated users to create custom / corporate / internal builds of Node that point to header/source tarballs that are at a different location e.g. inside a protected & HA network, thereby removing one more point of failure & build/bundle time (add in npme or similar and you can be _fully_ HA). Note that this requires a custom build of Node, so re security it's the same discussion as the `--link-module` feature we introduced.
* The ability to support existing versions of io.js with new versions of npm (no more "you have to use the npm that ships with that version of io.js). We also get the ability to re-use the io.js brand if we revive it and start shipping again for some purpose, no more floating patches.

What's not here?

We need to finally make node-gyp aware of locally installed headers: this has been on multiple people's radars for a very long time but has never got done but it should get done. When you install via a complete tarball, or a `make install` from the source, you get the header files installed for that version. node-gyp should be able to use those to avoid any downloading. However this is not a panacea because:
 1. This hasn't been enabled at all for Windows installers, although that's not an impossible task
 2. Some old versions don't unpack headers and some just have broken headers—from memory there's some broken stuff across 0.12 and a couple of versions of io.js shipped without openssl headers
 3. Some users don't have complete installs, we offer a plain node.exe for instance
 4. node-gyp still has the ability to "cross compile" addons which won't work here (`--target`)
 5. version managers will have to get creative with how they ship versions and communicate with node-gyp about the locations of header files
 7. Doesn't solve the node.lib problem on Windows, although we may be able to ship that in the msi as well
 6. ... more mess ..

So this is a TODO, someone needs to take a bite at this soon, but it's not a top priority for shipping core.

Possible reviewers: @nodejs/build, @nodejs/addon-api, @TooTallNate, @Fishrock123, @othiym23, @bnoordhuis, @piscisaureus


addon.gypi
```diff
@@ -5,6 +5,7 @@
     'product_prefix': '',
 
     'include_dirs': [
+      '<(node_root_dir)/include/node',
```
rvagg: headers tarball has a different structure to the source tarball, no deps or src subdirectories, but this addition makes existing addons work with a headers tarball, a source tarball or a `--nodedir`

lib/configure.js
```diff
@@ -310,42 +302,47 @@ function configure (gyp, argv, callback) {
...
     var addon_gypi = path.resolve(__dirname, '..', 'addon.gypi')
-    var common_gypi = path.resolve(nodeDir, 'common.gypi')
-    var output_dir = 'build'
-    if (win) {
-      // Windows expects an absolute path
-      output_dir = buildDir
-    }
-    var nodeGypDir = path.resolve(__dirname, '..')
+    var common_gypi = path.resolve(nodeDir, 'include/node/common.gypi')
+    fs.exists(common_gypi, function (exists) {
```
rvagg: we look for both `/include/node/common.gypi` (source) and `/common.gypi` (headers) and use the one that's there

rvagg: diff below here looks messy but it's mainly indenting

mzgol: Viewing the diff on GitHub with `?w=1` in the URL helps with that issue. :)

lib/install.js
```diff
@@ -350,40 +341,39 @@ function install (gyp, argv, callback) {
...
       function downloadNodeLib (done) {
-        log.verbose('on Windows; need to download `node.lib`...')
+        log.verbose('on Windows; need to download `' + release.name + '.lib`...')
         var dir32 = path.resolve(devDir, 'ia32')
           , dir64 = path.resolve(devDir, 'x64')
-          , nodeLibPath32 = path.resolve(dir32, 'node.lib')
-          , nodeLibPath64 = path.resolve(dir64, 'node.lib')
-          , nodeLibUrl32 = distUrl + '/v' + version + '/node.lib'
-          , nodeLibUrl64 = distUrl + '/v' + version + '/x64/node.lib'
+          , libPath32 = path.resolve(dir32, release.name + '.lib')
```
rvagg: we have local `libPath*`'s and also `release.libPath*`'s, the former are where we put the x.lib files for the compiler to use (`ia32` and `x64` directories) and the latter is their location on the server and in the shasums lookup

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+
+  var overrideDistUrl = gyp.opts['dist-url'] || gyp.opts.disturl
+  var iojs = !overrideDistUrl && !defaultRelease && semver.gte(versionStr, '1.0.0')
+  var name = (defaultRelease && defaultRelease.name) || (iojs ? 'iojs' : 'node')
+  var defaultDirUrl = (overrideDistUrl || iojs ? 'https://iojs.org/download/release' : 'https://nodejs.org/dist') + '/v' + versionStr + '/'
+  var baseUrl
+  var libUrl32
+  var libUrl64
+
+  if (defaultRelease && defaultRelease.headersUrl && !overrideDistUrl) {
```
rvagg: this branch is for `>=3.0.0`, it leans on `process.release`

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
+var semver = require('semver')
+var url = require('url')
+var path = require('path')
+
+var bitsre = /(x32|x64)/
+
+function processRelease (argv, gyp, defaultVersion, defaultRelease) {
```
rvagg: we don't look up anything on `process` here, those are passed in by the caller, this is mainly about keeping this self-contained and to make it possible to test it against different versions of those properties

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+    libPath64: path.relative(url.parse(baseUrl).path, url.parse(libUrl64).path)
+  }
+}
+
+function resolveLibUrl (name, defaultUrl, bytes, version) {
+  var base = url.resolve(defaultUrl, './')
+
+  if (!bitsre.test(base)) {
+    // let's assume it's a baseUrl then
+    if (semver.gte(version, '1.0.0'))
```
rvagg: `<1.0.0` we have the `/` and `/x64/` locations for node.lib, after that we introduce `/win-x64/` and `/win-x86/`

lib/configure.js
rvagg: neato! does that stand for "whitespace" or something?

lib/configure.js
```diff
@@ -135,35 +137,25 @@ function configure (gyp, argv, callback) {
 
     } else {
       // if no --nodedir specified, ensure node dependencies are installed
-      var version
-      var versionStr
-
-      if (gyp.opts.target) {
+      if ('v' + release.versionStr != process.versio) {
```
Fishrock123: `!== process.version`?

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
+var semver = require('semver')
+var url = require('url')
+var path = require('path')
+
+var bitsre = /(x32|x64)/
+
+function processRelease (argv, gyp, defaultVersion, defaultRelease) {
+  var versionStr = argv[0] || gyp.opts.target || defaultVersion
```
Fishrock123: `version`?

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+
+  // parse the version to normalize and ensure it's valid
+  var versionSemver = semver.parse(versionStr)
```
Fishrock123: Likewise, maybe just call this `semver`?

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+  if (!versionSemver) {
+    return { versionStr: versionStr }
+  }
+  // flatten version into String
+  versionStr = versionSemver.version
+
+  var overrideDistUrl = gyp.opts['dist-url'] || gyp.opts.disturl
+  var iojs = !overrideDistUrl && !defaultRelease && semver.gte(versionStr, '1.0.0')
```
Fishrock123: I'd semver-range it between 1.0.0 and 4.0.0, just to be sure/safe.

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+      tarballUrl: defaultRelease.headersUrl,
+      shasumsFile: 'SHASUMS256.txt',
+      shasumsUrl: url.resolve(baseUrl, 'SHASUMS256.txt'),
+      checksumAlgo: 'sha256',
+      versionDir: (defaultRelease.name != 'node' ? defaultRelease.name + '-' : '') + versionStr,
+      libUrl32: libUrl32,
+      libUrl64: libUrl64,
+      libPath32: path.relative(url.parse(baseUrl).path, url.parse(libUrl32).path),
+      libPath64: path.relative(url.parse(baseUrl).path, url.parse(libUrl64).path)
+    }
```
Fishrock123: Ah nvm, I guess these are named reasonably enough.

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+    return {
+      versionStr: versionStr,
+      versionSemver: versionSemver,
+      name: defaultRelease.name,
+      baseUrl: baseUrl,
+      tarballUrl: defaultRelease.headersUrl,
+      shasumsFile: 'SHASUMS256.txt',
+      shasumsUrl: url.resolve(baseUrl, 'SHASUMS256.txt'),
+      checksumAlgo: 'sha256',
+      versionDir: (defaultRelease.name != 'node' ? defaultRelease.name + '-' : '') + versionStr,
```
Fishrock123: !==

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+  var name = (defaultRelease && defaultRelease.name) || (iojs ? 'iojs' : 'node')
+  var defaultDirUrl = (overrideDistUrl || iojs ? 'https://iojs.org/download/release' : 'https://nodejs.org/dist') + '/v' + versionStr + '/'
+  var baseUrl
+  var libUrl32
+  var libUrl64
+
+  if (defaultRelease && defaultRelease.headersUrl && !overrideDistUrl) {
+    baseUrl = url.resolve(defaultRelease.headersUrl, './')
+    libUrl32 = resolveLibUrl(name, defaultRelease.libUrl || baseUrl || defaultDirUrl, 32, versionStr)
+    libUrl64 = resolveLibUrl(name, defaultRelease.libUrl || baseUrl || defaultDirUrl, 64, versionStr)
```
Fishrock123: Shouldn't we always have `defaultRelease.libUrl` here, or is this to guard against pre-releases?

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+  if (!versionSemver) {
+    return { versionStr: versionStr }
+  }
+  // flatten version into String
+  versionStr = versionSemver.version
+
+  var overrideDistUrl = gyp.opts['dist-url'] || gyp.opts.disturl
+  var iojs = !overrideDistUrl && !defaultRelease && semver.gte(versionStr, '1.0.0')
+  var name = (defaultRelease && defaultRelease.name) || (iojs ? 'iojs' : 'node')
+  var defaultDirUrl = (overrideDistUrl || iojs ? 'https://iojs.org/download/release' : 'https://nodejs.org/dist') + '/v' + versionStr + '/'
```
Fishrock123: Seems like this should go below the `process.release` branch, if it wasn't for [lines 28 & 29](https://github.com/nodejs/node-gyp/pull/711/files#r38764357)

lib/process-release.js
rvagg: well, this still works on !windows, not that it needs to but it does make it much easier to write tests for if there is no windows-specific logic in here

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+  return {
+    versionStr: versionStr,
+    versionSemver: versionSemver,
+    name: name,
+    baseUrl: baseUrl,
+    tarballUrl: baseUrl + name + '-v' + versionStr + '.tar.gz',
+    shasumsFile: shasumsFile,
+    shasumsUrl: baseUrl + shasumsFile,
+    checksumAlgo: checksumAlgo,
+    versionDir: (name != 'node' ? name + '-' : '') + versionStr,
```
Fishrock123: !==

lib/process-release.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function resolveLibUrl (name, defaultUrl, bits, version) {
+  var base = url.resolve(defaultUrl, './')
+
+  if (!bitsre.test(defaultUrl)) {
+    // let's assume it's a baseUrl then
+    if (semver.gte(version, '1.0.0'))
+      return url.resolve(base, 'win-x' + bits +'/' + name + '.lib')
+    return url.resolve(base, (bits == 64 ? 'x64/' : '') + name + '.lib')
```
Fishrock123: @rvagg I had forgotten. I want to make what i think is a better suggestion here.

Can we just modify those old folders to give them a new directory structure? I don't think that should effect the SHASUMS, and would probably be more robust.

lib/install.js
```diff
@@ -31,34 +34,29 @@ function install (gyp, argv, callback) {
     if (err) {
       log.warn('install', 'got an error, rolling back install')
       // roll-back the install if anything went wrong
-      gyp.commands.remove([ version ], function (err2) {
+      gyp.commands.remove([ release.versionStr ], function (err2) {
```
Fishrock123: This should't be `versionDir`, right? i can't tell where `version` came from here..

lib/process-release.js
rvagg: I've been thinking about this and considered putting in redirects to the server to make it all work normally but unfortunately it will affect the shasums files which I would have to modify with the new structure and I really don't want us to be messing with directories that users consider set in stone. I'm treating those 0.12 and prior directories as sacred and not even blowing the dust off them!

lib/install.js
```diff
@@ -31,34 +34,29 @@ function install (gyp, argv, callback) {
...
-    return callback(new Error('Minimum target version is `0.8.0` or greater. Got: ' + versionStr))
+  if (semver.lt(release.versionStr, '0.8.0')) {
+    return callback(new Error('Minimum target version is `0.8.0` or greater. Got: ' + release.versionStr))
   }
 
   // 0.x.y-pre versions are not published yet and cannot be installed. Bail.
-  if (version.prerelease[0] === 'pre') {
-    log.verbose('detected "pre" node version', versionStr)
+  if (release.versionSemver.prerelease[0] === 'pre') {
+    log.verbose('detected "pre" node version', release.versionStr)
```
Fishrock123: I don't think this will be quite true in the future. Maybe still for the 4.0 release though. (Not necessary for this PR, but..)

1. We should check for a `-pre.X` if possible, if so it is one of the official pre-releases. (If I understand what we are doing with versioning pre-release majors.)
2. There should be a flag / env variable to point to your local node repository, since you probably have if this is happening.

lib/install.js
```diff
@@ -156,7 +150,7 @@ function install (gyp, argv, callback) {
   }
 
   function getContentSha(res, callback) {
-    var shasum = crypto.createHash(checksumAlgo)
+    var shasum = crypto.createHash(release.checksumAlgo)
```
Fishrock123: Should we bother checking if the Algo is available? Everything has SHA265 these days I think?

rvagg: 0.8 isn't

lib/install.js
```diff
@@ -220,12 +213,12 @@ function install (gyp, argv, callback) {
...
       if (tarPath) {
-        var input = fs.createReadStream(tarballUrl)
+        var input = fs.createReadStream(tarPath)
         input.pipe(gunzip).pipe(extracter)
         return
       }
 
-      var req = download(tarballUrl)
+      var req = download(release.tarballUrl)
       if (!req) return
```
Fishrock123: Should this.. error?

lib/install.js
```diff
@@ -31,34 +34,29 @@ function install (gyp, argv, callback) {
...
-  if (semver.lt(versionStr, '0.8.0')) {
-    return callback(new Error('Minimum target version is `0.8.0` or greater. Got: ' + versionStr))
+  if (semver.lt(release.version, '0.8.0')) {
+    return callback(new Error('Minimum target version is `0.8.0` or greater. Got: ' + release.version))
   }
 
   // 0.x.y-pre versions are not published yet and cannot be installed. Bail.
-  if (version.prerelease[0] === 'pre') {
-    log.verbose('detected "pre" node version', versionStr)
+  if (release.semver.prerelease[0] === 'pre') {
```
rvagg: @Fishrock123 `-pre` matching comes from https://github.com/nodejs/node/blob/a338eb3c3e712fcaacdb475cc6f3378259a3d534/src/node_version.h#L19 i.e. you compiled this thing yourself, you should use `--nodedir` to point to it

lib/configure.js
rvagg: fixed

lib/process-release.js
rvagg: s/versionStr/version, done

lib/process-release.js
rvagg: done, but left the variable as `versionSemver`, exposing it as a `semver` property on the return object, remember we have a `semver` package required in here

lib/process-release.js
rvagg: done

lib/process-release.js
rvagg: done .. but you do know how silly that is right? also, do a grep for `==` vs `===` and `!=` vs `!==` in this codebase, there are no rules

lib/install.js
Fishrock123: `localLibPath*`?

lib/process-release.js
rvagg: done, under duress

lib/remove.js
```diff
@@ -19,23 +19,19 @@ function remove (gyp, argv, callback) {
...
-  if (!v) {
+  if (!version) {
     return callback(new Error('You must specify a version number to remove. Ex: "' + process.version + '"'))
   }
 
-  // parse the version to normalize and make sure it's valid
-  var version = semver.parse(v)
-
-  if (!version) {
-    return callback(new Error('Invalid version number: ' + v))
```
Fishrock123: Is statement now missing? What if the version is invalid?

lib/remove.js
```diff
@@ -19,23 +19,19 @@ function remove (gyp, argv, callback) {
...
+  var versionSemver = semver.parse(version)
+  if (versionSemver) {
+    // flatten the version Array into a String
+    version = versionSemver.version
   }
 
-  // flatten the version Array into a String
-  version = version.version
-
   var versionPath = path.resolve(gyp.devDir, version)
```
Fishrock123: Won't this need to specify like above? e.g. `(name != 'node' ? name + '-' : '') + versionStr`

lib/install.js
rvagg: yes, you're right, fixed

lib/install.js
rvagg: see my other dangling comment about `-pre` coming from self-compiled-from-repository builds, also you can `npm config set nodedir=/path/to/repo` and also set the `npm_config_nodedir=/path/to/repo` for the same effect

lib/process-release.js
Fishrock123: > there are no rules

Time to start some :)

lib/install.js
rvagg: old request bug I believe, related to https://github.com/nodejs/node-gyp/issues/114, I dare not touch it without tests to understand it

lib/process-release.js
Fishrock123: Oh wait, !windows doesn't have those. I forgot.

lib/install.js
rvagg: nothing else is called `local` in the codebase, I'm opting for blending in here

lib/process-release.js
Fishrock123: Interesting. Oh well.

lib/install.js
Fishrock123: I mean, no reason to check the platform's available algos, right?

lib/remove.js
rvagg: it could come in here like `io.js-3.0.0` which is invalid, i.e. `versionDir`, and anyway, if it's an invalid version then it just won't delete

lib/install.js
Fishrock123: I'd put a TODO or log something. I can in a different PR if I remember.

lib/remove.js
rvagg: that comes in through `versionDir` from install, and if you want to remove an io.js version then you'd need to call it as `io.js-3.3.0`

I doubt very much whether this is ever called directly though so I'm not so concerned about tuning this to some imaginary scenario of how someone might want to use it

lib/install.js
rvagg: no, that's not checked here anyway, it's in process-release.js and boils down to

```js
  var checksumAlgo = semver.gte(version, '0.10.0') ? 'sha256' : 'sha1'
```

that's all this is for because there are no SHASUM256.txt files in 0.8.x dirs

lib/configure.js
Fishrock123: Can I suggest doing a `stat` here? `exists` is sorta deprecated, or scheduled for deprecation.

(or just do it sync)

lib/install.js
rvagg: oh, you know what this is for, it's for the `cb(err)` case inside `download` where it doesn't return a `req` because there was an error setting it up; so this makes perfect sense and the error gets handled in the callback.

leaving it alone

lib/configure.js
rvagg: changed to .stat, just running the docker tests to confirm there is peace in the world

lib/process-release.js
Fishrock123: Maybe add a comment on that?

lib/process-release.js
Fishrock123: Also comment?

Fishrock123:
> but it also needs to be able to compile against full source as well

@rvagg only for older versions, right?

Fishrock123:
Also, Fixes: https://github.com/nodejs/node-gyp/issues/701

rvagg:
> > but it also needs to be able to compile against full source as well

> @rvagg only for older versions, right?

No, also for when you're compiling from the source tarball you can `--nodedir` and same for when you're doing it directly from the repo—like a lot of us in core have to do; we're stuck with both

rvagg:
comments added to process-release.js

lib/install.js
othiym23: How is this change going to interact with legacy deployments originally built using an older version of `node-gyp`? Will the old `node.lib` just be left there untouched until `node_modules` gets nuked?

othiym23:
I've given it a quick pass-through but I want to pull it down and run the tests, which is going to take an hour or two. So far: looks good, like the addition of tests, npm will probably just integrate this as-is once it's published.

If it weren't for Node 4, I'd probably only land this in `npm@3`, because there are probably going to be enough knock-on effects from the switch as to cause issues (or at least surprises) for some users upgrading across `npm@2` versions. I kind of wish that either `npm@3` were the version going into Node 4 or this had a little more time to bake in before Node 4 drops, because I'm not sure I can anticipate all the implications of some of the changes in here.

TooTallNate:
Great stuff @rvagg!

lib/install.js
rvagg: Should leave everything in place–and it should set up the .node-gyp/* folders identical to older versions of node-gyp for 0.x versions of Node, including directory names source unpack and node.lib files.

rvagg:
I've just run through the tests manually on Windows .. what a pain, but I've ironed out a bunch of issues with node.lib in the process, including needing to reach in to addon.gypi to insert a new `node_lib_file` parameter passed through with a `-D`. I can also confirm that we need https://github.com/nodejs/node-gyp/pull/708 on top of this in order to get addons working for anything >= 1.0.0. PTAL @Fishrock123, I've force-pushed over the single commit but if you could just scan through once more you'll see some new stuff.

Lastly, there's this:

```
$ node -pe process.release
{ name: 'io.js',
  sourceUrl: 'https://iojs.org/download/release/v3.3.0/iojs-v3.3.0.tar.gz',
  headersUrl: 'https://iojs.org/download/release/v3.3.0/iojs-v3.3.0-headers.tar.gz',
  libUrl: 'https://iojs.org/download/release/v3.3.0/win-ia32/iojs.lib' }
```

note the `ia32` which should be `x86`, baked in to the process.

doh!

so I'm bailing on getting an rc.1 out for v4 and going to bed and will tackle this final piece of the puzzle after some sleep.

Fishrock123:
> note the ia32 which should be x86, baked in to the process.

Crap, gota fix that for v4

rvagg:
adding in this:

```patch
diff --git a/lib/process-release.js b/lib/process-release.js
index e6dcc49..70e90ab 100644
--- a/lib/process-release.js
+++ b/lib/process-release.js
@@ -3,6 +3,8 @@ var url = require('url')
 var path = require('path')

 var bitsre = /(x86|x64)/
+var bitsreV3 = /(x86|ia32|x64)/ // io.js v3.x.x shipped with "ia32" but should
+                                // have been "x86"

 // Captures all the logic required to determine download URLs, local directory and
 // file names and whether we need to use SHA-1 or 2. Inputs come from command-line
@@ -83,8 +85,10 @@ function normalizePath (p) {

 function resolveLibUrl (name, defaultUrl, arch, version) {
   var base = url.resolve(defaultUrl, './')
+  var isV3 = semver.satisfies(version, '^3')
+  var hasLibUrl = bitsre.test(defaultUrl) || (isV3 && bitsreV3.test(defaultUrl))

-  if (!bitsre.test(defaultUrl)) {
+  if (!hasLibUrl) {
     // let's assume it's a baseUrl then
     if (semver.gte(version, '1.0.0'))
       return url.resolve(base, 'win-' + arch  +'/' + name + '.lib')
@@ -93,7 +97,7 @@ function resolveLibUrl (name, defaultUrl, arch, version) {
   }

   // else we have a proper url to a .lib, just make sure it's the right arch
-  return defaultUrl.replace(bitsre, arch)
+  return defaultUrl.replace(isV3 ? bitsreV3 : bitsre, arch)
 }

 module.exports = processRelease
```

which is special-casing io.js v3 since we can't change those and the binary says win-ia32 but the directory and shasums file have win-x86

ralphtheninja:
I played around with `./bin/node-gyp.js` and `lib/process-release.js` always use `process.release` even though `--target` is used.

Here the headers for `3.3.0` will be installed in `~/.node-typ/iojs-3.2.0`:

installing-wrong-headers

lib/configure.js
```diff
@@ -310,42 +302,49 @@ function configure (gyp, argv, callback) {
...
-    var common_gypi = path.resolve(nodeDir, 'common.gypi')
-    var output_dir = 'build'
-    if (win) {
-      // Windows expects an absolute path
-      output_dir = buildDir
-    }
-    var nodeGypDir = path.resolve(__dirname, '..')
+    var common_gypi = path.resolve(nodeDir, 'include/node/common.gypi')
+    fs.stat(common_gypi, function (err, stat) {
+      if (err || !stat.isFile()) {
```
bnoordhuis: I'd remove the `!stat.isFile()` check.  It skips symlinks and that makes for rather non-obvious behavior.

lib/install.js
```diff
@@ -31,34 +34,29 @@ function install (gyp, argv, callback) {
...
   // Determine which node dev files version we are installing
-  var versionStr = argv[0] || gyp.opts.target || process.version
-  log.verbose('install', 'input version string %j', versionStr)
+  log.verbose('install', 'input version string %j', release.version)
 
   // parse the version to normalize and ensure it's valid
-  var version = semver.parse(versionStr)
-  if (!version) {
-    return callback(new Error('Invalid version number: ' + versionStr))
+  if (!release.semver) {
```
bnoordhuis: Either the comment is wrong now or you should still call `semver.parse()` here.

lib/install.js
bnoordhuis: Just a suggestion but wouldn't it be easier to generate SHA256SUM.txt files for old releases than special-case here?

lib/install.js
```diff
@@ -350,40 +341,39 @@ function install (gyp, argv, callback) {
         })
       }
 
+      //TODO: download appropriate node.lib
```
bnoordhuis: Can you use `// TODO(rvagg)` and expound on what "appropriate" means here?

lib/process-release.js
```diff
@@ -0,0 +1,99 @@
+var semver = require('semver')
+var url = require('url')
+var path = require('path')
+
+var bitsre = /(x86|x64)/
+
+// Captures all the logic required to determine download URLs, local directory and 
+// file names and whether we need to use SHA-1 or 2. Inputs come from command-line
```
bnoordhuis: s/2/SHA-256/?  SHA-2 is a family of hash functions, of which SHA-256 is just one.

lib/install.js
Fishrock123: I think it means to actually download the `.lib`?

lib/configure.js
rvagg: fixed

lib/install.js
rvagg: updated comment and tweaked

lib/install.js
rvagg: yeah, sorry, that was an old TODO, removed

lib/process-release.js
rvagg: fixed, by removing

rvagg:
I've done a heavy refactor of process-release.js to fix a few problems and clear up the logic:

* `--dist-url` works properly, this is tested in a few ways in both the test-process-release.js and the docker.sh runner, there's a `simple-proxy.js` that's used as an alternate URL
* `--target` works now (thanks @ralphtheninja), this is tested in a few ways in both the test-process-release.js and the docker.sh runner
* I've removed `cryptoAlgo` and made everything `SHASUMS256.txt` because I've added it to all of the v0.8 directories on nodejs.org that were missing it, thanks @bnoordhuis for the suggestion.
* Added checks for the `NVM_IOJS_ORG_MIRROR` and `NVM_NODEJS_ORG_MIRROR` environment variables as was [suggested](https://github.com/rvagg/pangyp/pull/3) by @fengmk2 as alternatives to `--dist-url`, this feature is particularly useful for the Chinese community. Does anyone have serious objections to the `NVM_` prefixes here? They are already in use in the wild.
* Lots more tests in both JS and docker.sh

lib/process-release.js
```diff
@@ -0,0 +1,129 @@
...
+  if (!hasLibUrl) {
+    // let's assume it's a baseUrl then
+    if (semver.gte(version, '1.0.0'))
+      return url.resolve(base, 'win-' + arch  +'/' + name + '.lib')
+    // prior to io.js@1.0.0 32-bit node.lib lives in /, 64-bit lives in /x64/
+    return url.resolve(base, (arch === 'x64' ? 'x64/' : '') + name + '.lib')
+  }
+
+  // else we have a proper url to a .lib, just make sure it's the right arch
+  return defaultUrl.replace(isV3 ? bitsreV3 : bitsre, arch)
```
bnoordhuis: This could theoretically fail if someone uses a custom URL at `mx64.company.com` or `files.company.com/arch/x86`?

bnoordhuis:
LGTM with a hypothetical.  I didn't review the docker script but I assume it works.

lib/process-release.js
rvagg: yes, but that has to be one of the rules for use—if you're going to distribute a binary with a custom `libUrl` then you need to work with this ..

ralphtheninja:
@rvagg Will `node-gyp` have support for installing headers for nightlies as well? That would be super useful if you want to stay ahead of the curve and test your native addons with coming releases.

rvagg:
@ralphtheninja yep, that's exactly one of the use cases we're optimising for here, nightlies, release candidates, canaries, whatever else we want to put up for people to use, node-gyp will just work.

ralphtheninja:
> @ralphtheninja yep, that's exactly one of the use cases we're optimising for here, nightlies, release candidates, canaries, whatever else we want to put up for people to use, node-gyp will just work. 

:+1: This makes me really happy :)

rvagg:
Updated to address the concern @bnoordhuis pointed out, it now looks for `/\/win-(x86|x64)\//` in the URL to switch for the two arch's, so you have to have withe `/win-x86/` or `/win-x64/` in the URL, which is harder to overload.

Unless there are no objections, I might merge this and release a v3 eh?

bnoordhuis:
@rvagg The regex still looks the same to me.  Can it be that you didn't push the commit?

rvagg:
sorry, it seems that I have process.release branches on both this repo and my own fork .. pushed to the one that's behind this PR now.

```
  , bitsre = /\/win-(x86|x64)\//
  , bitsreV3 = /\/win-(x86|ia32|x64)\// // io.js v3.x.x shipped with "ia32" but should
                                        // have been "x86" 
```

bnoordhuis:
No problem and LGTM.

ralphtheninja:
> Unless there are no objections, I might merge this and release a v3 eh?

Yes! :dancers:

rvagg:
OK, two more changes. https://github.com/nodejs/node/issues/2719 reminded me that we can't trust `semver.satisfies()` any more so I've switched to checking major version numbers where needed, see bfaf7df. Also I've removed node_modules from the tree in 44dae28, I hope there are no objections to this.

mzgol:
@rvagg Argh, sorry for this suggestion, I totally forgot most semver ranges ignore pre-releases now.

mzgol:
> nodejs/node#2719 reminded me that we can't trust semver.satisfies() any more so I've switched to checking major version numbers where needed

Any chance for applying the same patch to pangyp?

Fishrock123:
> remove node_modules from tree

If we could do that elsewhere that would be nice. I'm actually curious if there are any advantages to them being in-tree for us.

rvagg:
@mzgol done and released a patch update, just note that as soon as node-gyp@3 goes out I'm using the deprecate hammer on all those versions

rvagg:
merged and published as v3.0.0, thanks for all the help with this one folks!

PR #713, opened 2015-09-05
Improve errors to show tarball url during failures

aredridel:


lib/install.js
```diff
@@ -241,14 +242,14 @@ function install (gyp, argv, callback) {
 
       req.on('close', function () {
         if (extractCount === 0) {
-          cb(new Error('Connection closed while downloading tarball file'))
+          cb(new Error('Connection closed while downloading tarball file "%s"', tarballUrl))
```
rvagg: I'm pretty sure this isn't going to work, it's not `console.log()`, you'll have to append

rvagg: I guess you mean `VError()` not `Error()`?

rvagg:
I think I'm inclined to agree with @fengmk2 @ https://github.com/rvagg/pangyp/pull/15 that the addition of a new module just for string concatenation might be overkill. verror would be nice if we were chaining errors, but you could use `+` or even `util.format()` if you want to get `console.log()` style formatting. I have no problems with the messages themselves though.

lib/install.js
aredridel: Whoops, yes!

bnoordhuis:
Merge conflicts and no activity in over a year so I'll go ahead and close this.

PR #716, opened 2015-09-08
normalise versions for target == this comparison

rvagg:
For the `node-gyp rebuild` case, everything is calculated fine for `configure` but then when it gets to `install` it's passing in an `argv[0]` with the version number and the `v` is stripped from that version but we are comparing with `process.version` which has a `v` so the naive comparison says it's not the same version and therefore we are not targetting the "default".

The solution here is to simply normalise `process.version` and match it to the already normalised target version so the comparison is fair.

Reported [here](https://github.com/nodejs/node/issues/2522#issuecomment-138447618) for 4.0.0-rc.4

### DIFF
--- a/lib/process-release.js
+++ b/lib/process-release.js
@@ -11,9 +11,9 @@ var semver = require('semver')
 // `process.version` and `process.release` where it exists.
 function processRelease (argv, gyp, defaultVersion, defaultRelease) {
   var version = argv[0] || gyp.opts.target || defaultVersion
-    , isDefaultVersion = version === defaultVersion
     , versionSemver = semver.parse(version)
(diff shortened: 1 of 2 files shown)


lib/process-release.js
```diff
@@ -28,6 +28,9 @@ function processRelease (argv, gyp, defaultVersion, defaultRelease) {
   // flatten version into String
   version = versionSemver.version
 
+  // defaultVersion should come from process.version so ought to be valid semve
```
targos: typo: semver

rvagg:
thanks @targos, fixed

rvagg:
landing to get this into 4.0.0-rc.5

PR #738, opened 2015-09-22
process.platform will never be kfreebsd

kapouer:
This platform value came from debian package, and now the value
from debian package is simply "freebsd", so this check is useless.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -21,7 +21,7 @@ exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the mod
 function build (gyp, argv, callback) {
   var release = processRelease(argv, gyp, process.version, process.release)
     , makeCommand = gyp.opts.make || process.env.MAKE
-      || (process.platform.indexOf('bsd') != -1 && process.platform.indexOf('kfreebsd') == -1 ? 'gmake' : 'make')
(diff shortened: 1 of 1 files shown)


rvagg:
Is there anything in the wild still identifying as this? do you have some docs to point to for this, I have zero context on this one.

/cc @nodejs/platform-freebsd anyone able to +1 this?

kapouer:
Oh yes !
https://github.com/nodejs/node-gyp/commit/7ec91d648750f8f85468de24812d17a8fff35c87#diff-26b681bd4dac10f25d5293c7595c435b

but now
http://anonscm.debian.org/cgit/collab-maint/nodejs.git/tree/debian/rules#n50

rvagg:
OK, if it's your addition then it can be your removal, still would appreciate a @nodejs/platform-freebsd +1 — also @kapouer, if you're a *BSD person then perhaps you'd consider being on the @nodejs/platform-freebsd group? We have special platform groups that get pinged when there is a platform-specific issue that needs expertise.

kapouer:
Sure, thanks for asking. I'd be glad being more aware of the issues.

bnoordhuis:
Not a member of @nodejs/platform-freebsd but LGTM.

For future reference, this undoes commit 7ec91d64 from 2013, also by @kapouer.

jbergstroem:
LGTM

rvagg:
101bed6, thanks

Issue #744, opened 2015-09-25
infinite loop in node-gyp configure if an empty key is present in the .npmrc config file


jmorille:
if a new line is created with a space in the userconfig file /home/[user]/.npmrc
like the ticket open in npm https://github.com/npm/npm/issues/9756

that create an empty key with the values true
```shell
# npm config list
```

```shell
; userconfig /home/[user]/.npmrc
 = true
```

the command *node-gyp configure* create an infinite loop that eat all the cpu forever.

bnoordhuis:
https://github.com/nodejs/node-gyp/pull/745 and https://codereview.chromium.org/1364373004/

PR #745, opened 2015-09-25
Fix infinite loop with zero-length options.

bnoordhuis:
The bundled copy of GYP enters an infinite loop when it encounters a
zero-length option so ensure those don't get through.

The root cause is a bad guard in gyp/pylib/gyp/input.py that doesn't
take into account that `'' in 'whatever'` evaluates to True.

Fixes: https://github.com/nodejs/node-gyp/issues/744

R=@rvagg


TooTallNate:
LGTM

bnoordhuis:
Upstream fix at https://codereview.chromium.org/1364373004/.

PR #753, opened 2015-10-02
Add support for AIX

mhdawson:
For AIX we need to use gmake

For AIX we need to set up the path to the exp file
which contains the symbols needed for linking.
The file will either be in one of the following
depeding on whether are are in installed or
development environment:
  - the include/node directory
  - the out/Release directory

The work to add the required node.exp file is being covered in this PR https://github.com/nodejs/node/pull/3114

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -19,9 +19,15 @@ var fs = require('graceful-fs')
 exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
 
 function build (gyp, argv, callback) {
+  var platformMake = 'make'
+  if (process.platform === 'aix') {
+    platformMake = 'gmake'
+  } else if (process.platform.indexOf('bsd') !== -1) {
+    platformMake = 'gmake'
+  }
+
   var release = processRelease(argv, gyp, process.version, process.release)
-    , makeCommand = gyp.opts.make || process.env.MAKE
-      || (process.platform.indexOf('bsd') != -1 ? 'gmake' : 'make')
+    , makeCommand = gyp.opts.make || process.env.MAKE || platformMake
     , command = win ? 'msbuild' : makeCommand
     , buildDir = path.resolve('build')
     , configPath = path.resolve(buildDir, 'config.gypi')
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -19,6 +19,7 @@ var fs = require('graceful-fs')
   , spawn = cp.spawn
   , execFile = cp.execFile
   , win = process.platform == 'win32'
+  , findNodeDirectory = require('./find-node-directory')
 
 exports.usage = 'Generates ' + (win ? 'MSVC project files' : 'a Makefile') + ' for the current module'
 
@@ -299,6 +300,34 @@ function configure (gyp, argv, callback) {
       argv.push('-I', config)
     })
 
+    // for AIX we need to set up the path to the exp file
+    // which contains the symbols needed for linking.
+    // The file will either be in one of the following
+    // depending on whether it is an installed or
+    // development environment:
+    //  - the include/node directory
+    //  - the out/Release directory
+    //  - the out/Debug directory
+    //  - the root directory
+    var node_exp_file = ''
+    if (process.platform === 'aix') {
+      var node_root_dir = findNodeDirectory()
+      var candidates = ['include/node/node.exp',
+                        'out/Release/node.exp',
+                        'out/Debug/node.exp',
+                        'node.exp']
+      for (var next = 0; next < candidates.length; next++) {
+         node_exp_file = path.resolve(node_root_dir, candidates[next])
+         try {
+           fs.accessSync(node_exp_file, fs.R_OK)
+           // exp file found, stop looking
+           break
+         } catch (exception) {
+           // this candidate was not found or not readable, do nothing
+         }
+      }
+    }
+
     // this logic ported from the old `gyp_addon` python file
     var gyp_script = path.resolve(__dirname, '..', 'gyp', 'gyp_main.py')
     var addon_gypi = path.resolve(__dirname, '..', 'addon.gypi')
@@ -319,6 +348,9 @@ function configure (gyp, argv, callback) {
       argv.push('-Dlibrary=shared_library')
       argv.push('-Dvisibility=default')
       argv.push('-Dnode_root_dir=' + nodeDir)
+      if (process.platform === 'aix') {
+        argv.push('-Dnode_exp_file=' + node_exp_file)
+      }
       argv.push('-Dnode_gyp_dir=' + nodeGypDir)
       argv.push('-Dnode_lib_file=' + release.name + '.lib')
       argv.push('-Dmodule_root_dir=' + process.cwd())
--- a/lib/find-node-directory.js
+++ b/lib/find-node-directory.js
@@ -1,0 +1,61 @@
+var path = require('path')
+  , log = require('npmlog')
+
+function findNodeDirectory(scriptLocation, processObj) {
+  // set dirname and process if not passed in
+  // this facilitates regression tests
+  if (scriptLocation === undefined) {
+    scriptLocation = __dirname
+  }
+  if (processObj === undefined) {
+    processObj = process
+  }
+
+  // Have a look to see what is above us, to try and work out where we are
+  npm_parent_directory = path.join(scriptLocation, '../../../..')
+  log.verbose('node-gyp root', 'npm_parent_directory is '
+              + path.basename(npm_parent_directory))
+  node_root_dir = ""
+
+  log.verbose('node-gyp root', 'Finding node root directory')
+  if (path.basename(npm_parent_directory) === 'deps') {
+    // We are in a build directory where this script lives in
+    // deps/npm/node_modules/node-gyp/lib
+    node_root_dir = path.join(npm_parent_directory, '..')
+    log.verbose('node-gyp root', 'in build directory, root = '
+                + node_root_dir)
+  } else if (path.basename(npm_parent_directory) === 'node_modules') {
+    // We are in a node install directory where this script lives in
+    // lib/node_modules/npm/node_modules/node-gyp/lib or
+    // node_modules/npm/node_modules/node-gyp/lib depending on the
+    // platform
+    if (processObj.platform === 'win32') {
+      node_root_dir = path.join(npm_parent_directory, '..')
+    } else {
+      node_root_dir = path.join(npm_parent_directory, '../..')
+    }
+    log.verbose('node-gyp root', 'in install directory, root = '
+                + node_root_dir)
+  } else {
+    // We don't know where we are, try working it out from the location
+    // of the node binary
+    var node_dir = path.dirname(processObj.execPath)
+    var directory_up = path.basename(node_dir)
+    if (directory_up === 'bin') {
+      node_root_dir = path.join(node_dir, '..')
+    } else if (directory_up === 'Release' || directory_up === 'Debug') {
+      // If we are a recently built node, and the directory structure
+      // is that of a repository. If we are on Windows then we only need
+      // to go one level up, everything else, two
+      if (processObj.platform === 'win32') {
+        node_root_dir = path.join(node_dir, '..')
+      } else {
+        node_root_dir = path.join(node_dir, '../..')
+      }
+    }
+    // Else return the default blank, "".
+  }
+  return node_root_dir
+}
+
+module.exports = findNodeDirectory
--- a/test/test-find-node-directory.js
+++ b/test/test-find-node-directory.js
@@ -1,0 +1,115 @@
+var test = require('tape')
+var path = require('path')
+var findNodeDirectory = require('../lib/find-node-directory')
+
+var platforms = ['darwin', 'freebsd', 'linux', 'sunos', 'win32', 'aix']
+
+// we should find the directory based on the directory
+// the script is running in and it should match the layout
+// in a build tree where npm is installed in
+// .... /deps/npm
+test('test find-node-directory - node install', function (t) {
+  t.plan(platforms.length)
+  for (var next = 0; next < platforms.length; next++) {
+    var processObj = {execPath: '/x/y/bin/node', platform: platforms[next]}
+    t.equal(
+      findNodeDirectory('/x/deps/npm/node_modules/node-gyp/lib', processObj),
+                        path.join('/x'))
+  }
+})
+
+// we should find the directory based on the directory
+// the script is running in and it should match the layout
+// in an installed tree where npm is installed in
+// .... /lib/node_modules/npm or .../node_modules/npm
+// depending on the patform
+test('test find-node-directory - node build', function (t) {
+  t.plan(platforms.length)
+  for (var next = 0; next < platforms.length; next++) {
+    var processObj = {execPath: '/x/y/bin/node', platform: platforms[next]}
+    if (platforms[next] === 'win32') {
+      t.equal(
+        findNodeDirectory('/y/node_modules/npm/node_modules/node-gyp/lib',
+                           processObj), path.join('/y'))
+    } else {
+      t.equal(
+        findNodeDirectory('/y/lib/node_modules/npm/node_modules/node-gyp/lib',
+                           processObj), path.join('/y'))
+    }
+  }
+})
+
+// we should find the directory based on the execPath
+// for node and match because it was in the bin directory
+test('test find-node-directory - node in bin directory', function (t) {
+  t.plan(platforms.length)
+  for (var next = 0; next < platforms.length; next++) {
+    var processObj = {execPath: '/x/y/bin/node', platform: platforms[next]}
+    t.equal(
+      findNodeDirectory('/nothere/npm/node_modules/node-gyp/lib', processObj),
+      path.join('/x/y'))
+  }
+})
+
+// we should find the directory based on the execPath
+// for node and match because it was in the Release directory
+test('test find-node-directory - node in build release dir', function (t) {
+  t.plan(platforms.length)
+  for (var next = 0; next < platforms.length; next++) {
+    var processObj
+    if (platforms[next] === 'win32') {
+      processObj = {execPath: '/x/y/Release/node', platform: platforms[next]}
+    } else {
+      processObj = {execPath: '/x/y/out/Release/node',
+                    platform: platforms[next]}
+    }
+
+    t.equal(
+      findNodeDirectory('/nothere/npm/node_modules/node-gyp/lib', processObj),
+                        path.join('/x/y'))
+  }
+})
+
+// we should find the directory based on the execPath
+// for node and match because it was in the Debug directory
+test('test find-node-directory - node in Debug release dir', function (t) {
+  t.plan(platforms.length)
+  for (var next = 0; next < platforms.length; next++) {
+    var processObj
+    if (platforms[next] === 'win32') {
+      processObj = {execPath: '/a/b/Debug/node', platform: platforms[next]}
+    } else {
+      processObj = {execPath: '/a/b/out/Debug/node', platform: platforms[next]}
+    }
+
+    t.equal(
+      findNodeDirectory('/nothere/npm/node_modules/node-gyp/lib', processObj),
+                        path.join('/a/b'))
+  }
+})
+
+// we should not find it as it will not match based on the execPath nor
+// the directory from which the script is running
+test('test find-node-directory - not found', function (t) {
+  t.plan(platforms.length)
+  for (var next = 0; next < platforms.length; next++) {
+    var processObj = {execPath: '/x/y/z/y', platform:next}
+    t.equal(findNodeDirectory('/a/b/c/d', processObj), '')
+  }
+})
+
+// we should find the directory based on the directory
+// the script is running in and it should match the layout
+// in a build tree where npm is installed in
+// .... /deps/npm
+// same test as above but make sure additional directory entries
+// don't cause an issue
+test('test find-node-directory - node install', function (t) {
+  t.plan(platforms.length)
+  for (var next = 0; next < platforms.length; next++) {
+    var processObj = {execPath: '/x/y/bin/node', platform: platforms[next]}
+    t.equal(
+      findNodeDirectory('/x/y/z/a/b/c/deps/npm/node_modules/node-gyp/lib',
+                        processObj), path.join('/x/y/z/a/b/c'))
+  }
+})
--- a/addon.gypi
+++ b/addon.gypi
@@ -65,6 +65,11 @@
           'DYLIB_INSTALL_NAME_BASE': '@rpath'
         },
       }],
+      [ 'OS=="aix"', {
+        'ldflags': [
+          '-Wl,-bimport:<(node_exp_file)'
+        ],
+      }],
       [ 'OS=="win"', {
         'libraries': [
           '-lkernel32.lib',


mhdawson:
I've tested this in a build of Node from master.  I've looked through the readme for node-gyp to see how changes are normally tested but I've not found any instructions. Can somebody point me at were it is documented what testing should be done ?

lib/build.js
```diff
@@ -21,6 +21,7 @@ exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the mod
 function build (gyp, argv, callback) {
   var release = processRelease(argv, gyp, process.version, process.release)
     , makeCommand = gyp.opts.make || process.env.MAKE
+      || (process.platform.indexOf('aix') != -1 && 'gmake')
```
rvagg: the mixed use of `&&` and the ternary operator below makes this a bit confusing, the duplicate resolving of `gmake` suggests redundancy, and, shouldn't it just be `aix` when it's AIX? the `indexOf` is for the *bsd variants I _think_.

Can I suggest squashing both these lines to: `((/(^aix$)|bsd/).test(process.platform) ? 'gmake' : 'make')`

lib/configure.js
```diff
@@ -161,6 +161,42 @@ function configure (gyp, argv, callback) {
     }
   }
 
+  function findNodeRootDirectory() {
+    // Have a look to see what is above us, to try and work out where we are
+    npm_parent_directory = path.join(__dirname, '..', '..','..','..')
```
rvagg: `path.join(__dirname, '../../../..')` should be fine here

lib/configure.js
```diff
@@ -161,6 +161,42 @@ function configure (gyp, argv, callback) {
...
+    log.verbose('node-gyp root', 'npm_parent_directory is ' + path.basename(npm_parent_directory))
+    node_root_dir = ""
+
+    log.verbose('node-gyp root','Finding node root directory')
```
rvagg: space after `,`

lib/configure.js
```diff
@@ -161,6 +161,42 @@ function configure (gyp, argv, callback) {
...
+    if (path.basename(npm_parent_directory) == 'lib') {
+      // We are in a node install directory
+      node_root_dir = path.join(npm_parent_directory, '..', '..')
```
rvagg: `'../..'`

lib/configure.js
```diff
@@ -161,6 +161,42 @@ function configure (gyp, argv, callback) {
...
+      log.verbose('node-gyp root', 'in install directory, root = ' + node_root_dir)
+    } else if (path.basename(npm_parent_directory) == 'node_modules') {
+      // We are in a node build directory, where npm lives under lib/
+      node_root_dir = path.join(npm_parent_directory, '..', '..', 'include', 'node')
```
rvagg: `path.join(npm_parent_directory, '../../include/node')`

lib/configure.js
```diff
@@ -161,6 +161,42 @@ function configure (gyp, argv, callback) {
...
+      var directory_up = path.basename(node_dir)
+      if (directory_up == 'bin') {
+        node_root_dir  = path.join(node_dir, '..')
+      } else if (directory_up == 'Release' || directory_up == 'Debug') {
+        // If we are a recently built node, and the directory structure is that of a repository
+        // If we are on Windows then we only need to go one level up, everything else, two
+        if (process.platform == 'win32') {
+          node_root_dir = path.join(node_dir, '..')
+        } else {
+          node_root_dir = path.join(node_dir, '..', '..')
```
rvagg: '../..`

lib/configure.js
```diff
@@ -319,6 +374,9 @@ function configure (gyp, argv, callback) {
       argv.push('-Dlibrary=shared_library')
       argv.push('-Dvisibility=default')
       argv.push('-Dnode_root_dir=' + nodeDir)
+      if (process.platform == 'aix') {
+        argv.push('-Dnode_exp_file=' + node_exp_file )
```
rvagg: remove space before `)`

lib/configure.js
```diff
@@ -299,6 +335,25 @@ function configure (gyp, argv, callback) {
...
     })
 
+    // for AIX we need to set up the path to the exp file
+    // which contains the symbols needed for linking.
+    // The file will either be in one of the following
+    // depeding on whether are are in installed or
+    // development environment:
+    //  - the include/node directory
+    //  - the out/Release directory
+    var node_exp_file = "";
```
rvagg: single quotes, no semicolon `= ''`

lib/configure.js
```diff
@@ -299,6 +335,25 @@ function configure (gyp, argv, callback) {
...
+    if (process.platform == 'aix') {
+      var node_root_dir = findNodeRootDirectory();
```
rvagg: no semicolon

lib/configure.js
```diff
@@ -299,6 +335,25 @@ function configure (gyp, argv, callback) {
...
+      node_exp_file = path.resolve(node_root_dir, 'include/node/node.exp')
```
rvagg: can you confirm these are included in the header tarballs created upstream with your recent changes?

rvagg:
I'm pretty sure the `findNodeRootDirectory()` approach is wrong here and makes too many unsafe assumptions. Can you have a look at how it finds `config.gypi` and emulate that instead? I understand the problem of finding node.exp when using `--nodedir` but that's only a minor case, the most important one being finding the file in `~/.node-gyp/version/include/node/node.exp`.

rvagg:
re tests, there isn't much in the way of tests here. There are some docker tests you can run in test/docker.sh but obviously only for Linux but it's worth making sure you haven't broken anything else. Other testing has to be manual, which can be quite tricky when we haven't got any headers tarballs on nodejs.org that you can trial with! Once we get an AIX machine (or two, or three, whatever) in CI then we could produce tarballs for releases || nightlies with node.exp in it and test with that.

But, that raises another question, is node.exp specific to the arch used? Does it need a new one for BE, LE, 32-bit, 64-bit? If so, then it can't go into the headers tarball and you're going to need to scrap a lot of this code and emulate the node.lib download process we have for Windows and we're going to need to set up new directories in nodejs.org/download/ to contain node.exp for all of the different architectures.

rvagg:
OK, another rethink on this, backing up on what I've just said I realise your approach now is to find `node.exp` as it's been installed on the system along with `node` from a binary tarball or some other official package, not necessarily have it in the headers or download from nodejs.org like `node.lib`. This is something that many of us have talked about for some time now wrt general headers because they are installed on many systems but node-gyp will always download them again anyway. A disparity that's existed since 0.10.

In that case, perhaps this is the right approach, _but_ I personally don't want to see this logic baked in to node-gyp and would rather see it split off to a separate module that can locate the Node `include` dir on any system if it exists. That way it can be independently tested and iterated on rather than becoming part of the big logic puzzle that node-gyp currently is. I know that's not the expedient path here but as it is there's no clear way to test this logic in here. Perhaps a middle ground is to do what I did with `process-release.js` (which I would also like to extract to a separate package) and treat it as independent and test it independently.

lib/build.js
bnoordhuis: Maybe platform-specific overrides are better done in parseArgv / parseOpts in lib/node-gyp.js.

lib/configure.js
```diff
@@ -299,6 +335,25 @@ function configure (gyp, argv, callback) {
...
+      if (!fs.existsSync(node_exp_file)) {
```
bnoordhuis: `fs.existsSync()` emits deprecation warnings.

lib/configure.js
```diff
@@ -299,6 +335,25 @@ function configure (gyp, argv, callback) {
...
+        node_exp_file = path.resolve(node_root_dir, 'out/Release/node.exp')
```
bnoordhuis: What about debug builds?

mhdawson:
Sorry for taking so long to get back to this one.  First part in Node is now in so will be moving on to this one soon.  I am out at a conference this week but it will my priority after that.

lib/configure.js
mhdawson: Good point, will add line that looks in out/Debug  as well

lib/configure.js
mhdawson: Noted will replace with one of the suggested alternates

lib/build.js
mhdawson: it is kind of of ugly.  What do you think of this. It's more verbose but also more obvious whats going on 

function build (gyp, argv, callback) {
  var platformMake = 'make';
  if (process.platform == 'aix') {
    platformMake = 'gmake';
  } else if (process.platform.indexOf('bsd') != -1) {
    platformMake = 'gmake');
  }

  var release = processRelease(argv, gyp, process.version, process.release)
    , makeCommand = gyp.opts.make || process.env.MAKE || platformMake
    , command = win ? 'msbuild' : makeCommand
    , buildDir = path.resolve('build')
    , configPath = path.resolve(buildDir, 'config.gypi')
    , jobs = gyp.opts.jobs || process.env.JOBS
    , buildType
    , config
    , arch
    , nodeDir
    , copyDevLib

lib/configure.js
mhdawson: ok

lib/configure.js
mhdawson: k

lib/configure.js
mhdawson: k

lib/configure.js
mhdawson: k

lib/configure.js
mhdawson: k

lib/configure.js
mhdawson: k

lib/configure.js
mhdawson: k

lib/configure.js
mhdawson: k

mhdawson:
Have updates to reflect the comments in my local repo, just need to test now

mhdawson:
@bnoordhuis  @rvagg  updated to address review comments

lib/build.js
```diff
@@ -19,9 +19,15 @@ var fs = require('graceful-fs')
 exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
 
 function build (gyp, argv, callback) {
+  var platformMake = 'make';
+  if (process.platform == 'aix') {
+    platformMake = 'gmake';
+  } else if (process.platform.indexOf('bsd') != -1) {
+    platformMake = 'gmake';
```
bnoordhuis: Style nits: no trailing semicolons (for better or worse)

lib/configure.js
```diff
@@ -299,6 +300,34 @@ function configure (gyp, argv, callback) {
...
+    //  - the out/Debug directory
+    //  - the root directory
+    var node_exp_file = ''
+    if (process.platform == 'aix') {
+      var node_root_dir = findNodeDirectory()
+      var candidates = ['include/node/node.exp',
+                        'out/Release/node.exp',
+                        'out/Debug/node.exp',
+                        'node.exp']
+      for(var next in candidates) {
```
bnoordhuis: Style nit: space after `for`, ditto for the `catch` below.

lib/find-node-directory.js
```diff
@@ -0,0 +1,40 @@
...
+    // We are in a node build directory, where npm lives under lib/
+    node_root_dir = path.join(npm_parent_directory, '../../include/node')
+    log.verbose('node-gyp root', 'in build directory, root = ' + node_root_dir)
+  } else {
+    // We don't know where we are, try working it out from the location
+    // of the node binary
+    var node_dir = path.dirname(process.execPath)
+    var directory_up = path.basename(node_dir)
+    if (directory_up == 'bin') {
+      node_root_dir  = path.join(node_dir, '..')
```
bnoordhuis: Style nit: one space too many

lib/find-node-directory.js
```diff
@@ -0,0 +1,40 @@
...
+
+function findNodeDirectory() {
+  // Have a look to see what is above us, to try and work out where we are
+  npm_parent_directory = path.join(__dirname, '../../../..')
+  log.verbose('node-gyp root', 'npm_parent_directory is ' + path.basename(npm_parent_directory))
+  node_root_dir = ""
+
+  log.verbose('node-gyp root', 'Finding node root directory')
+  if (path.basename(npm_parent_directory) == 'lib') {
+    // We are in a node install directory
```
bnoordhuis: I'm not sure I understand when this would happen.   With a path like `/opt/node/lib/node_modules/npm/node_modules/node-gyp/lib`, you arrive at `/opt/node/lib/node_modules`, not `/opt/node/lib`.

lib/find-node-directory.js
```diff
@@ -0,0 +1,40 @@
...
+      if (process.platform == 'win32') {
+        node_root_dir = path.join(node_dir, '..')
+      } else {
+        node_root_dir = path.join(node_dir, '../..')
+      }
+    }
+    // Else return the default blank, "".
+  }
+  return node_root_dir
+}
```
bnoordhuis: Can you add a few regression tests for this function?  I'd make it so that you can plug in a mock `__dirname` and `process` object.

lib/find-node-directory.js
```diff
@@ -0,0 +1,40 @@
+var path = require('path')
+  , log = require('npmlog')
+
+function findNodeDirectory() {
+  // Have a look to see what is above us, to try and work out where we are
+  npm_parent_directory = path.join(__dirname, '../../../..')
+  log.verbose('node-gyp root', 'npm_parent_directory is ' + path.basename(npm_parent_directory))
```
bnoordhuis: Style nit: can you try to keep lines <= 80 columns?  I know node-gyp isn't very consistent in that respect but I'd like new code to not have overly long lines.

mhdawson: my bad, need to be more diligent about that

lib/find-node-directory.js
mhdawson: From the history I can gather the function to get the Node directory used to exist somewhere else as well but no longer does.  Looking at the potential installation configurations I agree that 'lib' does not make sense.  The cases I've been testing include when you build from source and have that layout and the case were you install from the tar.gz.  From those I think we should be checking 2 cases. One where the parent is "node_modules" (which is covered) and one where the parent is "deps". I'll update and add more comments to try to make it clear which cases are being covered.

lib/build.js
mhdawson: k

lib/configure.js
mhdawson: k

mhdawson:
@bnoordhuis updated to address your latest comments. Added the unit tests and took a closer look as you had found a place were the code was obviously stale.  I fixed up the initial checks to be for 'node_modules' and 'deps' and validated directories on windows and fixed a few issues there as well.

test/test-find-node-directory.js
```diff
@@ -0,0 +1,98 @@
...
+  var processObj = {execPath: '/x/y/bin/node', platform: process.platform}
+  if (process.platform === 'win32') {
+    t.equal(
+      findNodeDirectory('/y/node_modules/npm/node_modules/node-gyp/lib',
+                         processObj), path.join('/y'));
+  } else {
+    t.equal(
+      findNodeDirectory('/y/lib/node_modules/npm/node_modules/node-gyp/lib',
+                         processObj), path.join('/y'));
+  }
```
bnoordhuis: Can you test both branches here?  I.e., rewrite it so it's called once with `processObj.platform === 'win32'` and once without?

bnoordhuis: I'd recommend doing that everywhere where you call findNodeDirectory() with `{ platform: process.platform }`.  The idea is that all code paths are tested on all platforms.

lib/build.js
```diff
@@ -19,9 +19,15 @@ var fs = require('graceful-fs')
 exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
 
 function build (gyp, argv, callback) {
+  var platformMake = 'make'
+  if (process.platform === 'aix') {
+    platformMake = 'gmake'
+  } else if (process.platform.indexOf('bsd') != -1) {
```
bnoordhuis: Can you change this to `!==` while you're here?

lib/configure.js
```diff
@@ -299,6 +300,34 @@ function configure (gyp, argv, callback) {
...
+                        'out/Release/node.exp',
+                        'out/Debug/node.exp',
+                        'node.exp']
+      for (var next in candidates) {
+         node_exp_file = path.resolve(node_root_dir, candidates[next])
+         try {
+           fs.accessSync(node_exp_file, fs.R_OK)
+           // exp file found, stop looking
+           break
+         } catch(exception) {
```
bnoordhuis: Micro-nit: space after `catch`.

test/test-find-node-directory.js
```diff
@@ -0,0 +1,98 @@
...
+// we should find the directory based on the directory
+// the script is running in and it should match the layout
+// in a build tree where npm is installed in
+// .... /deps/npm
+test('test find-node-directory - node install', function (t) {
+  t.plan(1)
+  var processObj = {execPath: '/x/y/bin/node', platform: process.platform}
+  t.equal(
+    findNodeDirectory('/x/deps/npm/node_modules/node-gyp/lib', processObj),
+                      path.join('/x'));
```
bnoordhuis: There are a few stray semicolons in this file.

lib/find-node-directory.js
```diff
@@ -0,0 +1,61 @@
...
+    // lib/node_modules/npm/node_modules/node-gyp/lib or
+    // node_modules/npm/node_modules/node-gyp/lib depending on the
+    // platform
+    if (processObj.platform === 'win32') {
+      node_root_dir = path.join(npm_parent_directory, '..')
+    } else {
+      node_root_dir = path.join(npm_parent_directory, '../..')
+    }
+    log.verbose('node-gyp root', 'in install directory, root = '
+    + node_root_dir)
```
thefourtheye: Minor nit: Shouldn't the `+` be on the previous line?

lib/build.js
mhdawson: sure

lib/configure.js
mhdawson: k

test/test-find-node-directory.js
mhdawson: Thanks its hard to stomp out that habit.

lib/find-node-directory.js
mhdawson: Is there a standard for this ?   I should at least line up with the start of the variables like this:

log.verbose('node-gyp root', 'in install directory, root = '
                  + node_root_dir)

like I have done elsewhere.  I'll make that change

mhdawson:
@bnoordhuis @thefourtheye updated to address latest comments

bnoordhuis:
LGTM.  @thefourtheye WDYT?

lib/configure.js
```diff
@@ -299,6 +300,34 @@ function configure (gyp, argv, callback) {
       argv.push('-I', config)
     })
 
+    // for AIX we need to set up the path to the exp file
+    // which contains the symbols needed for linking.
+    // The file will either be in one of the following
+    // depeding on whether are are in installed or
```
thefourtheye: depending on whether it is

lib/configure.js
```diff
@@ -299,6 +300,34 @@ function configure (gyp, argv, callback) {
...
+    //  - the out/Debug directory
+    //  - the root directory
+    var node_exp_file = ''
+    if (process.platform === 'aix') {
+      var node_root_dir = findNodeDirectory()
+      var candidates = ['include/node/node.exp',
+                        'out/Release/node.exp',
+                        'out/Debug/node.exp',
+                        'node.exp']
+      for (var next in candidates) {
```
thefourtheye: Iterating an array with `for...in` is generally not advisable.

lib/configure.js
```diff
@@ -319,6 +348,9 @@ function configure (gyp, argv, callback) {
       argv.push('-Dlibrary=shared_library')
       argv.push('-Dvisibility=default')
       argv.push('-Dnode_root_dir=' + nodeDir)
+      if (process.platform === 'aix') {
+        argv.push('-Dnode_exp_file=' + node_exp_file)
```
thefourtheye: If the file is not found, should we still add this flag?

test/test-find-node-directory.js
```diff
@@ -0,0 +1,115 @@
...
+// the script is running in and it should match the layout
+// in a build tree where npm is installed in
+// .... /deps/npm
+test('test find-node-directory - node install', function (t) {
+  t.plan(platforms.length)
+  for (currentPlatform in platforms) {
+    var processObj = {execPath: '/x/y/bin/node', platform: currentPlatform}
+    t.equal(
+      findNodeDirectory('/x/deps/npm/node_modules/node-gyp/lib', processObj),
+                        path.join('/x'))
```
thefourtheye: `path.join` is used all over the tests but it actually doesn't join paths. Am I totally missing the point of its usage?

test/test-find-node-directory.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+})
+
+// we should find the directory based on the directory
+// the script is running in and it should match the layout
+// in a installed tree where npm is installed in
```
thefourtheye: minor nit: `in an installed `

test/test-find-node-directory.js
```diff
@@ -0,0 +1,115 @@
...
+      processObj = {execPath: '/a/b/out/Debug/node', platform: currentPlatform}
+    }
+
+    t.equal(
+      findNodeDirectory('/nothere/npm/node_modules/node-gyp/lib', processObj),
+                        path.join('/a/b'))
+  }
+})
+
+// we shoudl not find it as it will not match based on the execPath nor
```
thefourtheye: should

thefourtheye:
Have few minor comments but overall change LGTM

lib/configure.js
bnoordhuis: Doesn't really matter for a test though.

lib/configure.js
mhdawson: fixed

lib/configure.js
mhdawson: I'd prefer that we still add the flag so that we can see what occurred.  In verbose mode we'll be able to see what the value for -Dnode_exp_file is

test/test-find-node-directory.js
mhdawson: without the join the patch ends up being incorrect on windows, the join fixes up the '/' to be appropriate for the platform

test/test-find-node-directory.js
mhdawson: fixed

lib/configure.js
thefourtheye: But this is not a test :O

lib/configure.js
thefourtheye: Oh okay then

test/test-find-node-directory.js
mhdawson: fixed

test/test-find-node-directory.js
thefourtheye: But you are hardcoding them right? I need to check if the join changes the slashes

lib/configure.js
mhdawson: Changed here and in the test files

mhdawson:
@thefourtheye addressed your comments.  If it looks ok to you I can squash into one commit

lib/configure.js
```diff
@@ -299,6 +300,34 @@ function configure (gyp, argv, callback) {
...
+                        'out/Release/node.exp',
+                        'out/Debug/node.exp',
+                        'node.exp']
+      for (var next = 0; next < candidates.length; next++) {
+         node_exp_file = path.resolve(node_root_dir, candidates[next])
+         try {
+           fs.accessSync(node_exp_file, fs.R_OK)
+           // exp file found, stop looking
+           break
+         } catch (exception) {
```
thefourtheye: If the file is not found anywhere, `node_exp_file` will still be `node.exp` at the end of the loop. Is that okay or you want to reset to empty string here?

mhdawson: I think its ok as if this does occur I don't think its going to be any clearer what the problem is with it set to an empty string.

thefourtheye: Cool, let's land it then. LGTM

mhdawson:
Squashed down to a single commit

bnoordhuis:
Thanks Michael, landed in 9bfa087.

PR #756, opened 2015-10-04
Add test for python executable search logic.

bnoordhuis:
Break out the search logic into a separate function and add a regression
test.

References: https://github.com/nodejs/node-gyp/pull/668

R=@rvagg?  /cc @refack

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -1,4 +1,5 @@
 module.exports = exports = configure
+module.exports.test = { findPython: findPython }
 
 /**
  * Module dependencies.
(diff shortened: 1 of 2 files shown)


lib/configure.js
```diff
@@ -360,3 +278,97 @@ function configure (gyp, argv, callback) {
...
+          return checkPython()
+        }
+        if (win) {
+          guessPython()
+        } else {
+          failNoPython()
+        }
+      } else {
+        log.verbose('`which` succeeded', python, execPath)
+        checkPythonVersion()
```
refack: you're missing the critical fix of using the `which`'s result
```
       python = execPath
```

bnoordhuis: Well, yes, that's intentional.  Mixing a general refactor and bug fixes in a single commit is bad style.

refack: :+1: got it.

bnoordhuis:
Maybe R=@TooTallNate?

lib/configure.js
```diff
@@ -360,3 +278,97 @@ function configure (gyp, argv, callback) {
...
+
+  function checkPythonVersion () {
+    var env = extend({}, process.env)
+    env.TERM = 'dumb'
+
+    execFile(python, ['-c', 'import platform; print(platform.python_version());'], { env: env }, function (err, stdout) {
+      if (err) {
+        return callback(err)
+      }
+      log.verbose('check python version', '`%s -c "import platform; print(platform.python_version());"` returned: %j', python, stdout)
```
thefourtheye: I guess we don't have the Max length linter rule.

test/test-find-python.js
```diff
@@ -0,0 +1,20 @@
...
+var configure = require('../lib/configure')
+var execFile = require('child_process').execFile
+
+test('find python executable', function (t) {
+  t.plan(4)
+
+  configure.test.findPython('python', function (err, found) {
+    t.equal(err, null)
+    var proc = execFile(found, ['-V'], function (err, stdout, stderr) {
+      t.equal(err, null)
```
thefourtheye: Wouldn't strict equal be better here?

test/test-find-python.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+var test = require('tape')
+var configure = require('../lib/configure')
+var execFile = require('child_process').execFile
+
+test('find python executable', function (t) {
+  t.plan(4)
+
+  configure.test.findPython('python', function (err, found) {
```
thefourtheye: `found` is unused?

lib/configure.js
```diff
@@ -360,3 +278,97 @@ function configure (gyp, argv, callback) {
...
+      var version = stdout.trim()
+      if (~version.indexOf('+')) {
+        log.silly('stripping "+" sign(s) from version')
+        version = version.replace(/\+/g, '')
+      }
+      if (~version.indexOf('rc')) {
+        log.silly('stripping "rc" identifier from version')
+        version = version.replace(/rc(.*)$/ig, '')
```
thefourtheye: Here case is ignored but indexOf check little above doesn't

test/test-find-python.js
bnoordhuis: See the execFile call two lines down. :-)

test/test-find-python.js
bnoordhuis: I guess so.  I'll change that.

lib/configure.js
```diff
@@ -360,3 +278,97 @@ function configure (gyp, argv, callback) {
...
+        callback(null, python)
+      } else {
+        failPythonVersion(version)
+      }
+    })
+  }
+
+  function failNoPython () {
+    callback(new Error('Can\'t find Python executable "' + python +
+          '", you can set the PYTHON env variable.'))
```
thefourtheye: Shouldn't this be aligned to the first parameter?

lib/configure.js
bnoordhuis: I agree it's quirky but please keep in mind that I only moved this code around, I didn't edit any of it.

lib/configure.js
```diff
@@ -1,4 +1,5 @@
 module.exports = exports = configure
+module.exports.test = { findPython: findPython }
```
thefourtheye: Why not directly attach it on the `configure` itself?

test/test-find-python.js
thefourtheye: My bad.

lib/configure.js
bnoordhuis: To make it clear that it's not part of the official API but that it's only exported for testing.

thefourtheye: Oh okay then.

lib/configure.js
thefourtheye: Ah I see it now. Thanks for pointing that out :)

bnoordhuis:
@thefourtheye What was the verdict on this PR?

rvagg:
I don't have time right now to review this but fwiw I extracted this logic into a separate package a while back: https://github.com/rvagg/check-python and figured one day I'd come back here and PR a full replacement. The logging is the tricky bit, however.

lib/configure.js
```diff
@@ -360,3 +278,97 @@ function configure (gyp, argv, callback) {
...
+      if (err) {
+        log.verbose('`which` failed', python, err)
+        if (python === 'python2') {
+          python = 'python'
+          return checkPython()
+        }
+        if (win) {
+          guessPython()
+        } else {
+          failNoPython()
```
thefourtheye: After the refactoring, `failNoPython` directly calls the `configure`'s `callback` and brings the execution to end. But ideally it should have called `findPython`'s `callback` and that should have propagated the error back to the caller of `configure`. This will still work but looks like bad design to me. Please point out if I am missing something obvious.

thefourtheye:
LGTM with a suggestion/question.

lib/configure.js
bnoordhuis: But it does, doesn't it?  It calls the `callback` from `findPython()`'s arguments.

thefourtheye: @bnoordhuis I am sorry. You are correct. As the change was longer, I was not able to view it properly I guess. LGTM.

bnoordhuis:
Landed in 817ed9b with the `s/equal/strictEqual/` fix.

PR #781, opened 2015-10-22
update gyp to b3cef02

iWuzHere:
Includes two patches for AIX. Adds support for both 32-bit and 64-bit
files[0] and uses -RPf for cp instead of -af (which is unsupported)[1]

[0] https://codereview.chromium.org/1319663007
[1] https://codereview.chromium.org/1368133002

CI run here https://ci.nodejs.org/job/node-test-commit/882/
Intermittent failure that passed on rerun:
https://ci.nodejs.org/job/node-test-commit-arm/958/

### DIFF
--- a/gyp/PRESUBMIT.py
+++ b/gyp/PRESUBMIT.py
@@ -125,15 +125,13 @@ def CheckChangeOnCommit(input_api, output_api):
 
 
 TRYBOTS = [
-    'gyp-win32',
-    'gyp-win64',
-    'gyp-linux',
-    'gyp-mac',
-    'gyp-android'
+    'linux_try',
+    'mac_try',
+    'win_try',
 ]
 
 
 def GetPreferredTryMasters(_, change):
   return {
-      'tryserver.nacl': { t: set(['defaulttests']) for t in TRYBOTS },
+      'client.gyp': { t: set(['defaulttests']) for t in TRYBOTS },
   }
--- a/gyp/buildbot/buildbot_run.py
+++ b/gyp/buildbot/buildbot_run.py
@@ -3,27 +3,17 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-
 """Argument-less script to select what to run on the buildbots."""
 
-
-import filecmp
 import os
 import shutil
 import subprocess
 import sys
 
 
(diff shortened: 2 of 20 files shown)


iWuzHere:
Note this also removes support for android as gyp has abandoned it. If you still want to keep android support in the code I can just pull the two AIX changes

mhdawson:
@bnoordhuis  can you take a look at this one

mhdawson:
@bnoordhuis, there were 2 PRs to update gyp if you can also look at this one it would be great.

@shigeki, since you commented on the other one giving you an FYI here as well

shigeki:
FYI: Currently, a floating patch of https://github.com/nodejs/node/commit/58e914f9bc866fe424aa38b5e9ac42e6216d5f30 is applied to tools/gyp in node so as to avoid a build error occured in MacOS without XCode.  Ben has already submitted the patch to upstream in https://codereview.chromium.org/1213123002/ but it's not fixed yet. 
The patch is needed to work in CI where only command line tools are installed but XCode is not.
I think it is a very limited environments and I'm not sure how much it is affected in using node-gyp.

bnoordhuis:
@iWuzHere Before I start reviewing this: node-gyp's copy of gyp was floating a few patches of its own.  Did you preserve those?

iWuzHere:
I have added the floating patches.

bnoordhuis:
I think you're missing nodejs/node-gyp@39a06bc.

iWuzHere:
I shouldn't easy_xml.py hasnt changed in https://github.com/iWuzHere/node-gyp/commit/2f1d5efa1c747b1c0c7d5d721b25942d283606f9

bnoordhuis:
LGTM but can you update the commit log with a quick overview of changes/additions?

iWuzHere:
I have updated the commit message

bnoordhuis:
Thanks, landed in f5d86eb...9049241.

@rvagg @TooTallNate Maybe one of you can do a release for this?

mhdawson:
@bnoordhuis when you say to a release is that how it gets pulled into the dependencies within Node itself  ?

bnoordhuis:
Sadly, no.  The chain of releases is node-gyp -> npm -> node.

PR #801, opened 2015-11-06
Updated the links for gyp user doc and input ref in README

jasoncoelho:
The links to the gyp user documentation and input reference were no longer valid, and I updated the links to the new location.


jasoncoelho:
Just noticed that this was a duplicate pull request.

PR #825, opened 2015-11-26
Fixes for when Python is installed in MSYS environment

ashleysommer:
Allow node-gyp to run in Windows when the python executable is installed in an MSYS environment.

Fixes are mainly to do with problems when handling path strings, and when adding relative file paths to the MSBuild project file.


bnoordhuis:
Can you send the changes to the upstream GYP project first?  I'm happy to cherry-pick the patch once it lands upstream.

bnoordhuis:
Seeing how there has been no follow-up, I'll go ahead and close this.

PR #826, opened 2015-11-26
MSBuild option 'TreatWarningsAsErrors' added.

alex--m:


bnoordhuis:
Can you send this to the upstream GYP project first?  I can cherry-pick the patch once it lands upstream.

bnoordhuis:
I'll close, no follow-up.

Issue #652, opened 2015-06-25
iojs 2.3.0 name 'openssl_fips' is not defined"


BlueWinds:
While iojs 2.2.1 has been working fine, I'm unable to install any node-gyp configured modules with --target=2.3.0
```
$ npm install --target=2.3.0
\
> usage@0.7.0 install /.../node_modules/usage
> node-gyp rebuild

gyp: name 'openssl_fips' is not defined while evaluating condition 'openssl_fips != ""' in binding.gyp while trying to load binding.gyp
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:343:16)
gyp ERR! stack     at emitTwo (events.js:87:13)
gyp ERR! stack     at ChildProcess.emit (events.js:172:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:199:12)
gyp ERR! System Darwin 14.1.0
gyp ERR! command "/usr/local/bin/iojs" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /.../node_modules/usage
gyp ERR! node -v v2.2.1
gyp ERR! node-gyp -v v1.0.3
gyp ERR! not ok 

```

This is reproducible on both mac and a debian box. Just in case it's relevant or helpful:

```
$ for FLAG in cflags-only-I libs-only-L libs-only-l; do pkg-config --$FLAG openssl; done

-lssl -lcrypto -lz
```

Fishrock123:
@BlueWinds what's your binding.gyp look like?

Fishrock123:
@BlueWinds does this still happen with a new version of io.js?

Fishrock123:
Works for me on OS X:

```
Jeremiahs-MacBook-Pro:here-be-dragons Jeremiah$ npm -v
2.11.1
Jeremiahs-MacBook-Pro:here-be-dragons Jeremiah$ node -v
v2.3.0
Jeremiahs-MacBook-Pro:here-be-dragons Jeremiah$ npm i usage
/
> usage@0.7.0 install /Users/Jeremiah/node_modules/usage
> node-gyp rebuild

  CXX(target) Release/obj.target/sysinfo/src/binding.o
[...]

PR #833, opened 2015-12-05
configure: Add params for inherit defaults config

medns:
By default, Configure don't inherit the "defaults" from
node's `process.config` object. But sometimes it need to be
inherit from node's compile params.

So add `process.env.GYP_INHERIT_DEFAULTS` environment variable
and `--enable-inherit` arguments to control it.


bnoordhuis:
I don't think that's a good idea.  Using `process.config` from the node binary that runs node-gyp instead of the binary that loads the add-on leads to issues, see https://github.com/nodejs/node-gyp/issues/652 for an example.

rvagg:
not a fan of magic env vars either, closing due to staleness

PR #837, opened 2015-12-09
Add --cafile command line option.

bnoordhuis:
R=@rvagg?


lib/install.js
```diff
@@ -444,3 +407,60 @@ function install (gyp, argv, callback) {
   }
 
 }
+
+function download (gyp, env, url, cb) {
```
rvagg: taking a cb option that's only used when erroring is a bit gross, perhaps it'd closer to idiomatic if the thrown error bubbled up and the try/catch was done at the `download()` call?

rvagg:
A bit smelly, I understand that it's hard to do much better without a more heavy refactor. Things I don't like:
  - the use of `cb` only for errors, the fact that it has a `cb` argument at all in fact
  - the `fs.readFileSync()` to read the cert file

A begrudging lgtm since I understand the challenges of doing a better job, I would prefer the `cb` thing be handled but if you feel strongly enough that the way you've implemented it is good enough then go ahead.

Yet another refactor for the future I suppose.

bnoordhuis:
I added a commit that lets the exception bubble up.  I'm a bit on the fence on whether it's an improvement; let me know what you think.

rvagg:
I'm happy with the clarity of the changes, but generally unhappy with the structure of the call chain but that's a matter for another day! lgtm

PR #844, opened 2015-12-15
Use npm's cafile config re #848

mschnee:
Do a google search for "UNABLE_TO_GET_ISSUER_CERT_LOCALLY" or "can't install behind corporate firewall" or "can't install behind proxy".

tl;dr I need to install a module, which requires something to be rebuilt, and it fails because node-gyp doesn't honor my npm configurations for cafiles (for corporate self-signed certificate).

Specifically, this addresses needing to download node's headers when an install/update requires a rebuild.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -425,7 +425,6 @@ function install (gyp, argv, callback) {
 
 function download (gyp, env, url) {
   log.http('GET', url)
-
   var requestOpts = {
       uri: url
     , headers: {
@@ -433,7 +432,8 @@ function download (gyp, env, url) {
       }
   }
 
-  var cafile = gyp.opts.cafile
+  var cafile = gyp.opts.cafile || process.env.npm_config_cafile;
+
   if (cafile) {
     requestOpts.ca = readCAFile(cafile)
   }


bnoordhuis:
Did you check if HEAD of master works?  `--cafile` was recently added - it's the most recent commit in fact - and is inherited implicitly through the environment.  See commit 8c4b0ff for details.

mschnee:
I intend to double-check tomorrow.  The use-case was `npm install ${somePackageRequiringARebuild}`, with ca configured via `npm config set cafile="/path/to/cert.pem"`, and invasive debugging of gyp.opts (e.g. console.log) showed that `if (caFile)...` was a false condition.

If this is the case, and the intention is that it should be inherited implicitly or explicitly into gyp.opts, then my 'fix' is in the wrong place.  I will look into how gyp.opts is populated.

bnoordhuis:
[Here](https://github.com/nodejs/node-gyp/blob/8c4b0ff/lib/node-gyp.js#L158-L172) is the code that is responsible for populating the `gyp.opts` object with `npm_config_*` environment variables.

longtimeago:
@mschnee @bnoordhuis guys, any progress here? Is there any chance to bypass the issue without the PR?

bnoordhuis:
Closing, `--cafile` has been out for some time now and seems to be working okay.  Please holler if you have reason to believe otherwise.

PR #855, opened 2016-01-07
Add `target_modules_version` variable into generated config.gypi file

pmed:
It may be handy to have multiple native addon binaries for several Node.js versions (i.e. 0.12, 4.X, 5.X). Node.js has a value for the native modules version as `process.versions.modules`.

A particular C++ addon can be loaded in JavaScript depending on Node.js version like this:

```js
  // loads 'addon_win32_x64_m47.node' for Node.js 5.0 on Windows x64
  // loads 'addon_linux_x64_m46.node' for Node.js 4.X on Linux x64
  var native = require('addon' + process.platform + '_'
              + process.arch + '_m' + process.versions.modules + '.node');
```

The target binary file name should be set with `product_name` directive in binding.gyp file in a such way:

```gyp
  {
    'targets': [
      {
        'product_name': 'addon_<(target_platform)_<(target_arch)_m<(target_modules_version)'
        ... # rest of GYP directives
      }
    ]
  }
```

where `target_platform` is a configuration variable with value from Node.js `process.platform` property.

But the `process.versions.modules` value can't be used for `target_modules_version` since node-gyp can build targets for different Node.js versions specified with a `--target` command-line option.

To resolve this issue, a function getModulesVersion() was added on `configure` step to get a value of `NODE_MODULE_VERSION` defined in `node_version.h` file in a supplied directory with Node.js development files.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -81,6 +81,29 @@ function configure (gyp, argv, callback) {
     }
   }
 
+  function getModulesVersion(nodeDir) {
+
+    var sources = [
+      path.join(nodeDir, 'include/node/node_version.h'),
+      path.join(nodeDir, 'src/node_version.h'),
+    ]
+
+    for (var i = 0; i < sources.length; ++i) {
+      try {
+        var source = fs.readFileSync(sources[i], { encoding: 'utf8'})
+        var version = source.match(/^#define\s+NODE_MODULE_VERSION\s+(\d+)/m)
+        if (!version) {
+          err = new Error("can't find #define NODE_MODULE_VERSION in " + sources[i]);
+          callback(err);
+        }
+        return parseInt(version[1], 10);
+      }
+      catch (e) {}
+    }
+    err = new Error("can't read node_version.h in " + nodeDir)
+    callback(err);
+  }
+
   function createBuildDir () {
     log.verbose('build dir', 'attempting to create "build" dir: %s', buildDir)
     mkdirp(buildDir, function (err, isNew) {
@@ -127,6 +150,12 @@ function configure (gyp, argv, callback) {
 
     // set the target_arch variable
     variables.target_arch = gyp.opts.arch || process.arch || 'ia32'
+
+    // set the target_platform variable
+    variables.target_platform = process.platform
+
+    // set the target modules version
+    variables.target_modules_version = getModulesVersion(nodeDir)
 
     // set the node development directory
     variables.nodedir = nodeDir
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -128,15 +128,18 @@ function configure (gyp, argv, callback) {
 
     // set the target_arch variable
     variables.target_arch = gyp.opts.arch || process.arch || 'ia32'
-    log.verbose('build/' + configFilename, 'target arch: ' + variables.target_arch)
+    log.verbose('build/' + configFilename,
+      'target arch: ' + variables.target_arch)
 
     // set the target_platform variable
     variables.target_platform = process.platform
-    log.verbose('build/' + configFilename, 'target platform: ' + variables.target_platform)
+    log.verbose('build/' + configFilename,
+      'target platform: ' + variables.target_platform)
 
     // set the target modules version
     variables.target_modules_version = nodeModulesVersion(nodeDir)
-    log.verbose('build/' + configFilename, 'modules version: ' + variables.target_modules_version)
+    log.verbose('build/' + configFilename,
+      'modules version: ' + variables.target_modules_version)
 
     // set the node development directory
     variables.nodedir = nodeDir
@@ -332,7 +335,7 @@ function nodeModulesVersion(nodeDir) {
   // parse version from the base name of nodeDir
   // probably removing optional `releaseName-` prefix
   var nodeVersion = semver.valid(versionStr)
-                 || semver.valid(versionStr.substr(versionStr.indexOf('-') + 1))
+        || semver.valid(versionStr.substr(versionStr.indexOf('-') + 1))
 
   if (!nodeVersion) {
     throw new Error('Can\'t extract target version from ' + nodeDir)
@@ -349,7 +352,7 @@ function nodeModulesVersion(nodeDir) {
   }
   nodeVersion = path.join(nodeDir, nodeVersion)
 
-  var modulesVersion = fs.readFileSync(nodeVersion, { encoding: 'utf8'})
+  var modulesVersion = fs.readFileSync(nodeVersion, { encoding: 'utf8' })
                           .match(/#define\s+NODE_MODULE_VERSION\s+\(?(\S+)\)?/)
   if (!modulesVersion) {
     throw new Error('#define NODE_MODULE_VERSION not found in ' + nodeVersion)
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -83,25 +83,35 @@ function configure (gyp, argv, callback) {
 
   function getModulesVersion(nodeDir) {
 
-    var sources = [
-      path.join(nodeDir, 'include/node/node_version.h'),
-      path.join(nodeDir, 'src/node_version.h'),
-    ]
+    var name, source, version
 
-    for (var i = 0; i < sources.length; ++i) {
-      try {
-        var source = fs.readFileSync(sources[i], { encoding: 'utf8'})
-        var version = source.match(/^#define\s+NODE_MODULE_VERSION\s+(\d+)/m)
-        if (!version) {
-          err = new Error("can't find #define NODE_MODULE_VERSION in " + sources[i]);
-          callback(err);
-        }
-        return parseInt(version[1], 10);
-      }
-      catch (e) {}
+    // NODE_MODULE_VERSION is defined in different files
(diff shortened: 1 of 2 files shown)


lib/configure.js
```diff
@@ -80,6 +80,29 @@ function configure (gyp, argv, callback) {
...
+  function getModulesVersion(nodeDir) {
+
+    var sources = [
+      path.join(nodeDir, 'include/node/node_version.h'),
+      path.join(nodeDir, 'src/node_version.h'),
+    ]
+
+    for (var i = 0; i < sources.length; ++i) {
+      try {
+        var source = fs.readFileSync(sources[i], { encoding: 'utf8'})
```
bnoordhuis: The try/catch should only be around this statement.  Style nit: space before `}`.

lib/configure.js
```diff
@@ -80,6 +80,29 @@ function configure (gyp, argv, callback) {
...
+        var version = source.match(/^#define\s+NODE_MODULE_VERSION\s+(\d+)/m)
+        if (!version) {
+          err = new Error("can't find #define NODE_MODULE_VERSION in " + sources[i]);
```
bnoordhuis: Style nit: can you use single quotes here and below?

lib/configure.js
```diff
@@ -80,6 +80,29 @@ function configure (gyp, argv, callback) {
...
+          callback(err);
+        }
+        return parseInt(version[1], 10);
+      }
+      catch (e) {}
+    }
+    err = new Error("can't read node_version.h in " + nodeDir)
+    callback(err);
```
bnoordhuis: Can this function end up calling the callback twice or thrice?  Looks like it.

lib/configure.js
```diff
@@ -80,6 +80,29 @@ function configure (gyp, argv, callback) {
...
+
+    for (var i = 0; i < sources.length; ++i) {
+      try {
+        var source = fs.readFileSync(sources[i], { encoding: 'utf8'})
+        var version = source.match(/^#define\s+NODE_MODULE_VERSION\s+(\d+)/m)
+        if (!version) {
+          err = new Error("can't find #define NODE_MODULE_VERSION in " + sources[i]);
+          callback(err);
+        }
+        return parseInt(version[1], 10);
```
bnoordhuis: This will fail when `version === null`.

bnoordhuis:
Can you add a couple of regression tests?  You'll probably need to adjust the logic a little so you can test it without trying to read from disk (or make it use fixtures somehow.)

pmed:
Reworked getModulesVersion() to read a particular file, depending on the target Node.js version.

No tests were added yet, because I don't know how to make appropriate fixtures for them.

saper:
Ideally this should be provided by node itself in it's config.gypi file.

pmed:
Renamed function to `nodeModulesVersion()` added tests for it.

pmed:
@bnoordhuis could you review recent changes please?

lib/configure.js
```diff
@@ -127,6 +128,15 @@ function configure (gyp, argv, callback) {
...
     variables.target_arch = gyp.opts.arch || process.arch || 'ia32'
+    log.verbose('build/' + configFilename, 'target arch: ' + variables.target_arch)
+
+    // set the target_platform variable
+    variables.target_platform = process.platform
+    log.verbose('build/' + configFilename, 'target platform: ' + variables.target_platform)
+
+    // set the target modules version
+    variables.target_modules_version = nodeModulesVersion(nodeDir)
+    log.verbose('build/' + configFilename, 'modules version: ' + variables.target_modules_version)
```
bnoordhuis: Can you wrap lines at 80 columns?  I realize there are long lines elsewhere but I'd like to avoid introducing new ones.

lib/configure.js
```diff
@@ -312,6 +322,42 @@ function configure (gyp, argv, callback) {
...
+ * from Node source code in specified directory
+ * for specified Node version
+ */
+function nodeModulesVersion(nodeDir) {
+  var versionStr = path.basename(nodeDir)
+
+  // parse version from the base name of nodeDir
+  // probably removing optional `releaseName-` prefix
+  var nodeVersion = semver.valid(versionStr)
+                 || semver.valid(versionStr.substr(versionStr.indexOf('-') + 1))
```
bnoordhuis: `||` should go on the previous line.

lib/configure.js
```diff
@@ -312,6 +322,42 @@ function configure (gyp, argv, callback) {
...
+  if (semver.gte(nodeVersion, '3.0.0')) {
+    nodeVersion = 'include/node/node_version.h'
+  } else if (semver.gt(nodeVersion, '0.11.4')) {
+    nodeVersion = 'src/node_version.h'
+  } else {
+    nodeVersion = 'src/node.h'
+  }
+  nodeVersion = path.join(nodeDir, nodeVersion)
+
+  var modulesVersion = fs.readFileSync(nodeVersion, { encoding: 'utf8'})
```
bnoordhuis: Style: space before `}`.

You should wrap the `fs.readFileSync` call in a try/catch in case the header file is missing.

test/test-node-modules-version.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+var test = require('tape')
+var path = require('path')
+var semver = require('semver')
+var processRelease = require('../lib/process-release')
+var configure = require('../lib/configure')
+
+var gyp = require('../lib/node-gyp')() // for gyp.devDir
+gyp.parseArgv([]) // to initialize gyp.opts
```
bnoordhuis: Style nit: two spaces before `//`.

test/test-node-modules-version.js
```diff
@@ -0,0 +1,34 @@
...
+
+test('modules version for the current release', function (t) {
+  t.plan(1)
+
+  var release = processRelease([], gyp, process.version)
+  var nodeDir = path.join(gyp.devDir, release.versionDir)
+  var modulesVersion = configure.test.nodeModulesVersion(nodeDir)
+  t.equal(modulesVersion, process.versions.modules,
+    nodeDir + ' modulesVersion=' + modulesVersion)
```
bnoordhuis: Style nit: can you line up the arguments here and below?

bnoordhuis:
This needs more tests, there are quite a few cases that aren't covered yet.  For example, the use case of `node-gyp --nodedir /path/to/master configure` (i.e., no version number in the path) should be tested because I'm quite sure that's broken now.

If you make the `fs.readFileSync` call pluggable, e.g. by making it an optional argument to `nodeModulesVersion()`, you can unit test it more easily.

pmed:
@bnoordhuis thank you very much! I will think on more tests.

lib/configure.js
```diff
@@ -127,6 +128,18 @@ function configure (gyp, argv, callback) {
 
     // set the target_arch variable
     variables.target_arch = gyp.opts.arch || process.arch || 'ia32'
+    log.verbose('build/' + configFilename,
+      'target arch: ' + variables.target_arch)
+
+    // set the target_platform variable
+    variables.target_platform = process.platform
```
saper: Maybe we should allow manual override here as well? What about cross-platform builds?

bnoordhuis:
> needs more tests, there are quite a few cases that aren't covered yet

Something that just came to mind: building for electron / nw.js / node-webkit.

Also, it's perhaps worth pointing out that node.js v6.3.0 introduced the `node_module_version` gyp variable, at least partially obviating this PR.

saper:
@bnoordhuis do you think that `node_module_version` will be backported to the older release trains?

bnoordhuis:
Maybe to v4 but not to v0.10 and v0.12.  Those are near their end of life anyway, though.

TheAlphaNerd:
@bnoordhuis should we open an issue on @nodejs/lts?

bnoordhuis:
@TheAlphaNerd Sure, go ahead.

saper:
Having this in some final releases of 0.10 and 0.12 would be ultra cool. 

Actually this patch could become a separate npm module which kicks in as a fallback if module version is not available via the gypi files.

bnoordhuis:
The `node_modules_version` variable was introduced as a set of bigger changes that make it possible to build node as a shared library.  That is not going to get back-ported but if you want to take a stab at back-porting just the variable, you're welcome.

Caveat emptor though, v0.10 and v0.12 are in maintenance mode.  Normally only security fixes and critical fixes are accepted but since `node_modules_version` is useful for building add-ons across versions, I think we can make an exception.

saper:
Thank you @bnoordhuis, this is good news! I will have a look at backporting just the variable.

saper:
Great, given that `tools/getnodeversion.py` has been there since 2001, the backport will be trivial.

saper:
A quick and seemingly working backport to v0.10: https://github.com/nodejs/node/pull/7808

bnoordhuis:
@pmed Do you think this PR is still necessary?

pmed:
As I understand, the variable `node_module_version` has been added into `config.gypi` in Node versions 4.x and 6.x

I think there is no more need to do the same thing in node-gyp, so let's close this PR.

bnoordhuis:
Okay thanks, I'll close.

PR #861, opened 2016-01-13
Allow for npmlog@2.x

iarna:
The breaking change introduced in npmlog 2.x was to anyone who was using the
log as an event emitter and was listening for error events. Since node-gyp
doesn't do these things, it isn't impacted by this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "minimatch": "1",
     "mkdirp": "^0.5.0",
     "nopt": "2 || 3",
-    "npmlog": "0 || 1",
+    "npmlog": "0 || 1 || 2",
     "osenv": "0",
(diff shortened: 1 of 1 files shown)


rvagg:
lgtm

rvagg:
@ralphtheninja I believe it's so that updates are not necessarily forced and that `dedupe` can make larger compromises .. or something. Basically all of these versions _work_ for the way that node-gyp uses it so there's not a strong reason to _need_ a later one.

rvagg:
merged @ 6e28ad1, we already have a semver-minor pending for release so I'd like to try and land #878 and #877 over the next couple of days before releasing, additional eyes on those would help.

PR #869, opened 2016-01-19
update PRESUBMIT.py to support python 2.6

zeldal:
To support node-gyp python 2.6 and so we can build on default redhat EL6 configurations.


bnoordhuis:
Thanks, but PRESUBMIT.py is not used by node-gyp itself.  It's used only when you submit patches to the upstream gyp project.

zeldal:
Well, we are just building rpm for EL6. since Nodesource distribution doesnt support. So i will just ignore PRESUBMIT.py while building rpm.
Thank you.

PR #873, opened 2016-01-27
Enable Node.js to build with Microsoft's ChakraCore engine

kunalspathak:
Enable node.js to build with Microsoft's chakracore engine on Windows x86, x64 and ARM architecture.

The changes were originally made as part of nodejs/node#4765, but based on @shigeki [comments](https://github.com/nodejs/node/pull/4765#commitcomment-15548451) and my [response](https://github.com/nodejs/node/pull/4765#issuecomment-173313320), I removed those changes from the PR and sending it separately to this repo.

Commits summary : 
- Add support to build node.js with chakracore
- Add support to build node.js with chakracore for Windows on ARM machines.

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -8,7 +8,7 @@
       '<(node_root_dir)/include/node',
       '<(node_root_dir)/src',
       '<(node_root_dir)/deps/uv/include',
-      '<(node_root_dir)/deps/v8/include'
+      '<(node_root_dir)/<(node_engine_include_dir)'
     ],
     'defines!': [
       'BUILDING_UV_SHARED=1',  # Inherited from common.gypi.
@@ -79,6 +79,12 @@
         ],
       }],
       [ 'OS=="win"', {
+        'conditions': [
+          ['node_engine=="chakracore"', {
+            'library_dirs': [ '<(node_root_dir)/$(ConfigurationName)' ],
+            'libraries': [ '<@(node_engine_libs)' ],
+          }],
+        ],
         'libraries': [
           '-lkernel32.lib',
           '-luser32.lib',
--- a/addon.gypi
+++ b/addon.gypi
@@ -1,15 +1,13 @@
 {
+  'variables' : {
+    'node_engine_include_dir%': 'deps/v8/include',
+  },
   'target_defaults': {
     'type': 'loadable_module',
     'win_delay_load_hook': 'true',
     'product_prefix': '',
 
     'conditions': [
-      [ 'node_engine=="v8"', {
-        'variables': {
-          'node_engine_include_dir%': 'deps/v8/include'
-        },
-      }],
       [ 'node_engine=="chakracore"', {
         'variables': {
           'node_engine_include_dir%': 'deps/chakrashim/include'
--- a/addon.gypi
+++ b/addon.gypi
@@ -4,6 +4,19 @@
     'win_delay_load_hook': 'true',
     'product_prefix': '',
 
+    'conditions': [
+      [ 'node_engine=="v8"', {
+        'variables': {
+          'node_engine_include_dir%': 'deps/v8/include'
+        },
+      }],
+      [ 'node_engine=="chakracore"', {
+        'variables': {
+          'node_engine_include_dir%': 'deps/chakrashim/include'
+        },
+      }]
+    ],
+
     'include_dirs': [
       '<(node_root_dir)/include/node',
       '<(node_root_dir)/src',
--- a/lib/build.js
+++ b/lib/build.js
@@ -226,7 +226,9 @@ function build (gyp, argv, callback) {
 
     // Specify the build type, Release by default
     if (win) {
-      var p = arch === 'x64' ? 'x64' : 'Win32'
+      var archLower = arch.toLowerCase()
+      var p = archLower === 'x64' ? 'x64' :
+              (archLower === 'arm' ? 'ARM' : 'Win32')
       argv.push('/p:Configuration=' + buildType + ';Platform=' + p)
       if (jobs) {
         var j = parseInt(jobs, 10)
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -272,7 +272,8 @@ function configure (gyp, argv, callback) {
       argv.push('-Dnode_gyp_dir=' + nodeGypDir)
       argv.push('-Dnode_lib_file=' + release.name + '.lib')
       argv.push('-Dmodule_root_dir=' + process.cwd())
-      argv.push('-Dnode_engine=' + (gyp.opts.node_engine || process.jsEngine || 'v8'))
+      argv.push('-Dnode_engine=' +
+        (gyp.opts.node_engine || process.jsEngine || 'v8'))
       argv.push('--depth=.')
       argv.push('--no-parallel')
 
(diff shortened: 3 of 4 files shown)


bnoordhuis:
Thanks for the pull request.  This can't land before the PR for core lands but let me point out now that the changes to files in the gyp/ directory should be submitted to the upstream gyp project first.

lib/build.js
```diff
@@ -227,6 +227,9 @@ function build (gyp, argv, callback) {
     // Specify the build type, Release by default
     if (win) {
       var p = arch === 'x64' ? 'x64' : 'Win32'
```
addaleax: @kunalspathak The previous `var p = …` here should be removed, I guess?

kunalspathak: Ah, that's right.

rvagg:
Sorry about the run-around @kunalspathak but to echo what @bnoordhuis said, we've maintained a fairly strict policy here of only using gyp as it comes from upstream and not landing additional changes ontop of it. Let us know if you need help on that front.

kunalspathak:
Thanks @rvagg , not a problem. Could you point me the upstream branch location?

shigeki:
The official of GYP project is https://gyp.gsrc.io/.

Submitting a patch is described in
https://groups.google.com/d/msg/gyp-developer/yPVVz46kds0/DJvWJzBBeO0J 
and 
http://dev.chromium.org/developers/contributing-code

@bnoordhuis knows its workflow well.

kunalspathak:
Thanks @shigeki , I will go through the work-flow and ping @bnoordhuis if i have more questions.
So we could close this PR then?

bnoordhuis:
I'm okay with keeping this open if you can rework it to drop the changes to gyp and link to the CL you're going to file.

bnoordhuis:
@kunalspathak Can you give a quick update on the status of this pull request?  I see some changes have landed in upstream gyp.  Does that mean this PR is ready to go through?

kunalspathak:
@munyirik , you worked on landing this work upstream, right? Can we get this PR merged in that case?

munyirik:
Yeah the gyp change (in msvs.py) got [committed](https://chromium.googlesource.com/external/gyp/+/02b145a1a4f4e1c62e8bae06045caf852d9ef17f). I don't see the change in nodejs:master. Isn't that where the change should go first before this gets merged?

lib/configure.js
```diff
@@ -272,6 +272,7 @@ function configure (gyp, argv, callback) {
       argv.push('-Dnode_gyp_dir=' + nodeGypDir)
       argv.push('-Dnode_lib_file=' + release.name + '.lib')
       argv.push('-Dmodule_root_dir=' + process.cwd())
+      argv.push('-Dnode_engine=' + (gyp.opts.node_engine || process.jsEngine || 'v8'))
```
bnoordhuis: `process.jsEngine` doesn't look quite correct.  The current node binary might not be the one you're targeting.  (node -> node-webkit is a common example.)

kunalspathak: @bnoordhuis , sorry I didn't exactly follow your comment. Could you please elaborate? 
`process.jsEngine` is something that was added in `node-chakracore` to distinguish between engines.

kunalspathak:
> The change to MSVSSettings.py seems to be missing. Isn't it needed?

Sorry, what changes did you mean? I might be missing something here.

bnoordhuis:
See bnoordhuis/gyp@02b145a1a4f4e1c62e8bae06045caf852d9ef17f - that's the gyp changes, right?  That patch adds a line to MSVSSettings.py but the patch in this pull request does not.

kunalspathak:
My impression was that patch for `MSVSSettings.py` would come from upstream `gyp` that's why I didn't include it here. But I guess now it is blocked, it is fine to get those changes as well with this PR. In that case, I will also have to pull in changes from `msvs.py`.

lib/configure.js
kunalspathak: Sorry for the late reply. The design decision we had taken is other way round. Once user specified `--node-engine`, we will include appropriate headers `deps\v8\include` or `deps\chakrasim\include`. 
E.g. On windows, it is `vcbuild.bat chakracore`
On Linux/Mac , it is `./configure --engine=chakracore`.

munyirik:
The change in MSVSSettings.py (the CompileAsWinRT change) wasn't required in this PR. It's something I needed for my UWP fork. 
One thing to note is that I changed `msvs_windows_target_platform_version` to `msvs_windows_sdk_version` in msvs.py.

addon.gypi
```diff
@@ -8,7 +8,7 @@
       '<(node_root_dir)/include/node',
       '<(node_root_dir)/src',
       '<(node_root_dir)/deps/uv/include',
-      '<(node_root_dir)/deps/v8/include'
+      '<(node_root_dir)/<(node_engine_include_dir)'
```
bnoordhuis: node_engine_include_dir is not defined anywhere, only node_engine.

lib/build.js
bnoordhuis: This is still unfixed.

lib/configure.js
bnoordhuis: Okay, I think I follow.  Still, using `process.jsEngine` here means we potentially build the add-on for the wrong binary, right?  Not the end of the world but I need to know if my understanding is correct.

lib/build.js
kunalspathak: My bad, this got missed out while merging. I will remove it.

addon.gypi
kunalspathak: Yeah, this is defined in `node-chakracore` [here](https://github.com/nodejs/node-chakracore/blob/xplat/common.gypi#L76-L78) and [here](https://github.com/nodejs/node-chakracore/blob/xplat/common.gypi#L92). I didn't bring in these changes when I created PR. I will move it from `node-chakracore` to `node-gyp` because that is the right place of those variables.

lib/configure.js
kunalspathak: If you build `node.exe` using `--node-engine=chakracore`, and using that binary if you are building an add-on, `process.jsEngine` will be automatically populated with `chakracore` and we will be building an add-on compatible with `node-chakracore`.  Likewise for `node-v8`.

Only scenario where things can go wrong is someone build `node-chakracore` and use it to build add-on passing `--node-engine=v8` or vice versa. So in that case, your understanding is correct.

kunalspathak:
ping. @bnoordhuis , do you have any other feedback?

lib/configure.js
bnoordhuis: Okay, thanks.  Minor nit: can you wrap this line at 80 columns?

kunalspathak:
> Can you make sure tests pass

I checked and it was because test were ran with `node-v8` that didn't have `node_engine_include_dir` definition. Moved the definition as default for v8.



> can you wrap this line at 80 columns?

Sure. I noticed that there were some lines that don't follow 80 columns rule, so didn't paid attention. Updated to be within 80 columns.

bnoordhuis:
Thanks.  New CI: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/37/

bnoordhuis:
All green.  Thanks Kunal, landed in 93d7fa8...ec5fc36.

PR #877, opened 2016-02-12
download headers tarball for ~0.12.10 || ~0.10.42

rvagg:
/R= @nodejs/build @ljharb I guess

### DIFF
--- a/lib/process-release.js
+++ b/lib/process-release.js
@@ -2,6 +2,8 @@ var semver = require('semver')
   , url = require('url')
   , path = require('path')
 
+    // versions where -headers.tar.gz started shipping
+  , headersTarballRange = '>= 3.0.0 || ~0.12.10 || ~0.10.42'
   , bitsre = /\/win-(x86|x64)\//
   , bitsreV3 = /\/win-(x86|ia32|x64)\// // io.js v3.x.x shipped with "ia32" but should
                                         // have been "x86"
@@ -21,6 +23,7 @@ function processRelease (argv, gyp, defaultVersion, defaultRelease) {
     , libUrl32
     , libUrl64
     , tarballUrl
+    , canGetHeaders
 
   if (!versionSemver) {
     // not a valid semver string, nothing we can do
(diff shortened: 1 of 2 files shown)


rvagg:
I'd consider this a semver-minor and warrants introduction into npm@2 if we can manage it. There's some queued up changes that need to go out with the next 0.10 and 0.12 so that could be done fairly soon. So /cc @zkat.

lib/process-release.js
```diff
@@ -93,7 +95,7 @@ function processRelease (argv, gyp, defaultVersion, defaultRelease) {
   // making the bold assumption that anything with a version number >3.0.0 will
   // have a *-headers.tar.gz file in its dist location, even some frankenstein
   // custom version
-  tarballUrl = url.resolve(baseUrl, name + '-v' + version + (versionSemver.major >= 3 ? '-headers' : '') + '.tar.gz')
+  tarballUrl = url.resolve(baseUrl, name + '-v' + version + (semver.satisfies(versionSemver, headersTarballRange) ? '-headers' : '') + '.tar.gz')
```
bnoordhuis: Overlong line.

bnoordhuis:
LGTM sans style nit.  The commit log could go into more detail.

rvagg:
thanks @bnoordhuis, both addressed

bnoordhuis:
LGTM but can you `s/^download/Download` before you land it?

jbergstroem:
Do we care about xz at this stage? Benefitting from headers only is good enough? Anyway, I can likely represent the Internet Bandwidth Committee stating LGTM.

rvagg:
@jbergstroem I'd love to use xz for these but we don't have xz support from zlib so uncompressing them without invoking system tools is tricky. I've considered pulling in a JS implementation but haven't done any research on whether they exist or whether they are any good. Given that we're talking about ~500kb the effort in researching and coding it up is unlikely to yield enough benefit to warrant me bothering!

rvagg:
^ that's not to say I wouldn't support someone else doing this work if they feel it's an interesting enough challenge and worth their time, it's just not worth my time

rvagg:
landed @ 070fe69

Issue #787, opened 2015-10-28
node-gyp rebuild downloads headers from incorrect location


mgol:
I thought that `node-gyp` `3.0.3` should have this all fixed by taking the data from `process.release` but it still downloads from the default one for me:

```
$ node --version
v4.2.1
$ npm --version
3.3.10
$ node -p process.release
{ name: 'node',
  lts: 'Argon',
  sourceUrl: 'https://nodejs.org/download/release/v4.2.1/node-v4.2.1.tar.gz',
  headersUrl: 'https://nodejs.org/download/release/v4.2.1/node-v4.2.1-headers.tar.gz' }

$ npm i node-sass@3.4.1
```

Switch to Node 5 & then (notice the incorrect URL https://nodejs.org/dist/v5.0.0-rc.1/node-v5.0.0-rc.1-headers.tar.gz):

```
$ node --version
v5.0.0-rc.1
$ npm --version
3.3.10
$ node -p process.release
{ name: 'node',
  sourceUrl: 'https://nodejs.org/download/rc/v5.0.0-rc.1/node-v5.0.0-rc.1.tar.gz',
  headersUrl: 'https://nodejs.org/download/rc/v5.0.0-rc.1/node-v5.0.0-rc.1-headers.tar.gz' }
$ npm rebuild node-sass

> node-sass@3.4.1 install /Users/mgol/_/node_modules/node-sass
> node scripts/install.js

Cannot download "https://github.com/sass/node-sass/releases/download/v3.4.1/darwin-x64-47_binding.node": 

HTTP error 404 Not Found

Hint: If github.com is not accessible in your location
      try setting a proxy via HTTP_PROXY, e.g. 

      export HTTP_PROXY=http://example.com:1234

or configure npm proxy via

      npm config set proxy http://example.com:8080

> node-sass@3.4.1 postinstall /Users/mgol/_/node_modules/node-sass
> node scripts/build.js

Building: /Users/mgol/.nvm/versions/node/v5.0.0-rc.1/bin/node /Users/mgol/_/node_modules/node-gyp/bin/node-gyp.js rebuild --verbose --libsass_ext= --libsass_cflags= --libsass_ldflags= --libsass_library=
gyp info it worked if it ends with ok
gyp verb cli [ '/Users/mgol/.nvm/versions/node/v5.0.0-rc.1/bin/node',
gyp verb cli   '/Users/mgol/_/node_modules/node-gyp/bin/node-gyp.js',
[...]

PR #878, opened 2016-02-12
introduce NODEJS_ORG_MIRROR and IOJS_ORG_MIRROR

rvagg:
/R= @nodejs/build @ljharb

to be followed soon by a semver-major that removes the `NVM_` variants as these cause grief, see #787

### DIFF
--- a/test/test-process-release.js
+++ b/test/test-process-release.js
@@ -519,3 +519,119 @@ test('test process release - process.release ~ node@4.0.0-rc.4 - bogus string pa
   })
 })
 
+test('test process release - NODEJS_ORG_MIRROR', function (t) {
+  t.plan(2)
+
+  process.env.NODEJS_ORG_MIRROR = 'http://foo.bar'
+
+  var release = processRelease([], { opts: {} }, 'v4.1.23', {
+    name: 'node',
+    headersUrl: 'https://nodejs.org/dist/v4.1.23/node-v4.1.23-headers.tar.gz'
+  })
+
+  t.equal(release.semver.version, '4.1.23')
+  delete release.semver
+
+  t.deepEqual(release, {
+    version: '4.1.23',
+    name: 'node',
+    baseUrl: 'http://foo.bar/v4.1.23/',
+    tarballUrl: 'http://foo.bar/v4.1.23/node-v4.1.23-headers.tar.gz',
+    shasumsUrl: 'http://foo.bar/v4.1.23/SHASUMS256.txt',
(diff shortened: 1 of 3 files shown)


ljharb:
:+1:

mgol:
I can't judge the build & tests but source changes look good. :+1:

ralphtheninja:
LGTM

rvagg:
Should we do a `warn` message (npmlog style) whenever `NVM_*` is encountered? I can't recall the details of how this is used internally in nvm but would it hit many users not trying to use those variables to control node-gyp? @ljharb?

ralphtheninja:
> Should we do a warn message (npmlog style) whenever NVM_* is encountered?

Like a deprecated warning? +1

ljharb:
@rvagg it's primarily that it's an env var that while I don't intend to change, I might - a warning seems reasonable so people know to switch.

rvagg:
Added a warning, run the tests to see it output:

```
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
```

@bnoordhuis can I ask for a review here as well please?

@zkat @iarna are there any guidelines on outputting warnings with `npmlog`? Is this wording going to be OK when it starts showing up for users who have that env var set?

test/test-process-release.js
```diff
@@ -519,3 +519,119 @@ test('test process release - process.release ~ node@4.0.0-rc.4 - bogus string pa
...
+})
+
+test('test process release - IOJS_ORG_MIRROR', function (t) {
+  t.plan(2)
+
+  process.env.IOJS_ORG_MIRROR = 'http://foo.bar'
+
+  var release = processRelease([], { opts: {} }, 'v3.2.24', {
+    name: 'io.js',
+    headersUrl: 'http://foo.bar/v3.2.24/iojs-v3.2.24-headers.tar.gz'
```
bnoordhuis: You use 'https://nodejs.org/` in the first two tests but `http://foo.bar` here and in the next test.  Is that intentional?

bnoordhuis:
JS changes LGTM with a question, no opinion on the docker shell script.  There are some long lines but that's not really a thing with node-gyp.

test/test-process-release.js
rvagg: No, search/replace error, the test should find that `process.release` is mostly ignored and it's not testing that as it is now. Thanks for noticing, will fix!

PR #879, opened 2016-02-14
Problem: command options aren't documented

hintjens:
Especially the --jobs and logging options are valuable and deserve to
be documented.

Solution: add 'Command options' section to README.md.

Note: I also added a -silent option to switch off all logging, for
consistency with the other logging options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,17 @@ Commands
 | `list`        | Lists the currently installed node development file versions
 | `remove`      | Removes the node development header files for the given version
 
+Command options
+---------------
+
+`node-gyp` accepts the following command options:
+
+| **Option**         | **Description**
+|:-------------------|:------------------------------------------------
+| `-j n`, `--jobs n` | For `build`: run parallel builds
+| `-silly`, `--loglevel=silly` | Log all progress to console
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Can you break this into two commits?  Extra karma if you add a small regression test for the `--silent` switch.

hintjens:
I can redo this, sure, and try to add a regression test. Do you mind if I push multiple commits into a single PR? Or do you prefer branches for each commit?

bnoordhuis:
Multiple commits are fine.  You don't need to file a new PR, you can force-push to your branch and it'll show up here (but please leave a comment when you do, GH doesn't send notifications for that.)

hintjens:
:-) great, will do.

hintjens:
@bnoordhuis turns out github wanted me to make a new PR, which I did: #880 

I've also added a new option, `--mklocal` to work around a bug in gyp that has been causing grief. It seems a rather major issue in gyp, yet I can't find it in the issue tracker before now. See https://bugs.chromium.org/p/gyp/issues/detail?id=508. I guess few people are mixing --generator-output with gyp dependencies.

PR #885, opened 2016-02-23
lib: do not check nvm path if not default release

TheAlphaNerd:
Currently the check for nvm mirror addresses will always happen,
even if the release in question is not default and provides it's own
url for finding headers.

This extra boolean check will ensure that non default releases will
use the embedded url for headers

### DIFF
--- a/lib/process-release.js
+++ b/lib/process-release.js
@@ -52,7 +52,8 @@ function processRelease (argv, gyp, defaultVersion, defaultRelease) {
   }
 
   // check for the nvm.sh standard mirror env variables
-  if (!overrideDistUrl) {
+  // do not check env if not a default release
+  if (!overrideDistUrl && !defaultRelease) {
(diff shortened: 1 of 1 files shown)


TheAlphaNerd:
it looks like there are some tests not passing here... I'm going to dig in and see what is going on tomorrow

rvagg:
If you're using `NVM_*` env vars then you're basically running in to https://github.com/nodejs/node-gyp/issues/787 which we can't really solve properly until node-gyp@4 where we can remove `NVM_*` entirely. Their introduction here was a mistake and needs to be reverted but we can't do it immediately because they are actually used and #878 is only just being rolled out now as an alternative.

The workaround is to _not_ use `NVM_*` vars, or avoid nvm entirely for non-standard releases.

TheAlphaNerd:
updated my nvm and all is right in the world again

PR #889, opened 2016-03-04
gyp: fix android generator

robertchiras:
When generating Android .mk files, ready to be used within Android build
system, we don't need LOCAL_SDK_VERSION specified, since the Android
build system already uses the latest one.
By specifying LOCAL_SDK_VERSION in Android makefiles, the build using
Android build system will fail, because the build system won't set the
path to stlport from prebuilts.

Signed-off-by: Robert Chiras <robert.chiras@intel.com>

#Note: This patch is very useful to help integrate node into an Android build system, mostly used by vendors when generating their products. I know the android generator is not maintained anymore, but fixing this generator is crucial for building node as target in an Android build system.

### DIFF
--- a/gyp/pylib/gyp/generator/android.py
+++ b/gyp/pylib/gyp/generator/android.py
@@ -164,7 +164,7 @@ class AndroidMkWriter(object):
     if self.toolset == 'host':
       self.WriteLn('LOCAL_IS_HOST_MODULE := true')
       self.WriteLn('LOCAL_MULTILIB := $(GYP_HOST_MULTILIB)')
-    else:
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Thanks Robert, landed in a981ef8 and released as v3.3.1.

bnoordhuis:
Also, /cc @othiym23.

othiym23:
Thanks for the heads up, @bnoordhuis. `node-gyp@3.3.1` will get landed in the next testing versions of `npm@3` and `npm@2` when they're cut next Thursday.

Issue #890, opened 2016-03-07
Provide fallback to Python launcher on Windows when python.exe cannot be found


poke:
I regularly get the following error:

    Error: Can't find Python executable "python", you can set the PYTHON env variable.

This is because Python on Windows **by default** does not place a Python executable into the PATH (and for what it’s worth, I would personally also opt out of that). Instead, Python has this great thing on Windows called the *Python launcher*. It’s an executable that is placed into the WINDIR directory that allows to run *any* installed Python version on the computer. Even more importantly, the installation of this Python launcher is the default and has been for a number of CPython releases. So for Windows, it’s a much better idea to depend on the Python launcher than on a `python` executable.

So I’d like to ask for a fallback in node-gyp that attempts to call `py.exe -2` before ultimately failing with the above error message. So the logic would be something like this:

1. Check for PYTHON environment variable; if set, use that executable
2. Otherwise attempt to call the `python` executable
3. Otherwise, if on Windows, attempt to call the `py -2` executable
4. Otherwise show the error, and ask the user to set the PYTHON environment variable

Note that the `py.exe` launcher should be called with a `-2` as its first argument. This ensures that a Python 2 executable is used, regardless of what has been set as the system default. Since node-gyp does not appear to support Python 3 at all yet, this would be the safest solution.

Btw. I also attempted to set the PYTHON environment variable to `py -2` to make it use the Python launcher, but that didn’t work: It sadly couldn’t find the `py -2` executable (i.e. it didn’t interpret the argument properly). Just using `py` did work, but called the wrong Python version on my system.

DbauchdSloth:
I am at a loss to understand why the added external dependency on a binary installation of python would be required just to build this app?  That added dependency adds a huge amount of hassle when trying to deploy (particularly on windows), and there is literally no runtime code that demands it.  Is there some reason this is absolutely needed or can I just refactor out the dependency on python using javascript equivalents?  Anyone mind or is there something I am missing here?  This seems a whole lot easier than trying to solve the problem of deploying python along with the app just to run the build.

poke:
I’m not sure what you mean. All I am suggesting here that when a Python executable is required to build things (which seems to be the way pretty often), that instead of just looking for `python` in the PATH or using the `PYTHON` environment variable, node-gyp should *also* attempt to call `py -2` on Windows systems before failing. Because that’s how Python will usually be available when it’s installed.

bnoordhuis:
[...]

PR #894, opened 2016-03-14
#890 Add support for the Python launcher on Windows

poke:
This is a pull request for the issue #890: “Provide fallback to Python launcher on Windows when python.exe cannot be found”.

The changes here attempt to have very little impact. The new check is just inserted between the `checkPython` call (looking for the `'python'` executable) and the `guessPython` call on Windows. Since the Python launcher will give an explicit Python 2 executable, it is more explicit than if we found a match using `guessPython`, so I gave the launcher solution more priority.

As explained in the commit message, we cannot make a call to `py.exe -2` since that wouldn’t work with `execFile`, so I had to make an actual call to the Python launcher to get the full path of the Python executable. But if the command succeeds, then that reliably gives us a valid Python 2 executable.

If there is no Python 2 version registered with the launcher, the launcher will fail when attempting to call it with `-2`. When there is no Python launcher, `execFile` will fail. In both cases, the old next step `guessPython` will run.

> When looking for a Python executable on Windows, before falling back to guessing the default location or failing completely, attempt to use the Python launcher to figure out the location of the Python executable.
> 
> The Python launcher is being distributed by default with Python distributions on Windows, and is placed in the %WINDIR% folder (which is in the PATH). This allows us to locate a Python installation even if it was installed without putting the python.exe executable itself into the PATH.
> 
> Because the Python launcher supports all versions of Python, we have to explicitly request a Python 2 version. This is done by supplying "-2" as the first command line argument. Since "py.exe -2" would be an invalid executable for "execFile", we have to use the launcher to figure out where the actual "python.exe" executable is located.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -325,7 +325,7 @@ function findPython (python, callback) {
           return checkPython()
         }
         if (win) {
-          guessPython()
+          checkPythonLauncher()
         } else {
           failNoPython()
         }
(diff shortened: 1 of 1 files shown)


lib/configure.js
```diff
@@ -340,6 +340,30 @@ function findPython (python, callback) {
...
+  // explicitly request a Python 2 version. This is done by supplying "-2" as
+  // the first command line argument. Since "py.exe -2" would be an invalid
+  // executable for "execFile", we have to use the launcher to figure out
+  // where the actual "python.exe" executable is located.
+  function checkPythonLauncher () {
+    log.verbose('could not find "' + python + '". checking python launcher')
+    var env = extend({}, process.env)
+    env.TERM = 'dumb'
+
+    execFile('py.exe', ['-2', '-c', 'import sys; print sys.executable'], { env: env }, function (err, stdout) {
```
bnoordhuis: Can you wrap this line at 80 columns?

bnoordhuis:
Mostly LGTM, thanks.  Can you verify that `npm test` still passes and maybe add one or two tests to test/test-find-python.js if possible?

poke:
Updated my pull request to wrap at 80 characters; for what it’s worth, the `checkPythonVersion` a few lines below of my code is happily breaking the 80 characters limit too – do you want me to adjust that too?

As for the test, with my changes, `npm test` does pass for me. It did not before since the `find python executable` failed on my machine (for the exact reasons why I am submitting this pull request). As such I’m also not sure how I could add additional tests for finding the python executable since it highly depends on the environment the test is run on.

bnoordhuis:
LGTM and, seeing how there are no comments from other maintainers, I'll go ahead and land this.  Thanks Patrick, merged in 3bcb172.

poke:
Glad to see this land, thanks a lot! :)

PR #908, opened 2016-04-20
gyp: inherit CC/CXX for CC/CXX.host

bnoordhuis:
Gyp defaults to gcc/g++ if CC.host/CXX.host is unset. This is not
suitable for environments that only uses the clang toolchain.

Since we already assume that the user will provide clang/clang++
through CC/CXX, lean against it (then drop to gcc/g++).

Refs: https://github.com/nodejs/node/pull/6173

/cc @jbergstroem


jbergstroem:
LGTM, lets hold off and hope that upstream merges first. I will merge in nodejs on Monday unless anything happens.

bnoordhuis:
Merged as a follow-up to nodejs/node#6173 which was merged a few days ago.

PR #911, opened 2016-04-22
Use python2 binary in gyp script

flying-sheep:
extends work started by #527

also falls back to `python` on not [PEP 394](https://www.python.org/dev/peps/pep-0394/) compatible platforms (OSX and very old linux distributions)

### DIFF
--- a/gyp/gyp
+++ b/gyp/gyp
@@ -4,5 +4,13 @@
 # found in the LICENSE file.
 
 set -e
+
+# Detect if it exists: http://stackoverflow.com/a/677212/247482
+if command -v python2 >/dev/null 2>&1; then
+  python=python2
+else
+  python=python
+fi
+
 base=$(dirname "$0")
-exec python "${base}/gyp_main.py" "$@"
+exec $python "${base}/gyp_main.py" "$@"


bnoordhuis:
The stuff in gyp/ comes from the upstream gyp project and should be upstreamed there first.  As a general rule, we don't take patches that haven't been accepted by upstream first.

flying-sheep:
where is it being developed?

bnoordhuis:
https://gyp.gsrc.io/ - be prepared to check out (and use!) depot_tools as well. =)

bnoordhuis:
Seeing how there has been no follow-up, I'll go ahead and close the pull request.

flying-sheep:
yeah, sorry. still should be fixed though. is there an issue tracking this here?

bnoordhuis:
Not to my knowledge, neither for gyp nor node-gyp.  You are welcome to file one.

ioquatix:
Can anyone actually use this package? I can't seem to install it.

ioquatix:
I've made a post on the gyp-developer mailing list. It looks like they are adding Python 3 support but it might be a while off. Is there any harm to merging in a fix here which solves the problem? as a temporary work around until the GYP team figure it out.

ADTC:
Is this in place? I have never installed Python 3 (only Python2) yet gyp is looking for python2 instead of python. I guess I have to alias the executable to workaround this problem.

Issue #914, opened 2016-04-28
Preprocessor defines BUILDING_V8_SHARED=1 and BUILDING_UV_SHARED=1


Flarna:
I tried the HelloWorld addon sample (using node 4.4.3 for windows) and found that node-gyp sets following defines in the generated VS solution (besides others):
```
BUILDING_V8_SHARED=1
BUILDING_UV_SHARED=1
BUILDING_NODE_EXTENSION
```
While BUILDING_NODE_EXTENSION is clear the other two seem to be strange because of following statement in v8.h:
```
// Setup for Windows DLL export/import. When building the V8 DLL the
// BUILDING_V8_SHARED needs to be defined. When building a program which uses
// the V8 DLL USING_V8_SHARED needs to be defined. When either building the V8
// static library or building a program which uses the V8 static library neither
// BUILDING_V8_SHARED nor USING_V8_SHARED should be defined.
#if defined(BUILDING_V8_SHARED) && defined(USING_V8_SHARED)
#error both BUILDING_V8_SHARED and USING_V8_SHARED are set - please check the\
  build configuration to ensure that at most one of these is set
#endif

#ifdef BUILDING_V8_SHARED
# define V8_EXPORT __declspec(dllexport)
#elif USING_V8_SHARED
# define V8_EXPORT __declspec(dllimport)
#else
# define V8_EXPORT
#endif  // BUILDING_V8_SHARED
```
In case I first include node.h this seems to be "repaired" via following statements in node.h:
```
#ifdef BUILDING_NODE_EXTENSION
# undef BUILDING_V8_SHARED
# undef BUILDING_UV_SHARED
# define USING_V8_SHARED 1
# define USING_UV_SHARED 1
#endif
```
But if I first include v8.h or other V8 files like v8-profiler.h the corrections done in node.h are missing.

I think that `USING_V8_SHARED=1` and `USING_UV_SHARED=1` should be set instead `BUILDING_V8_SHARED=1` and `BUILDING_UV_SHARED=1` for node extensions.

bnoordhuis:
https://github.com/nodejs/node-gyp/pull/915

PR #915, opened 2016-04-28
Override BUILDING_UV_SHARED, BUILDING_V8_SHARED.

bnoordhuis:
We were inheriting BUILDING_UV_SHARED and BUILDING_V8_SHARED from
the common.gypi from the node.js source tree.  Override them with
USING_UV_SHARED and USING_V8_SHARED so symbols get tagged with
`__declspec(dllimport)` instead of `__declspec(dllexport)`, which
is the desired attribute when building add-ons.

Fixes: https://github.com/nodejs/node-gyp/issues/914

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -10,8 +10,14 @@
       '<(node_root_dir)/deps/uv/include',
       '<(node_root_dir)/deps/v8/include'
     ],
+    'defines!': [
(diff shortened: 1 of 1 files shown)


rvagg:
this one is a bit beyond me, @indutny, @kkoopa?

kkoopa:
Seems correct

indutny:
LGTM

PR #916, opened 2016-04-28
Add --devdir flag.

bnoordhuis:
Overrides the location where node-gyp will look for (and download to)
the node.js SDK.

For harmonization with the --nodedir flag, `osenv.home()` is now used
instead of `process.env.HOME || process.env.USERPROFILE`.

`osenv.home()` defaults to the built-in `os.homedir()` when available
and falls back to sane defaults based on the platform.

`%USERPROFILE%` is now preferred on Windows and `%HOME%` is ignored;
similarly, UNIX systems ignore `$USERPROFILE` from now on.  Neither
behavior made much sense, IMO.

Fixes: https://github.com/nodejs/node-gyp/issues/21

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ Command Options
 | `--thin=yes`                      | Enable thin static libraries
 | `--arch=$arch`                    | Set target architecture (e.g. ia32)
 | `--tarball=$path`                 | Get headers from a local tarball
+| `--devdir=$path`                  | SDK download directory (default=~/.node-gyp)
 | `--ensure`                        | Don't reinstall headers if already present
 | `--dist-url=$url`                 | Download header tarball from custom URL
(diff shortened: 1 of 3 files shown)


bnoordhuis:
Obsoletes https://github.com/nodejs/node-gyp/pull/540.

rvagg:
lgtm

bnoordhuis:
Rebased.  CI: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/9/

as-cii:
Great work here, @bnoordhuis! Thank you so much! ✨ Is there an ETA for the next node-gyp release that will contain this change? We would like to stop using the `$HOME` hack in Atom/Electron and the `--dev-dir` flag would be super helpful!

bnoordhuis:
@as-cii I can probably do a v3.5.0 release next week.  There is one open pull request (#919) that I need to look at again.

PR #919, opened 2016-05-06
Allow build directory to be changed.

### LINKED ISSUE #263: Configurable build dir
It would be nice if the build dir can be set to something other than "build". Either via the binding.gyp or via an option of the node-gyp commands.

peterbraden:
Change with `NODE_GYP_BUILD_DIR` environment variable

Fixes #263


bnoordhuis:
It should be a switch, like `--builddir`.  The way node-gyp works, it can then also be configured by setting `npm_config_builddir` in the environment.  See #916 for an example.

guymguym:
Planning to merge? Would be great to allow it for tensorflow.

lib/configure.js
```diff
@@ -153,12 +153,12 @@ function configure (gyp, argv, callback) {
...
     }
 
-    log.silly('build/' + configFilename, config)
+    log.silly(gyp.buildDir + '/' + configFilename, config)
 
     // now write out the config.gypi file to the build/ dir
     var prefix = '# Do not edit. File was generated by node-gyp\'s "configure" step'
       , json = JSON.stringify(config, boolsToString, 2)
-    log.verbose('build/' + configFilename, 'writing out config file: %s', configPath)
+    log.verbose(gyp.buildDir + '/' + configFilename, 'writing out config file: %s', configPath)
```
bnoordhuis: Can you wrap this line at 80 columns?  I'd break at the the comma, then line up the second argument with the first one.

bnoordhuis:
@peterbraden Can you add a regression test to test/test-addon.js?

peterbraden:
added a very basic test.

test/test-addon.js
```diff
@@ -26,3 +28,26 @@ test('build simple addon', function (t) {
...
+    var lastLine = logLines[logLines.length-1]
+    t.strictEqual(err, null)
+    t.strictEqual(lastLine, 'gyp info ok', 'should end in ok')
+
+    try {
+      var binding = require('hello_world')
+      t.strictEqual(binding.hello(), 'world')
+
+      // Cleanup
+      exec('rm -rf foo', t.end)
```
bnoordhuis: This won't work on Windows.  Can you run `node-gyp --build_dir=foo clean`?

Naming bikeshed: what about calling it `--build-dir`?  That's more in line with `--dist-url`.  (OTOH there's `--msvs_version` but that mimics the gyp variable.)

test/test-addon.js
```diff
@@ -26,3 +28,26 @@ test('build simple addon', function (t) {
   proc.stdout.setEncoding('utf-8')
   proc.stderr.setEncoding('utf-8')
 })
+
+test('build with different build dir', function(t){
+  var cmd = [nodeGyp, 'rebuild', '-C', addonPath, '--loglevel=verbose', '--build_dir=foo']
```
bnoordhuis: Can you wrap this line at 80 columns?

test/test-addon.js
```diff
@@ -26,3 +28,26 @@ test('build simple addon', function (t) {
   proc.stdout.setEncoding('utf-8')
   proc.stderr.setEncoding('utf-8')
 })
+
+test('build with different build dir', function(t){
```
bnoordhuis: Style nit: space before '{'.

bnoordhuis:
@peterbraden Can you rebase and squash?  Thanks.

peterbraden:
Done

rvagg:
Your test doesn't work, for two reasons: require cache saving "hello_world" and hello.js using the bindings module to load from the existing build directory. So here's a patch to fix your test:

```diff
diff --git a/test/test-addon.js b/test/test-addon.js
index 4b8bc5e..2816cd1 100644
--- a/test/test-addon.js
+++ b/test/test-addon.js
@@ -7,6 +7,20 @@ var exec = childProcess.exec
 var path = require('path')
 var addonPath = path.resolve(__dirname, 'node_modules', 'hello_world')
 var nodeGyp = path.resolve(__dirname, '..', 'bin', 'node-gyp.js')
+var rimraf = require('rimraf')
+
+function cleanup (dir) {
+  return function teardown (t) {
+    Object.keys(require.cache)
+      .forEach(function (d) {
+        if (d.indexOf(addonPath) == 0)
+          delete require.cache[d]
+      })
+    rimraf(dir, t.end)
+  }
+}
+
+test('build simple addon - setup', cleanup(path.join(addonPath, 'build')))

 test('build simple addon', function (t) {
   t.plan(3)
@@ -29,6 +43,10 @@ test('build simple addon', function (t) {
   proc.stderr.setEncoding('utf-8')
 })

+test('build simple addon - teardown', cleanup(path.join(addonPath, 'build')))
+
+test('build with different build dir - setup', cleanup(path.join(addonPath, 'foo')))
+
 test('build with different build dir', function(t) {
   var cmd = [nodeGyp, 'rebuild', '-C',
               addonPath, '--loglevel=verbose', '--build-dir=foo']
@@ -39,7 +57,7 @@ test('build with different build dir', function(t) {
     t.strictEqual(lastLine, 'gyp info ok', 'should end in ok')

     try {
-      var binding = require('hello_world')
+      var binding = require(path.join(addonPath, 'foo/Release/hello.node'))
       t.strictEqual(binding.hello(), 'world')

       // Cleanup
@@ -52,3 +70,5 @@ test('build with different build dir', function(t) {
   proc.stdout.setEncoding('utf-8')
   proc.stderr.setEncoding('utf-8')
 })
+
+test('build with different build dir - teardown', cleanup(path.join(addonPath, 'foo')))
```

Fishrock123:
ping @peterbraden

peterbraden:
Sorry, been busy with a new job. Done now, and squashed and rebased.

Thanks!

test/test-addon.js
```diff
@@ -26,3 +42,33 @@ test('build simple addon', function (t) {
...
+    var lastLine = logLines[logLines.length-1]
+    t.strictEqual(err, null)
+    t.strictEqual(lastLine, 'gyp info ok', 'should end in ok')
+
+    try {
+      var binding = require(path.join(addonPath, 'foo/Release/hello.node'))
+      t.strictEqual(binding.hello(), 'world')
+
+      // Cleanup
+      exec('node-gyp --build-dir=foo clean', t.end)
```
bnoordhuis: Can you use `execFile(process.execPath, [nodeGyp, ...])` here?  Calling out to node-gyp risks picking up the version that is on the path, not what's under test.

Aside: I know they're Rod's work but can you wrap lines at 80 columns?  Thanks. :-)

peterbraden:
Updated, squashed

Fishrock123:
CI failed on windows, @bnoordhuis could you take a look?

richardlau:
Test failures on Windows are because the newly introduced `cleanup()` function in the tests is trying to remove a currently loaded .node file. Even if the module is deleted from `require.cache` the .node is not unloaded (Is it even possible to unload a required .node in Node.js?).

bnoordhuis:
> (Is it even possible to unload a required .node in Node.js?)

No.

richardlau:
I don't think clearing entries from `require.cache` is needed, because the second `require()` call that was introduced to the test is requiring `hello.node` from a different location (`path.join(addonPath, 'foo/Release/hello.node')`) than `hello.node` that is required from `require('hello_world')` (via the `bindings` module) so would end up as a separate entry in the cache. 

If the fear is contamination of the second test from the first, then perhaps the new test could be in a separate file? Or does `tape` run all tests in the same process?

bnoordhuis:
@peterbraden Are you still interested in pursuing this?  From https://github.com/nodejs/node-gyp/pull/919#issuecomment-266825606 it sounds like it's almost there, it just needs one or two small fix-ups.

> If the fear is contamination of the second test from the first, then perhaps the new test could be in a separate file? Or does tape run all tests in the same process?

I think tape uses a process-per-file model so that sounds like a good idea.

peterbraden:
Sure, let me try when I have some free time, maybe this weekend.

bnoordhuis:
I found out the hard way today that tape runs all tests in the same process.  I've filed #1123.

599316527:
Has this feature been merged into master? I really need it.

jlipps:
@peterbraden I'd still be interested in this patch if you want to resurrect it :-)

rvagg:
if @peterbraden or someone else wants to pick this up, please either rebase this or take the code and open a new PR against the current master
test failures because of the way `tape` works should be fixed once we migrate to `tap` which I hope we can achieve soon.

peterbraden:
I don't have time for this any more, hopefully someone else can take this on.

bnoordhuis:
It's with slightly sad eyes that I'm closing this out. I just closed out #263 after no one stepped forward. Thanks anyway, @peterbraden!

crowder-pw:
I'm interested in helping to get this issue over the line, is the "tap/tape" thing mentioned above still an issue? What remains to be done here.

cclauss:
https://github.com/nodejs/tap2junit

PR #927, opened 2016-05-13
Add support for a custom release name when downloading headers

MarshallOfSound:
This fixes an underlying issue which caused this issue over at `electron-rebuild` https://github.com/electron/electron-rebuild/issues/66

Basically `node-gyp` assumes that any version in the semver major range `1 --> 4` is `iojs`.  Recently electron hit a major version bump to `1.x.y`.  This caused `node-gyp` to start attempting to download `iojs` headers when they are in fact still node headers but the version number tracks electron not node.

This PR adds a `releaseName` parameter that will allow `electron-rebuild` (and other tools) to specify their own release names thus allowing electron to force the name to be "node" regardless of the major version they choose to use.

### DIFF
--- a/lib/node-gyp.js
+++ b/lib/node-gyp.js
@@ -99,6 +99,7 @@ proto.configDefs = {
   , 'tarball': String // 'install'
   , jobs: String      // 'build'
   , thin: String      // 'configure'
+  , releaseName: String // 'build'
 }
 
 /**
@@ -222,4 +223,3 @@ Object.defineProperty(proto, 'version', {
     }
   , enumerable: true
 })
-
--- a/lib/process-release.js
+++ b/lib/process-release.js
@@ -9,16 +9,16 @@ var semver = require('semver')
   , bitsreV3 = /\/win-(x86|ia32|x64)\// // io.js v3.x.x shipped with "ia32" but should
                                         // have been "x86"
 
-// Captures all the logic required to determine download URLs, local directory and 
+// Captures all the logic required to determine download URLs, local directory and
 // file names. Inputs come from command-line switches (--target, --dist-url),
 // `process.version` and `process.release` where it exists.
 function processRelease (argv, gyp, defaultVersion, defaultRelease) {
   var version = (semver.valid(argv[0]) && argv[0]) || gyp.opts.target || defaultVersion
     , versionSemver = semver.parse(version)
     , overrideDistUrl = gyp.opts['dist-url'] || gyp.opts.disturl
+    , name = gyp.opts.releaseName
     , isDefaultVersion
     , isIojs
-    , name
     , distBaseUrl
     , baseUrl
     , libUrl32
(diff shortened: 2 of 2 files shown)


bnoordhuis:
I'm not a fan of adding flags willy-nilly.  If a flag is going to be necessary anyway in order to download the right files, is there a reason you can't use `--dist-url`?

Does Electron define `process.release`?  That would obviate the need for flags, new or existing.  node-gyp can then easily be taught where to look for files.

Aside: whatever direction this PR goes, it will need regression tests.

MarshallOfSound:
> is there a reason you can't use --dist-url?  

@bnoordhuis The `--dist-url` is already used but the `name` variable is automatically placed in the file name by the `resolveLibUrl` function.

The URL is turned into something that looks like this

```
https://atom.io/download/atom-shell/v1.0.1/node-v1.0.1.tar.gz
```

The **node** part of that URL is the `name` parameter.

> I'm not a fan of adding flags willy-nilly  

Yeah, I don't normally like adding flags randomly.  But it appears that with the current API literally every part of the URL is changeable with flags **apart** from that `name` variable.

> Does Electron define process.release? That would obviate the need for flags, new or existing. node-gyp can then easily be taught where to look for files.  

Currently the `electron-rebuild` tool is still used from `nodejs` not from within electron.  So although electron does set the `process.release` variable correctly when we spawn a `node-gyp` process it is using node's `process.release` values.

bnoordhuis:
Okay, I see.  In that case I think it's better to add a `--runtime=node|electron|node-webkit` switch because that's what node-pre-gyp uses and what people are used to passing to npm.

MarshallOfSound:
Would `io.js` be an option in that switch as well?  If that's the way you feel it should go I'll redo the PR :+1:

bnoordhuis:
I'd leave io.js out.  `npm install --runtime=iojs` would presumably confuse node-pre-gyp no end.

MarshallOfSound:
@bnoordhuis Interesting thought though.  I'm not sure how `node-webkit` is used but I know that electron hosts there headers with the `node` name so that they worked with `node-gyp`'s logic.

Should the electron runtime set `name` to be `electron` or `node`?  

Setting to `electron` although semantically correct won't fix the problem.  It just means `electron-rebuild` will have to call `node-gyp` with `--runtime=node`.

Do you think this is still a good solution to the issue?

bnoordhuis:
> electron hosts there headers with the node name so that they worked with node-gyp's logic

I'm confused now.  Doesn't that mean `--dist-url` is sufficient?

(Looks like it - `https://atom.io/download/atom-shell/v1.0.1/node-v1.0.1.tar.gz` exists according to `curl --head`.)

MarshallOfSound:
@bnoordhuis Sorry, that URL was the one I copied from my patched version of `node-js`.  If I revert my changes the URL that `node-gyp` attempts to use is.

```
https://atom.io/download/atom-shell/v1.0.1/iojs-v1.0.1.tar.gz
```

However it appears that someone on the electron team has pulled off some server magic so that that URL now resolves to the same thing that `node-` resolves to.  I'm going to close this as the issue that this is intended to resolve has been fixed :+1:

PR #930, opened 2016-05-17
More verbose error on locating msbuild.exe failure

purplecode:
The first error usage was misleading. In my case the error was saying 'Access denied' and it was because of limited privileges, not because of missing VS installation. It was also hard to debug because   one of the deeply nested dependencies was failing.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -125,19 +125,19 @@ function build (gyp, argv, callback) {
 
   function findMsbuild () {
     log.verbose('could not find "msbuild.exe" in PATH - finding location in registry')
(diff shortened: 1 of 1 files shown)


bnoordhuis:
LGTM but you show up as "purplecode" in the commit's Author field.  Want me to land it under your real name?

Also, if you can reword the commit log so it conforms a bit better to the general format we use, that would be great.

purplecode:
I rewrote the author field and commit message in my commit

bnoordhuis:
Thanks Mateusz, landed in a4fa07b.  I massaged the commit log's status line so it fit in 50 columns.

PR #935, opened 2016-05-19
gyp: improvements for android generator

robertchiras:
When generating Android.mk files for HOST binaries, we need to specify
the LOCAL_CXX_STL to use when linking.
Also, set the variable GYP_HOST_MULTILIB to 'first'. This is used as
LOCAL_MULTILIB, and since we don't provide flags for both types of
builds (32-bit and 64-bit), the build system will try to build both and
fail.
More details about LOCAL_MULTILIB can be found here:
https://source.android.com/source/64-bit-builds.html

Signed-off-by: Robert Chiras <robert.chiras@intel.com>


Fishrock123:
cc @bnoordhuis?

bnoordhuis:
Rubber-stamp LGTM.  Landed in 7f1c1b9, thanks Robert.

Issue #941, opened 2016-05-25
Upgrade to glob 7.0.3 to solve deprecated graceful-fs@1.2.3 warnings


superclarkk:
Currently, npm gives the following warning when installing the latest version of node-gyp:

~~~
npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
~~~

This is because glob is not at the latest version.

~~~
  └─┬ node-gyp@3.3.1 (latest version)
    └─┬ glob@3.1.21 (not latest version - should be 7.0.3) 
      └── graceful-fs@1.2.3
~~~

PR #943, opened 2016-05-25
Upgrade to glob@7.0.3.

bnoordhuis:
Matches what npm v2.15.5 and npm v3.8.6 use.

Fixes: https://github.com/nodejs/node-gyp/issues/941

R=@rvagg?

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -7,7 +7,6 @@ module.exports.test = { findPython: findPython }
 
 var fs = require('graceful-fs')
   , path = require('path')
-  , glob = require('glob')
   , log = require('npmlog')
   , osenv = require('osenv')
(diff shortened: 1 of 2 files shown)


Fishrock123:
lgtm if it works I suppose. 

Tests pass here.

rvagg:
Yeah, LGTM.

I can't find info on what broken between `3||4` and `7` but given that our only single usage is to look up `'build/*.sln'` I can't imagine we could go wrong anywhere.

While you're in there, you could remove the `glob` require in lib/configure.js, it's not used.

bnoordhuis:
Landed in ac29d23 with the usage in lib/configure.js removed.  Thanks for the reviews.

PR #945, opened 2016-05-27
Issue 297 fixed

roman-yerin:


bnoordhuis:
I appreciate you took the time to write up a patch, but as a rule, we don't except changes to code in gyp/ because those files come from the upstream gyp project.  For easy_xml.py, that rule has been bent in the past but I don't want to make a habit of that.

If you can get upstream to accept the patch, I'll be happy to cherry-pick it and do a release.

bnoordhuis:
Seeing how there has been no follow-up, I'll go ahead and close the pull request.

PR #948, opened 2016-06-05
Disable strict-ssl mode when a custom dist-url is provided…

demkada:
Since we provide a custom HTTPS URL for headers tarball, we take responsibility of the provided server certificate validity. 

This PR is for fixing an issue I have in my corporate network and I think it can fix a lot of issues related to headers download in a corporate network which are opened in other modules based on node-gyp.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -452,6 +452,12 @@ function download (gyp, env, url) {
     }
   }
 
+  //strictSSL Disabled for custom dist-url
+  var customDistUrl = gyp.opts['dist-url'] || gyp.opts.disturl;
+  if (customDistUrl) {
+    requestOpts.strictSSL = false
+  }
+
   var req = request(requestOpts)
   req.on('response', function (res) {
     log.http(res.statusCode, url)
--- a/test/test-download.js
+++ b/test/test-download.js
@@ -82,6 +82,50 @@ test('download over https with custom ca', function (t) {
   })
 })
 
+test('download over https with custom dist-url', function (t) {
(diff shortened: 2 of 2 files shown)


bnoordhuis:
Sorry, but I don't think this is a good change.  If you really want to use a HTTPS endpoint without a valid certificate, you can set `NODE_TLS_REJECT_UNAUTHORIZED=0` in the environment.

You might as well use HTTP though, it's just as secure.

demkada:
Indeed, But apply NODE_TLS_REJECT_UNAUTHORIZED=0 in process env can compromise the whole installation process, because, as I think, this config is injected in all https requests going from the process. 

The idea behind this PR, is to allow desactivation of strcit-ssl mode only on https requests made by node-gyp when a custom HTTPS DIST-URL is provided for downloading headers tarball.

In our case, all of our servers are accessible only  throughTLS with self signed certificates.

Maybe this feature is not neccessary. In that case, how can we handle this kind of problem with node-gyp and custom dist-url in selt signed certificate without compromising our installation process?

bnoordhuis:
You could set up a self-signed CA that issues a certificate for your dist-url endpoint.  You then pass the CA certificate to node-gyp with the `--cafile` switch.  (You can also set it in your .npmrc, of course.)

I did mean it though when I wrote "You might as well use HTTP though, it's just as secure."  Self-signed certificates don't offer protection against man-in-the-middle attacks.  There really is no point in using them, they just give a false sense of security.

PR #950, opened 2016-06-07
Allow for npmlog@3.x

iarna:
npmlog version 3 switches to gauge version 2.  The interface change is in
setGaugeTemplate, which node-gyp does not use.

(Allowing version 3, will let npm dedupe npmlog. As it is, we'll have two copies.)


bnoordhuis:
Thanks, landed in 3df8222.

PR #951, opened 2016-06-09
Copying node.lib timestamp to prevent redundant link with msbuild

idobh2:
I noticed node-gyp build always creates a new .node binary even when no code/lib were changed.
Turns out that the fact that node.lib copy is created by writing the file each time, which changes its timestamp, is causing msbuild.exe to trigger a link operation.
writing the file, and copying the timestamp resolves this issue.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -200,7 +200,15 @@ function build (gyp, argv, callback) {
       rs.pipe(ws)
       rs.on('error', callback)
       ws.on('error', callback)
-      rs.on('end', doBuild)
+      rs.on('end', function(){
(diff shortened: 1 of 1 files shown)


pmed:
It seems there is no need to copy `node.lib` at all. Such node libs are already lie in `$node_root_dir/[ia32|x64]` directories. We can use a `target_arch` variable declared in `build/config.gypi` to link an appropriate node.lib from there.

I have such a change in PR #962@21eee45ba5e54359fb6dabdd8d7bac1a91e49936

joaocgreis:
https://github.com/nodejs/node-gyp/pull/964 landed addressing this

Issue #949, opened 2016-06-06
error compiling win_delay_load_hook.c with visual studio preview 2


xtaltas:
Apparently <delayimp.h> header changed since visual studio 2015 update  3 setting const to PfnDliHook

// Prior to Visual Studio 2015 Update 3, these hooks were non-const.  They were
// made const to improve security (global, writable function pointers are bad).
// If for backwards compatibility you require the hooks to be writable, define
// the macro DELAYIMP_INSECURE_WRITABLE_HOOKS prior to including this header and
// provide your own non-const definition of the hooks.
 
Probably should include this in win_delay_load_hook.c source file
 
#ifndef DELAYIMP_INSECURE_WRITABLE_HOOKS
#define DELAYIMP_INSECURE_WRITABLE_HOOKS
#endif

xtaltas:
Yes apparently it looks fine

Christian Taltas
KPMG Technologies Services - Directeur Général
ctaltas@kpmg.fr<mailto:ctaltas@kpmg.fr>
+44 7482003570<tel:+44%207482003570>




On Mon, Jun 6, 2016 at 10:28 AM -0700, "Ben Noordhuis" <notifications@github.com<mailto:notifications@github.com>> wrote:


Would const-ifying __pfnDliNotifyHook2 in src/win_delay_hook.c work?

/cc @piscisaureus<https://github.com/piscisaureus>

-

piscisaureus:
No objections to const-ification from me.

bnoordhuis:
https://ci.nodejs.org/job/node-test-commit-windows/975/ <- quick test

bnoordhuis:
Okay, it clearly doesn't work with older VS versions.  I'm not going to look into it deeper but I'll be happy to review patches.

piscisaureus:
@orangemocha you have time to dig into this?

orangemocha:
I am investigating...

orangemocha:
https://github.com/nodejs/node-gyp/pull/952

RanzQ:
Just installed the latest [Visual C++ Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools) and getting the same error with `node-gyp@3.4.0`.

bnoordhuis:
@RanzQ Can you file a new issue and include the build log?

RanzQ:
[...]

PR #952, opened 2016-06-09
win: work around __pfnDliNotifyHook2 type change

orangemocha:
Visual Studio 2015 Update 3 defines __pfnDliNotifyHook2 as const.

Fixes: https://github.com/nodejs/node-gyp/issues/949

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -43,7 +43,7 @@
         'conditions': [
           [ 'OS=="win"', {
             'sources': [
-              '<(node_gyp_dir)/src/win_delay_load_hook.c',
+              '<(node_gyp_dir)/src/win_delay_load_hook.cc',
             ],
             'msvs_settings': {
               'VCLinkerTool': {
--- a/src/win_delay_load_hook.cc
+++ b/src/win_delay_load_hook.cc
@@ -31,6 +31,6 @@ static FARPROC WINAPI load_exe_hook(unsigned int event, DelayLoadInfo* info) {
   return (FARPROC) m;
 }
 
-PfnDliHook __pfnDliNotifyHook2 = load_exe_hook;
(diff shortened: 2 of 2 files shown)


bnoordhuis:
I don't think using decltype is acceptable.  node-gyp needs to work with VS 2010 and 2012 (because of node v0.10 and v0.12) and those don't support decltype, AFAIK.

orangemocha:
@bnoordhuis it is supported since VS 2010: https://msdn.microsoft.com/en-us/library/dd537655(v=vs.100).aspx

I just tested it with VS 2010 and it compiles.

bnoordhuis:
LGTM in that case.  As a quick sanity check, can you land it in deps/npm/node_modules/node-gyp and see if the Windows CI is green?

orangemocha:
CI has failures and they clearly look related: https://ci.nodejs.org/job/node-test-binary-windows/2483/RUN_SUBSET=0,VS_VERSION=vs2015,label=win2012r2/console

I will keep investigating.

orangemocha:
Those errors were due to win_delay_load_hook.c being compiled as C, while decltype is C++.

I updated this PR to also change the file type to .cpp. Tested in Node in this run and it's green: https://ci.nodejs.org/job/node-test-commit-windows-fanned/2929/

bnoordhuis:
.cc but otherwise LGTM.

orangemocha:
Updated to use .cc. I tested it locally with vcbuild test-addons and also in VS 2010 to be sure. All ok.

orangemocha:
@bnoordhuis what is the process for getting this into Node and npm?

Fishrock123:
@orangemocha We just need to release a version.

Fishrock123:
https://github.com/nodejs/node-gyp/issues/956

PR #953, opened 2016-06-09
Move VC++ Build Tools to Build Tools landing page

AndrewPardoe:
We've (finally!) created an actual landing page for the VC++ Build Tools. We'd like your readme to refer people to the landing page rather than directly to the download. This change moves the target of that link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ You will also need to install:
       * This step will install `gcc` and the related toolchain containing `make`
   * On Windows:
     * Visual C++ Build Environment:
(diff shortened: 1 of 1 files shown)


bnoordhuis:
I get a 404 when I open http://landinghub.visualstudio.com/visual-cpp-build-tool.

AndrewPardoe:
I deleted an 's' at the end of the URL. It's "tools". Sorry, I'll fix that in an hour or so.

orangemocha:
The new URL seems correct. LGTM

bnoordhuis:
Landed in af35b2a, thanks Andrew.

PR #955, opened 2016-06-13
Replace fs.accessSync calls

richardlau:
9bfa0876b46764978492fffea074d2d7aa8954f9 (#753) added calls to `fs.accessSync` on AIX to test if the candidate exports file exists and is readable. Unfortunately `accessSync` does not exist in Node 0.10.x, so the exports file is not found and results in a linking error:
```
ld: 0706-003 Cannot find or read import file: /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/node.exp
        ld:accessx(): A file or directory in the path name does not exist.
collect2: ld returned 255 exit status
gmake: *** [Release/obj.target/hello.node] Error 1
gyp ERR! build error
gyp ERR! stack Error: `gmake` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/home/users/riclau/sandbox/github/node-gyp/lib/build.js:276:23)
gyp ERR! stack     at ChildProcess.emit (events.js:98:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:820:12)
gyp ERR! System AIX 1
```

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -1,5 +1,6 @@
 module.exports = exports = configure
-module.exports.test = { findPython: findPython }
+module.exports.test = { findAccessibleSync: findAccessibleSync, 
+  findPython: findPython }
 
 /**
  * Module dependencies.
@@ -20,6 +21,7 @@ var fs = require('graceful-fs')
   , execFile = cp.execFile
   , win = process.platform == 'win32'
   , findNodeDirectory = require('./find-node-directory')
+  , msgFormat = require('util').format
 
 exports.usage = 'Generates ' + (win ? 'MSVC project files' : 'a Makefile') + ' for the current module'
 
@@ -226,22 +228,21 @@ function configure (gyp, argv, callback) {
     //  - the out/Release directory
     //  - the out/Debug directory
     //  - the root directory
-    var node_exp_file = ''
+    var node_exp_file = undefined
     if (process.platform === 'aix') {
       var node_root_dir = findNodeDirectory()
       var candidates = ['include/node/node.exp',
                         'out/Release/node.exp',
                         'out/Debug/node.exp',
                         'node.exp']
-      for (var next = 0; next < candidates.length; next++) {
-         node_exp_file = path.resolve(node_root_dir, candidates[next])
-         try {
-           fs.accessSync(node_exp_file, fs.R_OK)
-           // exp file found, stop looking
-           break
-         } catch (exception) {
-           // this candidate was not found or not readable, do nothing
-         }
+      var logprefix = 'find exports file'
+      node_exp_file = findAccessibleSync(logprefix, node_root_dir, candidates)
+      if (node_exp_file !== undefined) {
+        log.verbose(logprefix, 'Found exports file: %s', node_exp_file)
+      } else {
+        var msg = msgFormat('Could not find node.exp file in %s', node_root_dir)
+        log.error(logprefix, 'Could not find exports file')
+        return callback(new Error(msg))
       }
     }
 
@@ -308,6 +309,29 @@ function configure (gyp, argv, callback) {
     }
   }
 
+}
+
+/**
+ * Returns the first file or directory from an array of candidates that is 
+ * readable by the current user, or undefined if none of the candidates are
+ * readable. 
+ */
+function findAccessibleSync (logprefix, dir, candidates) {
+  for (var next = 0; next < candidates.length; next++) {
+     var candidate = path.resolve(dir, candidates[next])
+     try {
+       var fd = fs.openSync(candidate, 'r')
+     } catch (e) {
+       // this candidate was not found or not readable, do nothing
+       log.silly(logprefix, 'Could not open %s: %s', candidate, e.message)
+       continue
+     }
+     fs.closeSync(fd)
+     log.silly(logprefix, 'Found readable %s', candidate)
+     return candidate
+  }
+
+  return undefined
 }
 
 function findPython (python, callback) {
--- a/test/node_modules/hello_world/hello.js
+++ b/test/node_modules/hello_world/hello.js
@@ -1,0 +1,3 @@
+'use strict'
+var addon = require('bindings')('hello');
+exports.hello = function() { return addon.hello() }
--- a/test/test-addon.js
+++ b/test/test-addon.js
@@ -1,0 +1,28 @@
+'use strict'
+
+var test = require('tape')
+var execFile = require('child_process').execFile
+var path = require('path')
+var addonPath = path.resolve(__dirname, 'node_modules', 'hello_world')
+var nodeGyp = path.resolve(__dirname, '..', 'bin', 'node-gyp.js')
+
+test('build simple addon', function (t) {
+  t.plan(3)
+
+  // Set the loglevel otherwise the output disappears when run via 'npm test'
+  var cmd = [nodeGyp, 'rebuild', '-C', addonPath, '--loglevel=verbose']
+  var proc = execFile(process.execPath, cmd, function (err, stdout, stderr) {
+    var logLines = stderr.toString().trim().split(/\r?\n/)
+    var lastLine = logLines[logLines.length-1]
+    t.strictEqual(err, null)
+    t.strictEqual(lastLine, 'gyp info ok', 'should end in ok')
+    try {
+      var binding = require('hello_world')
+      t.strictEqual(binding.hello(), 'world')
+    } catch (error) {
+      t.error(error, 'load module')
+    }
+  })
+  proc.stdout.setEncoding('utf-8')
+  proc.stderr.setEncoding('utf-8')
+})
--- a/package.json
+++ b/package.json
@@ -41,7 +41,9 @@
     "node": ">= 0.8.0"
   },
   "devDependencies": {
-    "tape": "~4.2.0"
+    "tape": "~4.2.0",
+    "bindings": "~1.2.1",
+    "nan": "^2.0.0"
   },
   "scripts": {
     "test": "tape test/test-*"
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
   "devDependencies": {
     "tape": "~4.2.0",
     "bindings": "~1.2.1",
-    "nan": "^2.0.0"
+    "nan": "^2.0.0",
+    "require-inject": "~1.3.0"
   },
   "scripts": {
     "test": "tape test/test-*"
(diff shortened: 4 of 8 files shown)


richardlau:
I've replaced the call to `fs.accessSync` with `fs.statSync` since that does exist on Node 0.10.x (and above) and the alternative `fs.existsSync` is deprecated on later versions of Node.

There also didn't appear to be an existing test that failed because of this, so I've added one in a separate commit.

richardlau:
cc @mhdawson

mhdawson:
Is this page wrong

https://nodejs.org/api/fs.html#fs_fs_fstat_fd_callback

It would seem to say that fs.accessSync was added in v0.1.93  ?

The doc is also not all that clear for fs.fstatSync.  Would it through if the file existed by was not readable to the current user ?

richardlau:
Yes, looks like the page you have linked to is incorrect, there's no `accessSync` in 0.10.x, e.g. https://nodejs.org/dist/v0.10.45/docs/api/fs.html. 

Just tried removing read permissions from the exports file -- `fstatSync` doesn't throw (so the candidate is chosen) but the subsequent `gyp` invocation results in a similar access error as originally:
```
-bash-4.2$ chmod u-r /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/include/node/node.exp
-bash-4.2$ tape test/test-addon.js
TAP version 13
# build simple addon
not ok 1 should be equal
  ---
    operator: equal
    expected: |-
      null
    actual: |-
      { [Error: Command failed: gyp info it worked if it ends with ok
...
ld: 0706-003 Cannot find or read import file: /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/include/node/node.exp
        ld:accessx(): The file access permissions do not allow the specified action.
collect2: ld returned 255 exit status
gmake: *** [Release/obj.target/hello.node] Error 1
gyp ERR! build error
gyp ERR! stack Error: `gmake` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/home/users/riclau/sandbox/github/node-gyp/lib/build.js:276:23)
gyp ERR! stack     at ChildProcess.emit (events.js:98:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:820:12)
gyp ERR! System AIX 1
...
```
This is better than the original case, because the file pointed to is there on the filesystem but with insufficient access permissions (in the original the code fell through to the last tried candidate location which turned out to not exist).

Flarna:
I think you have to check the mode field of the stat result for read permissions to reach the same result as with access. Or you could try to read at least a part of the file.

mhdawson:
Right unless we find a better check we'll be making it worse for versions other than 0.10.X.  Can you look into how we can achieve a check which is closer to the original.

test/node_modules/hello_world/hello.js
```diff
@@ -0,0 +1,3 @@
+var addon = require('bindings')('hello');
+exports.hello = function() { return addon.hello() }
+
```
bnoordhuis: Can you drop the blank line and add `'use strict'` at the top?

test/test-addon.js
```diff
@@ -0,0 +1,26 @@
...
+var execFile = require('child_process').execFile
+var path = require('path')
+var addonPath = path.resolve(__dirname, 'node_modules', 'hello_world')
+var nodeGyp = path.resolve(__dirname, '..', 'bin', 'node-gyp.js')
+
+test('build simple addon', function (t) {
+  t.plan(3)
+
+  // Set the loglevel otherwise the output disappears when run via 'npm test'
+  var proc = execFile(process.execPath, [nodeGyp, 'rebuild', '-C', addonPath, '--loglevel=verbose'], function (err, stdout, stderr) {
```
bnoordhuis: Can you try keeping this line < 80 columns, e.g., by declaring the args aray as a variable first?

gibm:
@mhdawson I would guess that the easiest solution would be to use `fs.statSync` if `fs.accessSync` is undefined (i.e. in 0.10). That way we'd be the same for 0.12 and up, and for 0.10 there'd only be a problem if someone didn't have read permissions on a node.exp file in their node directory on AIX. 

Would that make sense?

Presumably the alternative is for @richardlau to implement something that mimics the functionality of `fs.accessSync` on v0.10. Given that `fs.accessSync` wasn't backported to 0.10, I'd suggest that this is overkill.

gibm:
@bnoordhuis is there currently a test in node-gyp that builds a module?

richardlau:
Updated the test commit to address comments from @bnoordhuis. I found another line that was over 80 characters so have shortened that one (via another var) as well.

Also updated the fix so that it checks the mode based on the uid and gid of the file being tested to check if the file is readable based on comments from @mhdawson and @Flarna. I also decided to add extra logging to this area, including one at log level `error` if the code could not find a suitable exports file to use.

lib/configure.js
```diff
@@ -233,16 +233,45 @@ function configure (gyp, argv, callback) {
...
+               log.silly(prefix, '%s has others read permission', node_exp_file)
+               found = true
+               break
+             }
+           }
+           log.silly(prefix, 
+             '%s exists but has no read permissions', node_exp_file)
          } catch (exception) {
            // this candidate was not found or not readable, do nothing
+           log.silly(node_exp_file, 'does not exist')
```
richardlau: Missed this, should be using prefix like other log messages. Will fix tomorrow.

lib/configure.js
```diff
@@ -233,16 +233,45 @@ function configure (gyp, argv, callback) {
...
            // this candidate was not found or not readable, do nothing
+           log.silly(prefix, '%s does not exist', node_exp_file)
          }
       }
+      if (found) {
+        log.verbose(prefix, 'Found exports file: %s', node_exp_file)
+      } else {
+        log.error(prefix, 'Could not find exports file')
+      }
     }
```
bnoordhuis: Can you break out this block into a separate function?

Also, the try body should be smaller.  It should basically just be the `fs.statSync()` call, nothing more, else there is too much space for bugs to hide.

richardlau: Yep, okay will do.

richardlau:
Extracted function as requested by @bnoordhuis. Added tests for extracted function. 

I've also made the code bail out with an error if it can't find a readable exports file (before it would pass the last tried candidate on to gyp even though we know it isn't readable). CC @mhdawson

lib/configure.js
```diff
@@ -310,6 +311,50 @@ function configure (gyp, argv, callback) {
...
+         log.silly(logprefix, '%s has others read permission', candidate)
+         found = candidate
+         break
+       }
+     }
+     log.silly(logprefix, '%s exists but has no read permissions', candidate)
+  }
+
+  return found
+}
```
bnoordhuis: Question: if the sole purpose is to find out if the file can be read from, why don't you try to open it in read-only mode?  It's shorter and less error-prone, IMO.
```js
for (var i = 0; i < candidates.length; i += 1) {
  var filename = path.resolve(candidates[i])
  try {
    var fd = fs.openSync(filename, 'r')
  } catch (e) {
    continue
  }
  fs.closeSync(fd)
  return filename
}
```

richardlau: Fair point, that's much more concise. Should it still be broken out into its own function, or in-lined as it was originally with the fs.accessSync call?

bnoordhuis: I'd keep it in a separate function, easier to test in isolation.

richardlau:
Code and tests reworked to use `fs.openSync`. Verified the tests pass on Windows, Linux (x64) and AIX with Node v0.10.45.

richardlau:
So with the current version of the changes, if I sabotage my AIX Node installation, `chmod u-r /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/include/node/node.exp`, I get:

```
-bash-4.2$ /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/bin/node  "/home/users/riclau/sandbox/github/node-gyp/bin/node-gyp.js" rebuild -C /home/users/riclau/sandbox/github/node-gyp/test/node_modules/hello_world
gyp info using node-gyp@3.3.1
gyp info using node@0.10.45 | aix | ppc64
gyp info chdir /home/users/riclau/sandbox/github/node-gyp/test/node_modules/hello_world
gyp ERR! find exports file Could not find exports file
gyp ERR! configure error
gyp ERR! stack Error: Could not find node.exp file in /tmp/usenode.riclau/node-v0.10.45-aix-ppc64
gyp ERR! stack     at runGyp (/home/users/riclau/sandbox/github/node-gyp/lib/configure.js:245:25)
gyp ERR! stack     at findConfigs (/home/users/riclau/sandbox/github/node-gyp/lib/configure.js:170:23)
gyp ERR! stack     at /home/users/riclau/sandbox/github/node-gyp/lib/configure.js:176:11
gyp ERR! stack     at Object.oncomplete (fs.js:108:15)
gyp ERR! System AIX 1
gyp ERR! command "/tmp/usenode.riclau/node-v0.10.45-aix-ppc64/bin/node" "/home/users/riclau/sandbox/github/node-gyp/bin/node-gyp.js" "rebuild" "-C" "/home/users/riclau/sandbox/github/node-gyp/test/node_modules/hello_world"
gyp ERR! cwd /home/users/riclau/sandbox/github/node-gyp/test/node_modules/hello_world
gyp ERR! node -v v0.10.45
gyp ERR! node-gyp -v v3.3.1
gyp ERR! not ok
-bash-4.2$
```

With `--loglevel=silly`, we get these extra diagnostics:
```
gyp sill find exports file Could not open /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/include/node/node.exp: EACCES, permission denied '/tmp/usenode.riclau/node-v0.10.45-aix-ppc64/include/node/node.exp'
gyp sill find exports file Could not open /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/out/Release/node.exp: ENOENT, no such file or directory '/tmp/usenode.riclau/node-v0.10.45-aix-ppc64/out/Release/node.exp'
gyp sill find exports file Could not open /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/out/Debug/node.exp: ENOENT, no such file or directory '/tmp/usenode.riclau/node-v0.10.45-aix-ppc64/out/Debug/node.exp'
gyp sill find exports file Could not open /tmp/usenode.riclau/node-v0.10.45-aix-ppc64/node.exp: ENOENT, no such file or directory '/tmp/usenode.riclau/node-v0.10.45-aix-ppc64/node.exp'
gyp ERR! find exports file Could not find exports file
```

bnoordhuis:
Thanks Richard, LGTM.  Landed in c4344b3...77383dd.

mhdawson:
A bit late I guess but LGTM from me as well

PR #937, opened 2016-05-23
doc: add command options to README.md

gibm:
Documents options accepted by node-gyp.
Continuation of #880.

@bnoordhuis

PR #958, opened 2016-06-17
doc: add --target option to README

gibm:
Missed this in #937 

If there are any other changes that need making to the README (or other documentation) I can add them in, this is a bit small by itself.


bnoordhuis:
Thanks, landed in c4344b3.

Issue #960, opened 2016-06-17
Consider gyp.js?


indutny:
Hello everyone!

We're working on https://github.com/indutny/gyp.js , which is a port of GYP to Node.js. I think the project is reaching the point when it could be actually used as an alternative "backend" for `node-gyp`.

What do you think about bundling it here under a flag to see how well it will work?

Thank you,
Fedor.

Fishrock123:
> What do you think about bundling it here under a flag to see how well it will work?

I think that is an excellent starting point.

I suppose a flag would just change the child-process invocation to gyp.js?

Not sure I'm the best person to lead on this though, I'm pretty pre-occupied.

indutny:
@Fishrock123 yep, the flag will just do this.

indutny:
Though, important thing about `gyp.js` is that it may not work on older versions of node, because of extensive ES6 usage. Is this going to be a problem for us, or can we manage it through semver?

Fishrock123:
@indutny  We might be able to transpile it back. Does it work on v4 or just v6?

indutny:
Just v6 at the moment.

pmed:
@indutny See PR: #962

indutny:
@rvagg sorry, but I have feeling that we are not moving fast enough on this. Is there any way to expedite the review process? Should we invite more collaborators to this project?

lazaridis-com:
@indutny , strange thing. Wondering why your work is not give more priority.

prasannavl:
It's been quite a while. Any update on the state of things?

refack:
I'll follow up on #962

stalkerg:
Good news: https://chromium-review.googlesource.com/c/479691

JamesMGreene:
@stalkerg: Hmm, maybe, but AFAICT, that:
 - was last updated in April,
 - has merge conflicts,
 - has not been reviewed, and
 - has not been merged.

😢

stalkerg:
ah... but I hope this problem is will fix

refack:
> Good news: https://chromium-review.googlesource.com/c/479691

@stalkerg thank you for bringing this to attention

zixia:
+1

Pomax:
[...]

PR #962, opened 2016-06-19
Initial gyp.js support

pmed:
Added a command-line switch `--gypjs` to use https://github.com/indutny/gyp.js instead of gyp.

Removed copyNodeLib() on Windows in favor to use `target_arch` variable from generated configuration file `build/config.gypi`

Fixes: #960

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,37 @@ you can require the `.node` file with Node and run your tests!
 __Note:__ To create a _Debug_ build of the bindings file, pass the `--debug` (or
 `-d`) switch when running either the `configure`, `build` or `rebuild` command.
 
+#### Usage without Python
+
+There is a [`gyp.js`](https://github.com/indutny/gyp.js/) project, a GYP
+implementation in JavaScript. It generates [`Ninja`](https://ninja-build.org/)
+build files and requires no Python installation.
+
+In this case you will need to install [`Ninja`](https://ninja-build.org/) build
+tool and a C/C++ compiler toolchain.
+
+To generate projects files with `gyp.js` instead of `gyp` and to build them with
+`ninja`, please supply a command-line option `--gypjs` in `node-gyp` command.
+
+``` bash
+$ node-gyp configure --gypjs
+```
+
+It is also possible to set `npm_config_gypjs` environment variable to turn on
+the `gyp.js` usage.
+
+``` bash
+$ npm_config_gypjs=1 node-gyp build
+```
+
+Path to an existing `ninja` installation can be set with a `--ninja` command-line
+option or in a `NINJA` environment variable.
+
+``` bash
+$ node-gyp configure --gypjs --ninja=/my/path/to/ninja
+$ npm_config_gypjs=1 NINJA=/my/path/to/ninja node-gyp build
+```
+
 
 The "binding.gyp" file
 ----------------------
@@ -181,6 +212,8 @@ Command Options
 | `--python=$path`                  | Set path to the python (2) binary
 | `--msvs_version=$version`         | Set Visual Studio version (win)
 | `--solution=$solution`            | Set Visual Studio Solution version (win)
+| `--gypjs`                         | Use gyp.js instead of gyp
+| `--ninja=$ninja`                  | Override ninja command (with --gypjs)
 
 
 License
--- a/addon.gypi
+++ b/addon.gypi
@@ -92,7 +92,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/$(ConfigurationName)/<(node_lib_file)"'
+          '-l"<(node_root_dir)/<(target_arch)/<(node_lib_file)"'
         ],
         'msvs_disabled_warnings': [
           # warning C4251: 'node::ObjectWrap::handle_' : class 'v8::Persistent<T>'
--- a/addon.gypi
+++ b/addon.gypi
@@ -92,7 +92,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/<(target_arch)/<(node_lib_file)"'
+          '-l"<(node_lib_file)"'
         ],
         'msvs_disabled_warnings': [
           # warning C4251: 'node::ObjectWrap::handle_' : class 'v8::Persistent<T>'
--- a/lib/build.js
+++ b/lib/build.js
@@ -11,7 +11,6 @@ var fs = require('graceful-fs')
   , glob = require('glob')
   , log = require('npmlog')
   , which = require('which')
-  , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
@@ -36,7 +35,6 @@ function build (gyp, argv, callback) {
     , config
     , arch
     , nodeDir
-    , copyDevLib
 
   loadConfigGypi()
 
@@ -60,7 +58,6 @@ function build (gyp, argv, callback) {
       buildType = config.target_defaults.default_configuration
       arch = config.variables.target_arch
       nodeDir = config.variables.nodedir
-      copyDevLib = config.variables.copy_dev_lib == 'true'
 
       if ('debug' in gyp.opts) {
         buildType = gyp.opts.debug ? 'Debug' : 'Release'
@@ -115,7 +112,7 @@ function build (gyp, argv, callback) {
         return
       }
       log.verbose('`which` succeeded for `' + command + '`', execPath)
-      copyNodeLib()
+      doBuild()
     })
   }
 
@@ -173,36 +170,12 @@ function build (gyp, argv, callback) {
             return
           }
           command = msbuildPath
-          copyNodeLib()
+          doBuild()
         })
       })()
     })
   }
 
-  /**
-   * Copies the node.lib file for the current target architecture into the
-   * current proper dev dir location.
-   */
-
-  function copyNodeLib () {
-    if (!win || !copyDevLib) return doBuild()
-
-    var buildDir = path.resolve(nodeDir, buildType)
-      , archNodeLibPath = path.resolve(nodeDir, arch, release.name + '.lib')
-      , buildNodeLibPath = path.resolve(buildDir, release.name + '.lib')
-
-    mkdirp(buildDir, function (err, isNew) {
-      if (err) return callback(err)
-      log.verbose('"' + buildType + '" dir needed to be created?', isNew)
-      var rs = fs.createReadStream(archNodeLibPath)
-        , ws = fs.createWriteStream(buildNodeLibPath)
-      log.verbose('copying "' + release.name + '.lib" for ' + arch, buildNodeLibPath)
-      rs.pipe(ws)
-      rs.on('error', callback)
-      ws.on('error', callback)
-      rs.on('end', doBuild)
-    })
-  }
 
   /**
    * Actually spawn the process and compile the module.
--- a/lib/build.js
+++ b/lib/build.js
@@ -15,7 +15,8 @@ var fs = require('graceful-fs')
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
 
-exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
+exports.usage = 'Invokes `' + (process.env.npm_config_gypjs ? 'ninja' : 
+  (win ? 'msbuild' : 'make')) + '` and builds the module'
 
 function build (gyp, argv, callback) {
   var platformMake = 'make'
@@ -36,6 +37,9 @@ function build (gyp, argv, callback) {
     , arch
     , nodeDir
 
+  if (gyp.opts.gypjs) {
+    command = gyp.opts.ninja || process.env.NINJA || 'ninja'
+  }
   loadConfigGypi()
 
   /**
@@ -70,7 +74,7 @@ function build (gyp, argv, callback) {
       log.verbose('architecture', arch)
       log.verbose('node dev dir', nodeDir)
 
-      if (win) {
+      if (win && !gyp.opts.gypjs) {
         findSolutionFile()
       } else {
         doWhich()
@@ -102,7 +106,7 @@ function build (gyp, argv, callback) {
     // First make sure we have the build command in the PATH
     which(command, function (err, execPath) {
       if (err) {
-        if (win && /not found/.test(err.message)) {
+        if (win && !gyp.opts.gypjs && /not found/.test(err.message)) {
           // On windows and no 'msbuild' found. Let's guess where it is
           findMsbuild()
         } else {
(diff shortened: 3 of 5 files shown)


rvagg:
Can you explain why the need for all the Windows changes in here? Without a solid test suite and the lack of Windows expertise around here this PR is likely going to be very difficult to land given how heavy-handed it is with those changes.

rvagg:
Ah, I see https://github.com/nodejs/node-gyp/pull/951, I'm guessing that what it's all about?

you should open one PR per concern to increase the likelihood of getting changes in, if gyp.js doesn't have any reason to depend on the node.lib location changes then they don't belong together in a PR.

pmed:
On Windows `node.lib` library is [listed in a `libraries` section of `addon.gypi`](https://github.com/nodejs/node-gyp/blob/master/addon.gypi#L95) in a strange manner. The library path contains a MSBuild-specific macro `$(ConfigurationName)`. That path is placed into a generated `.ninja` build file and the `$(ConfigurationName)` macro is obviously staying unexpanded during the building with `ninja`.

I wonder, is anybody use `node-gyp` on Windows with other than Visual C++ toolset?

As I understand `copyNodeLib()` function is only copies an existing `<(node_root_dir)/<(target_arch)/node.lib` to `<(node_root_dir)/$(ConfigurationName)/node.lib` which is unnecessary from my point of view. 

I also added a `findMSVS()` function to locate Visual C++ directory and to its paths with `vcvarsall.bat` invocation on a build stage, so `node-gyp build` would work on Windows without additional actions. Probably, `vcvarsall.bat` should be called explicitly by users, because they may have another C++ toolsets.

pmed:
@rvagg Extracted the fix with `target_arch` usage into a separate PR #964

README.md
```diff
@@ -106,6 +106,37 @@ you can require the `.node` file with Node and run your tests!
...
+tool and a C/C++ compiler toolchain.
+
+To generate projects files with `gyp.js` instead of `gyp` and to build them with
+`ninja`, please supply a command-line option `--gypjs` in `node-gyp` command.
+
+``` bash
+$ node-gyp configure --gypjs
+```
+
+It is also possible to set `npm_config_gypjs` environment variable to turn on
```
indutny: I would suggest all-capitals env var name: `NPM_CONFIG_GYPJS`, or do we use lower-case now?

gyp/gyp.bat
```diff
@@ -1,5 +1,5 @@
-@rem Copyright (c) 2009 Google Inc. All rights reserved.
-@rem Use of this source code is governed by a BSD-style license that can be
-@rem found in the LICENSE file.
-
-@python "%~dp0gyp_main.py" %*
+@rem Copyright (c) 2009 Google Inc. All rights reserved.
+@rem Use of this source code is governed by a BSD-style license that can be
+@rem found in the LICENSE file.
+
+@python "%~dp0gyp_main.py" %*
```
indutny: Em... what happens in this file? I don't see any changes.

gyp/samples/samples.bat
```diff
@@ -1,5 +1,5 @@
-@rem Copyright (c) 2009 Google Inc. All rights reserved.
-@rem Use of this source code is governed by a BSD-style license that can be
-@rem found in the LICENSE file.
-
-@python %~dp0/samples %*
+@rem Copyright (c) 2009 Google Inc. All rights reserved.
+@rem Use of this source code is governed by a BSD-style license that can be
+@rem found in the LICENSE file.
+
+@python %~dp0/samples %*
```
indutny: Ditto...

lib/build.js
```diff
@@ -11,12 +11,11 @@ var fs = require('graceful-fs')
   , glob = require('glob')
   , log = require('npmlog')
   , which = require('which')
-  , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
 
-exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
+exports.usage = 'Invokes `' + (process.env.npm_config_gypjs? 'ninja' : (win ? 'msbuild' : 'make')) + '` and builds the module'
```
indutny: Very likely needs space between `npm_config_gypjs` and `?`.

lib/build.js
```diff
@@ -212,54 +216,74 @@ function build (gyp, argv, callback) {
...
-
-    if (win) {
-      // did the user specify their own .sln file?
-      var hasSln = argv.some(function (arg) {
-        return path.extname(arg) == '.sln'
-      })
-      if (!hasSln) {
-        argv.unshift(gyp.opts.solution || guessedSolution)
+      // invoke vcvarsall.bat before build
+      if (win && vcDir) {
```
indutny: I this necessary with the recent changes to `gyp.js`?

lib/configure.js
```diff
@@ -188,35 +195,37 @@ function configure (gyp, argv, callback) {
...
-        argv.push('-f', 'msvs')
-      } else {
-        log.verbose('gyp', 'gyp format was not specified; forcing "make"')
-        // force the 'make' target for non-Windows
-        argv.push('-f', 'make')
+    if (!gyp.opts.gypjs) {
+      if (!~argv.indexOf('-f') && !~argv.indexOf('--format')) {
+        if (win) {
+          log.verbose('gyp', 'gyp format was not specified; forcing "msvs"')
+          // force the 'make' target for non-Windows
```
indutny: Typo in original comment too, but anyway. Should be `msvs` not `make` here.

gyp/gyp.bat
pmed: Accidental reset of executable bit. On Windows it's not meaningful, on Posix systems this file shouldn't be executed, so I left it in this state.

gyp/samples/samples.bat
pmed: The same as above

lib/build.js
pmed: Do you mean environment support in ninja? I will check it.

indutny: Yeah, I have recently committed it: https://github.com/indutny/gyp.js/commit/ce93190063b323abf47f390845d57177424e3cf0 (NOTE: that this is just a starting commit, I have fixed many things since then)

README.md
pmed: Environment variables named with prefix `npm_config_` are mapped to `gyp.opts` object. Thus, `npm_config_gypjs` variable will set `gyp.opts.gypjs` that used later. As I know an environment variable names are case-sensitive.

indutny: Ah, ok. My bad then!

indutny: Though, they are case sensitive.

indutny:
Is there anything left before we can move forward on this? cc @rvagg

addon.gypi
```diff
@@ -92,7 +92,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/$(ConfigurationName)/<(node_lib_file)"'
+          '-l"<(node_lib_file)"'
```
indutny: Why?

gyp/gyp.bat
indutny: I'd rather not touch it, will result in less noise here.

lib/build.js
```diff
@@ -11,12 +11,12 @@ var fs = require('graceful-fs')
   , glob = require('glob')
   , log = require('npmlog')
   , which = require('which')
-  , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
 
-exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
+exports.usage = 'Invokes `' + (process.env.npm_config_gypjs ? 'ninja' :
```
indutny: At this point it may be useful to just have several ifs here, IMO.

lib/build.js
```diff
@@ -60,7 +62,6 @@ function build (gyp, argv, callback) {
       buildType = config.target_defaults.default_configuration
       arch = config.variables.target_arch
       nodeDir = config.variables.nodedir
-      copyDevLib = config.variables.copy_dev_lib == 'true'
```
indutny: Why?

addon.gypi
pmed: $(ConfigurationName) is a MSBuild macro unknown for Ninja. So I changed it to `target_arch` variable. But this won't work with `--nodedir` option on Windows. As a result `node_lib_file` contains full path to the node.lib now. 

See comments in PR #964. Actually that request should be merged before this one.

lib/build.js
pmed: There is no need to cope node.lib into node_dir/[Debug | Release] directory. See PR #964

gyp/gyp.bat
pmed: @indutny extracted to PR #969

indutny:
Update from me: `ninja` is no longer required to be present on the system, as `gyp.js` bundles a minimal implementation of it.

indutny:
Ahem, ping?

rvagg:
Let's do some `@next` releases for a v4 eh? I'll make a `next` branch and do it off there and we can keep them isolated from `@latest` so they are just for testing.

@pmed can you rebase this off master and squash it down to just the essential commits (or just one commit if that makes sense), it doesn't apply neatly to the current master.

rvagg:
right, just leave that one out of the squash and we'll deal with it separately

pmed:
@rvagg I've combined all changes for this PR in a single commit, and I've made a forced push into my `feature-gypjs` branch. Probably you will need to pull this branch again due this.

Please note that #964 (on which this PR is based) contains 2 commits as before.

springmeyer:
@pmed I'm interested in your work and tried using your `node-gyp` fork. But I am stuck on an error that seems to indicate that the configure step is not working right. Let me know if there is a preferred place to provide feedback/ask questions. Here is what I tried:

 - Added `"gyp.js": "^1.4.9"` to `dependencies` in package.json
 - Added `"node-gyp": "https://github.com/pmed/node-gyp/tarball/feature-gypjs",` to `devDependencies` in package.json

And the error I'm hitting is:

```
ninja: fatal:  chdir to 'build/Release' - No such file or directory
```

Full output is:

```
$ ./node_modules/node-gyp/bin/node-gyp.js rebuild --gypjs
gyp info using node-gyp@3.3.1
gyp info using node@6.3.0 | darwin | x64
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp info spawn ninja
ninja: Entering directory `build/Release'
gypninja: fatal:  chdir to 'build/Release' - No such file or directory
info spawn args [ '-C', 'build/Release', '-j3' ]
gyp ERR! build error 
gyp ERR! stack Error: `ninja` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onExit (/Users/dane/projects/node-osrm/node_modules/node-gyp/lib/build.js:268:23)
gyp ERR! stack     at emitTwo (events.js:106:13)
gyp ERR! stack     at ChildProcess.emit (events.js:191:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:204:12)
gyp ERR! System Darwin 15.5.0
gyp ERR! command "/Users/dane/.nvm/versions/node/v6.3.0/bin/node" "/Users/dane/projects/node-osrm/node_modules/node-gyp/bin/node-gyp.js" "rebuild" "--gypjs"
gyp ERR! cwd /Users/dane/projects/node-osrm
gyp ERR! node -v v6.3.0
gyp ERR! node-gyp -v v3.3.1
gyp ERR! not ok 
```

pmed:
Hi @springmeyer! Thank you for the report. I haven't tested this branch with recent `gyp.js` versions. Just checked it now with `gyp.js` 1.4.9,  it works on Windows, let me check it on OS X later.

pmed:
@spinder try to install older `gyp.j`s version < 1.4.2, it works at least on Windows. I've opened an issue in gyp.js project: https://github.com/indutny/gyp.js/issues/24

indutny:
Made a [fix][0], things should be better now.

[0]: https://github.com/indutny/gyp.js/pull/25

springmeyer:
@indutny thank you - updating to your latest gyp.js work and things are working for me now. It would be great to see this PR landed.

indutny:
Released fixed gyp.js version 1.4.11 . Everything should work normally now.

indutny:
ping @rvagg

README.md
```diff
@@ -106,6 +106,37 @@ you can require the `.node` file with Node and run your tests!
 __Note:__ To create a _Debug_ build of the bindings file, pass the `--debug` (or
 `-d`) switch when running either the `configure`, `build` or `rebuild` command.
 
+#### Usage without Python
+
+There is a [`gyp.js`](https://github.com/indutny/gyp.js/) project, a GYP
+implementation in JavaScript. It generates [`Ninja`](https://ninja-build.org/)
+build files and requires no Python installation.
+
+In this case you will need to install [`Ninja`](https://ninja-build.org/) build
```
indutny: @pmed this is no longer the case. I would say "For fast and incremental builds `ninja` build too should be installed, otherwise a default JavaScript shim will be used to build your addon".

indutny: Though, `node-gyp` will probably need to be modified to use it.

indutny: Or alternatively `npm install -g ninja.js` may be recommended, if binary is not available.

indutny: This shouldn't be a blocker for this PR, though.

pmed: There should be a fallback to `ninja.js` in `doWhich()` checks, when ninja was not found.

piranna: What about using ninja.js directly instead?

pmed: Current status of `ninja.js` is "work-in-progress", and I've never tried it. Maybe later, in next PR after @indutny will release `ninja.js`

indutny: Let's do it in next PR indeed. `ninja.js` is kind of working, but we don't want to delay this thing.

pmed: BTW I don't know how to spawn `ninja.js` as a command-line utility in cross-platfrom way.

indutny: Don't worry to much about it. I'll send a PR once this will land.

indutny:
@rvagg ping

mko-io:
any update?

indutny:
Hello everyone!

I'd like to ping @nodejs/addon-api and @nodejs/node-gyp . Let's get this one resolved.

pmed:
@indutny, as I understand, PR #964 should be accepted at first.

mko-io:
any update? This is a great idea, why progress so slow

jbergstroem:
@bnoordhuis anything I/someone can do to see this land? Asking you since you're generally the goto `node-gyp` bro :)

bnoordhuis:
In all honesty, I like the concept of gyp.js but it's not going to be a drop-in replacement anytime soon and I don't really see the point of merging it into node-gyp.

indutny:
@bnoordhuis the point of merging it under the flag is to bring it to the point where it is going to be a drop-in replacement. Right now, as far we are aware gyp.js works pretty good on all of the addons that we tested. Having this flag will allow us to fix remaining bugs (if there are any), and move forward.

That being said, I don't see any downside of having potentially good functionality under a flag.

jbergstroem:
Perhaps we can document it as "not supported"/"don't depend on it", meaning we can drop it at any release?

indutny:
I agree, this totally makes sense.

@bnoordhuis care to take another look at this? cc @rvagg too

bnoordhuis:
Can you sell me on it?  The reason it's not a drop-in replacement is that gyp allows arbitrary python expressions (and people use that) while gyp.js does not and probably never will.  Also, ninja instead of make.

indutny:
@bnoordhuis there is naive `ninja.js` port available for `gyp.js`, so there is no need to have actual `ninja` binary on the system.

What particular python expressions are you referring to? It does support logical expressions, comparisons, and quite a lot other things that are supported too: https://github.com/indutny/gyp.js/blob/master/test/py-compile-condition-test.js . This could be extended very easily to make it support most of the common use cases.

The selling points:

1. No `python` will be generally needed to build addons (and this applies to Windows too)
2. It is just much faster than GYP
3. GYP is going away, and will not be maintained eventually

jbergstroem:
In terms of logic; I feel we should not aim for "full" compatibility (whatever that means). I don't think we need to treat `gyp.js` as a drop-in replacement, rather a long term replacement that might require minor changes in  complicated gyp-files.

indutny:
Even with this target in mind, we still can make some of the edge-cases work with gyp.js . It should be pretty easy

bnoordhuis:
> What particular python expressions are you referring to?

`10 <= int(var.strip().split(".")) <= 12`, `var.strip().lower()[:4] in ("arm","mips")`, `var.replace("-", "_")`, stuff like that.  Diverse enough that you probably end up implementing a sizable subset of the language.

> there is naive ninja.js port available for gyp.js, so there is no need to have actual ninja binary on the system.

Okay, that speaks in gyp.js's favor. :-)

indutny:
@bnoordhuis Mmm.. supporting that expression doesn't seem to be really painful, we kind of support half of it already.

indutny:
@bnoordhuis though, I'm sure that it may be simplified to something that is already supported, or just expanded in terms of supported expressions.

bnoordhuis:
In any case, this PR will need to be rebased.

pmed:
Yet another selling point for `gyp.js` is the `node-gyp` supports only Python 2.x version. So people with already Python 3.x installed would have to install Python 2.x and here they may face with difficulties of co-existing two Python versions, since they are not fully compatible.

indutny:
@pmed I don't mind if you rebase this, #1092 was opened just in case if you wouldn't want to follow up.

pmed:
@indutny You've already rebased this PR, and #1092 has some comments, so let's run it there.

addon.gypi
refack: For the record  #964 landed

lib/build.js
refack: Now node-gyp is for node >=4 can declare a const and use ` templates

README.md
refack: no need for this after #1185

refack:
@pmed are you interested with moving this forward? I'd be happy to help.

refack:
Superseded by  #1092

PR #964, opened 2016-06-20
build: use target_arch config variable to link against node.lib

pmed:
PR extracted form #962

 on Windows

Using `target_arch` in addon.gypi to link against Node.js library.
This variable was written into build/config.gypi on the `configure` stage.

Do not copy node.lib into node_root_dir/Release or node_root_dir/Debug
on Windows, link it from node_root_dir/target_arch directory.

Removed unused copyNodeLib() function
Removed unused `copy_dev_lib` variable introduced in commit
@84d24189735e19350a93aaf9f6a327bb4c52349e

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -92,7 +92,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/$(ConfigurationName)/<(node_lib_file)"'
+          '-l"<(node_root_dir)/<(target_arch)/<(node_lib_file)"'
         ],
         'msvs_disabled_warnings': [
           # warning C4251: 'node::ObjectWrap::handle_' : class 'v8::Persistent<T>'
--- a/addon.gypi
+++ b/addon.gypi
@@ -92,7 +92,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/<(target_arch)/<(node_lib_file)"'
+          '-l"<(node_lib_file)"'
         ],
         'msvs_disabled_warnings': [
           # warning C4251: 'node::ObjectWrap::handle_' : class 'v8::Persistent<T>'
--- a/lib/build.js
+++ b/lib/build.js
@@ -11,7 +11,6 @@ var fs = require('graceful-fs')
   , glob = require('glob')
   , log = require('npmlog')
   , which = require('which')
-  , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
@@ -36,7 +35,6 @@ function build (gyp, argv, callback) {
     , config
     , arch
     , nodeDir
-    , copyDevLib
 
   loadConfigGypi()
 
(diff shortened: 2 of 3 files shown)


addon.gypi
```diff
@@ -92,7 +92,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/$(ConfigurationName)/<(node_lib_file)"'
+          '-l"<(node_root_dir)/<(target_arch)/<(node_lib_file)"'
```
richardlau: This change will break users on Windows who compile Node from source and use --nodedir to point at their source tree. When building from source, the node.lib file is written to either the `Release` or `Debug` directory depending on the build configuration used.

pmed:
@richardlau, thanks for pointing out the `--nodedir` option.

I've changed the `node_lib_file` variable to use full path to the node lib there. In case when the `--nodedir` option was set, this full path to the node lib includes `$(Configuration)` part, otherwise it includes `target_arch` variable.

rvagg:
I need to defer to @nodejs/platform-windows folks on this, is it safe to compile against a node.lib that's not in the build directory? Is it statically linked so that if it gets deleted in the original location then it won't cause problems?

@pmed does this work on Node.js v0.10 and v0.12? Lots of stuff changed on the v4 bump including file locations.

pmed:
I've checked these changes with Node versions `0.10.46`, `0.12.15` and they work, among with `--nodedir` option and nodejs sources at tags `0.10.45`,  `0.12.14` on combinations of Debug|Release, x64|ia32 switches.

rvagg:
@pmed don't let us forget about this, we're going to start work on v4 and this would be a good inclusion for testing

bnoordhuis:
Ping @nodejs/platform-windows, see @rvagg's question here: https://github.com/nodejs/node-gyp/pull/964#issuecomment-228724642

> Is it safe to compile against a node.lib that's not in the build directory? Is it statically linked so that if it gets deleted in the original location then it won't cause problems?

pmed:
It seems there could be a case, when a user would remove nodeir after `node-gyp configure --nodedir=some/location` and before `node-gyp build`.

But I don't think this is a issue, because in this case `node-gyp configure` could be invoked again with new `--nodedir` option.

indutny:
ping @nodejs/platform-windows we need your help to get this landed.

seishun:
`node.lib` is statically linked, it's not a dll. For the other questions, I need some steps to test this because it's been a while since I last used node-gyp.

gibfahn:
For the commit messages, maybe something like:

```
build, win: use target_arch to link with node.lib

Using `target_arch` in addon.gypi to link against Node.js library. This
variable was written into build/config.gypi on the `configure` stage.

Do not copy node.lib into node_root_dir/Release or node_root_dir/Debug
on Windows, link it from node_root_dir/target_arch directory.
```

and

```
configure: use full path in node_lib_file GYP var

Set path to node lib in `$(Configuration)` dir when `--nodedir` option
is supplied, otherwise use value of `target_arch` variable.
```

joaocgreis:
@rvagg @bnoordhuis can I land this?

@pmed let me know if you object to me rewording the commit messages to @gibfahn 's suggestion above.

bnoordhuis:
@joaocgreis Yes, if you think it's good to go and CI is green.

joaocgreis:
CI of rebased version:
- v0.10: https://ci.nodejs.org/job/nodegyp-test-commit/117/
- v0.12: https://ci.nodejs.org/job/nodegyp-test-commit/118/
- v4: https://ci.nodejs.org/job/nodegyp-test-commit/114/
- v6: https://ci.nodejs.org/job/nodegyp-test-commit/116/
- v7: https://ci.nodejs.org/job/nodegyp-test-commit/115/

Landed in https://github.com/nodejs/node-gyp/compare/c307b30...386746c

PR #965, opened 2016-06-20
install: de-duplicate code in downloadNodeLib() function

pmed:
Modified the result of processRelease() to return uniform names `libUrl`,
`libPath` for Windows in objects named as target platforms (`ia32` and `x64`)

This allows to loop over the target platforms in downloadNodeLib() and to use
the universal names instead of two almost identical code parts with `libUrl32`,
`libPath32` and `libUrl64`, `libPath64` names.


lib/install.js
```diff
@@ -313,75 +313,44 @@ function install (gyp, argv, callback) {
...
-            return cb(e)
-          }
-
-          req.on('error', done)
-          req.on('response', function (res) {
-            if (res.statusCode !== 200) {
-              done(new Error(res.statusCode + ' status code downloading 32-bit ' + release.name + '.lib'))
-              return
+        var archs = ['ia32', 'x64']
+        var async = archs.length
```
rvagg: best avoid the `async` keyword for future-proofing

rvagg:
I don't mind this refactor although I consider the changes in process-release to be breaking because that module is usable outside of node-gyp (I've used it myself that way). The change is good, but I'd rather hold this off till v4, which we're close to because of the `NVM_*` env var removals we need to do.

lib/install.js
pmed: There are many variables named `async` in the code above,  I just used the same name.

pmed:
As I see `process-release` is not exported from the node-gyp in any way. Thus these changes in the `process-release` are only details of implementation. And I think no other external projects can rely on internals of `node-gyp`.

rvagg:
`require('node-gyp/lib/process-release')`

but anyway, we can break stuff in v4 so whatever

rvagg:
sorry this got stale @pmed, still want to pursue it? if so, rebase to current master and let's take a look

pmed:
Hi @rvagg 

I've rebased it on the master.

rvagg:
thanks @pmed, landed in 03683f09d
it's pretty surprising the duplication lasted this long isn't it? good work.

PR #969, opened 2016-06-26
Unset executable bit for .bat files

pmed:
On Windows this bit is not meaningful, on Posix systems such files
should not be executable.


rvagg:
@nodejs/platform-windows

RReverser:
Makes sense, LGTM to me.

bnoordhuis:
Landed in afc766a, thanks.

PR #970, opened 2016-06-29
Add windows-build-tools to Readme

felixrieseberg:
Hi from Microsoft's open source team! We're trying to make things better - and one of the things we wanted to make easier is the installation of requirements to run node-gyp. Instead of installing stuff manually, Windows users can now simply run `npm i -g windows-build-tools`, which will install the C++ build tools and Python (conflict-free) and then configure npm appropriately.

This PR simply adds a reference to those tools to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ You will also need to install:
       * You also need to install the `Command Line Tools` via Xcode. You can find this under the menu `Xcode -> Preferences -> Downloads`
       * This step will install `gcc` and the related toolchain containing `make`
   * On Windows:
-    * Option 1: Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install -g windows-build-tools` from an elevated PowerShell (run as Administrator).
+    * Option 1: Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
     * Option 2: Install tools and configuration manually:
       * Visual C++ Build Environment:
         * Option 1: Install [Visual C++ Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools) using the **Default Install** option.
--- a/README.md
+++ b/README.md
@@ -42,15 +42,17 @@ You will also need to install:
       * You also need to install the `Command Line Tools` via Xcode. You can find this under the menu `Xcode -> Preferences -> Downloads`
       * This step will install `gcc` and the related toolchain containing `make`
   * On Windows:
-    * Visual C++ Build Environment:
-      * Option 1: Install [Visual C++ Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools) using the **Default Install** option.
+    * Option 1: Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install -g windows-build-tools` from an elevated PowerShell (run as Administrator).
+    * Option 2: Install tools and configuration manually:
+      * Visual C++ Build Environment:
+        * Option 1: Install [Visual C++ Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools) using the **Default Install** option.
 
-      * Option 2: Install [Visual Studio 2015](https://www.visualstudio.com/products/visual-studio-community-vs) (or modify an existing installation) and select *Common Tools for Visual C++* during setup. This also works with the free Community and Express for Desktop editions.
+        * Option 2: Install [Visual Studio 2015](https://www.visualstudio.com/products/visual-studio-community-vs) (or modify an existing installation) and select *Common Tools for Visual C++* during setup. This also works with the free Community and Express for Desktop editions.
 
-      > :bulb: [Windows Vista / 7 only] requires [.NET Framework 4.5.1](http://www.microsoft.com/en-us/download/details.aspx?id=40773)
+        > :bulb: [Windows Vista / 7 only] requires [.NET Framework 4.5.1](http://www.microsoft.com/en-us/download/details.aspx?id=40773)
 
-    * Install [Python 2.7](https://www.python.org/downloads/) (`v3.x.x` is not supported), and run `npm config set python python2.7` (or see below for further instructions on specifying the proper Python version and path.)
-    * Launch cmd, `npm config set msvs_version 2015`
+      * Install [Python 2.7](https://www.python.org/downloads/) (`v3.x.x` is not supported), and run `npm config set python python2.7` (or see below for further instructions on specifying the proper Python version and path.)
+      * Launch cmd, `npm config set msvs_version 2015`
 
     If the above steps didn't work for you, please visit [Microsoft's Node.js Guidelines for Windows](https://github.com/Microsoft/nodejs-guidelines/blob/master/windows-environment.md#compiling-native-addon-modules) for additional tips.
 


README.md
```diff
@@ -42,15 +42,17 @@ You will also need to install:
       * You also need to install the `Command Line Tools` via Xcode. You can find this under the menu `Xcode -> Preferences -> Downloads`
       * This step will install `gcc` and the related toolchain containing `make`
   * On Windows:
-    * Visual C++ Build Environment:
-      * Option 1: Install [Visual C++ Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools) using the **Default Install** option.
+    * Option 1: Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install -g windows-build-tools` from an elevated PowerShell (run as Administrator).
```
rvagg: What happens if you don't use PowerShell when doing this? What happens when you don't use elevated privileges?

README.md
```diff
@@ -42,15 +42,17 @@ You will also need to install:
...
+    * Option 2: Install tools and configuration manually:
+      * Visual C++ Build Environment:
+        * Option 1: Install [Visual C++ Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools) using the **Default Install** option.
 
-      * Option 2: Install [Visual Studio 2015](https://www.visualstudio.com/products/visual-studio-community-vs) (or modify an existing installation) and select *Common Tools for Visual C++* during setup. This also works with the free Community and Express for Desktop editions.
+        * Option 2: Install [Visual Studio 2015](https://www.visualstudio.com/products/visual-studio-community-vs) (or modify an existing installation) and select *Common Tools for Visual C++* during setup. This also works with the free Community and Express for Desktop editions.
```
rvagg: these extra indentations look wrong

rvagg:
Amazing, this'll be fantastic if it's as simple as it looks. @felixrieseberg can you point to where you're modifying npm config to make this work? I'd like to understand how you're doing this that makes it conflict-free and yet makes npm/node-gyp safely aware of it.

@nodejs/platform-windows could you please take a look at this and give some opinions?

README.md
felixrieseberg: * Not using cmd: We'll spawn a little ps1 (invisible) in the background
* Not admin: You'll get a pretty error message (telling you to try again as admin) and an exit code 1. Installation is not attempted.

README.md
felixrieseberg: You'd think so, but that's required to convince GitHub's markdown parser to understand the hierarchy ([see the output here](https://github.com/felixrieseberg/node-gyp/blob/6efeeb5cb4daed965a9acf3d4bce15d88efc6ec3/README.md)). Obviously happy to change if you want a different hierarchy!

felixrieseberg:
Of course (and thanks for the :tulip:!). The project is [here](https://github.com/felixrieseberg/windows-build-tools) and is really just a little wrapper around the required MSIs, running them headless with the appropriate parameters. IT Admins have been installing things on Windows headless for years, so a coin dropped when somebody asked me why the heck she/he can't install the build tools directly from npm.

In detail, the C++ build tools come with no-conflict (they simply won't install if the appropriate packages have already been installed by either the tools themselves or other applications). Python's 2.7 MSI does not modify `PATH`, so we can simply install it into a custom dotfolder inside the user's home folder without hurting anyone (wasting files, maybe, but certainly not breaking things). We then tell npm [where we put Python](https://github.com/felixrieseberg/windows-build-tools/blob/master/ps1/set-environment.ps1).

If any of the smart people in @nodejs/platform-windows have feedback/input/things to yell at me, I'd :heart: to hear it.

README.md
rvagg: ah, ok, nmind then!

README.md
joaocgreis: Can you recommend Command Prompt instead? I suspect it will be simpler for many users. Or both: "Command Prompt or PowerShell".

rvagg:
Ah nice, so the C++ tools go into a standard location, that's the bit I was imagining being hard to do a custom location, Python being in a custom location makes sense although I wonder if it'll make a difference not having `python` in the `PATH`. I'd love to hear feedback on having this tried out on a suite of popular addons to confirm that it works well. If it does then this will be absolutely huge and you'll be a hero @felixrieseberg! This is one of the biggest pain points experienced by Node users.

joaocgreis:
@felixrieseberg thanks for submitting this! I tried this and it did not go too well, opened an issue in your repo: https://github.com/felixrieseberg/windows-build-tools/issues/5 . But I'm sure we can make it work!

joaocgreis:
cc @mousetraps

README.md
RReverser: In fact, PowerShell is rather easier to use for most users due to various unix-like aliases, nicer output and various UX issues (like multiline copy) solved. As a developer working on Windows, I'd say this is what most developers on Windows use nowadays.

xzyfer: > PowerShell is rather easier ... I'd say this is what most developers on Windows use nowadays

Anecdotally in my time running a relative popular native extension, node-sass, I don't believe this to be the case. At least not among developers using node primarily as their asset pipeline.

eljefedelrodeodeljefe:
Left two tickets also. Most importantly: can you state what `BuildTools_Full.exe` does or contains? Are the contents under some kind of license? Ref: https://github.com/felixrieseberg/windows-build-tools/issues/7

Also I would be very happy if compatibility and user friendliness are No. 1 priority for this. Since we don't have quite the history doing so on NT, I fear the [promise issue](https://github.com/felixrieseberg/windows-build-tools/issues/7), v4 compat and introducing PowerShell aren't helping.

I personally would love to see more OSS behind what BuildTools_Full.exe ships. 

Otherwise I love where that heads. Thx.

felixrieseberg:
@eljefedelrodeodeljefe: Super fair point - the build tools package comes straight from the Visual Studio C++ Compiler team. You can read more about [it here](https://blogs.msdn.microsoft.com/vcblog/2016/03/31/announcing-the-official-release-of-the-visual-c-build-tools-2015/). They are free. Included, in detail: 

Visual C++ Compilers (targeting x86, X64 and ARM)
Visual C++ headers & libraries (CRT & STL)
Visual C++ build scripts (targeting Windows desktop)
Microsoft Build Tools 2015 (MSBuild)
Windows SDK 8.1 (optional, on by default)
Windows SDK 10 (optional, off by default)
ATL and MFC (optional, off by default)
C++ Build tools specific command prompts

(I hear your point though, this should be in the readme - and will be added).

eljefedelrodeodeljefe:
@felixrieseberg awesome and thanks for addressing this. And thanks for the for link. I assume W10 SDK is included then in what we download, right (it's marked as optional above)? Then it's everything I hoped and everything we'd need for NT build environment.

Shout outs to the compiler team. I dig into their doc a little then.

felixrieseberg:
In 99% of cases, you probably want the 8.1 Kit, so that's what's in there right now. To be clear, 8.1 binaries run well on Windows 10, you just need 10 for 10-specific features - and I'm not aware of any npm modules that do. The other way around, the 10 SDK does require 10, which is why a slightly older SDK probably hits more people. Again, thanks for helping out and being interested.

eljefedelrodeodeljefe:
I was doing some related work, basically guessing on include paths. There was at least one W10, but tbh, I can be wrong or it's currently obsolete. Just asking b/c I had to specifically install it on my VS2015 setup afterwards. Please also note, I write on OS X but am enthusiastic for the NT stuff.

I was attempting building a toolchain without python [here](https://github.com/eljefedelrodeodeljefe/platform-tools/blob/master/lib/flags.js#L27-L65), but in any case lack of compiler or unknown locations of it would have been a major obstacle. I am integrating your work and probably can give more feedback afterwards. Really excited now.

felixrieseberg:
Thanks to those of you trying it out - thanks to your help, we managed to figure out that smaller machines would run into memory/cpu limits, mostly because I went a bit overboard with tailing the logfiles. The whole thing is now *a lot* less resource hungy and installs fine on a little Windows machine with less than 1 GB in total memory.

rvagg:
@joaocgreis @orangemocha can you have a look at the latest incarnation of this and see if it's something we might be prepared to promote?

joaocgreis:
@felixrieseberg Have you had a chance to look at https://github.com/felixrieseberg/windows-build-tools/issues/8#issuecomment-230424888 ?

@rvagg We certainly want to promote this, but it is currently downloading the wrong package and has to be fixed upstream first.

felixrieseberg:
@joaocgreis That was a fixed a few days ago, I just forgot to update folks - sorry for forgetting to comment on that one.

joaocgreis:
Tested, LGTM.

Will land tomorrow if there are no objections.

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/commit/f6eab1f9e4b160b60b2736bb01a09c83785ba979

PR #971, opened 2016-06-29
gyp: fix runtime "NameError"s

thefourtheye:
In "gyp/gyptest.py", `shlex` is used, but it is not imported. So it
will throw `NameError` at runtime.

In "gyp/pylib/gyp/mac_tool.py", the `except` block expects the type of
exception to be the first item or nothing. As it is, the block will
throw a `NameError` at runtime.

### DIFF
--- a/gyp/gyptest.py
+++ b/gyp/gyptest.py
@@ -10,6 +10,7 @@ gyptest.py -- test runner for GYP tests.
 
 import os
 import optparse
+import shlex
 import subprocess
 import sys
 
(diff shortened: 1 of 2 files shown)


thefourtheye:
Oh, does this come from the upstream? I suggested the same there as well. //codereview.chromium.org/2108583005

bnoordhuis:
Correct.  Changes to files under gyp/ should go upstream first.

thefourtheye:
Oh. Closing this issue then.

PR #990, opened 2016-07-12
Remove path-array

mhart:
This reverts commits 81eadc5692eb33f308209412daa7db5d53bcb744 and ff88e5fd1211877a9a6eb2a906f7e4d4ff377604.

Of the 5.5MB footprint of node-gyp, path-array makes up 3.6MB.

Removing this reduces the weight of npm (and ultimately node) appreciably (by 20%)

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -287,9 +287,11 @@ function configure (gyp, argv, callback) {
       argv.unshift(gyp_script)
 
       // make sure python uses files that came with this particular node package
-      var pypath = process.env.PYTHONPATH ? [process.env.PYTHONPATH] : []
-      pypath.unshift(path.resolve(__dirname, '..', 'gyp', 'pylib'))
-      process.env.PYTHONPATH = pypath.join(process.platform === 'win32' ? ';' : ':')
+      var pypath = [path.join(__dirname, '..', 'gyp', 'pylib')]
+      if (process.env.PYTHONPATH) {
+        pypath.push(process.env.PYTHONPATH)
+      }
+      process.env.PYTHONPATH = pypath.join(win ? ';' : ':')
 
       var cp = gyp.spawn(python, argv)
       cp.on('exit', onCpExit)
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -14,7 +14,6 @@ var fs = require('graceful-fs')
   , semver = require('semver')
   , mkdirp = require('mkdirp')
   , cp = require('child_process')
-  , PathArray = require('path-array')
   , extend = require('util')._extend
   , processRelease = require('./process-release')
   , spawn = cp.spawn
@@ -288,8 +287,9 @@ function configure (gyp, argv, callback) {
       argv.unshift(gyp_script)
 
       // make sure python uses files that came with this particular node package
-      var pypath = new PathArray(process.env, 'PYTHONPATH')
-      pypath.unshift(path.join(__dirname, '..', 'gyp', 'pylib'))
+      var pypath = process.env.PYTHONPATH ? [process.env.PYTHONPATH] : []
+      pypath.unshift(path.resolve(__dirname, '..', 'gyp', 'pylib'))
+      process.env.PYTHONPATH = pypath.join(process.platform === 'win32' ? ';' : ':')
 
       var cp = gyp.spawn(python, argv)
       cp.on('exit', onCpExit)
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "nopt": "2 || 3",
     "npmlog": "0 || 1 || 2 || 3",
     "osenv": "0",
-    "path-array": "^1.0.0",
     "request": "2",
     "rimraf": "2",
     "semver": "2.x || 3.x || 4 || 5",
--- a/test/test-configure-python.js
+++ b/test/test-configure-python.js
@@ -1,0 +1,73 @@
+'use strict'
+
+var test = require('tape')
+var path = require('path')
+var gyp = require('../lib/node-gyp')
+var requireInject = require('require-inject')
+var configure = requireInject('../lib/configure', {
+  'graceful-fs': {
+    'writeFile': function (file, data, cb) { cb() },
+    'stat': function (file, cb) { cb(null, {}) }
+  }
+})
+
+var EXPECTED_PYPATH = path.join(__dirname, '..', 'gyp', 'pylib')
+var SEPARATOR = process.platform == 'win32' ? ';' : ':'
+var SPAWN_RESULT = { on: function () { } }
+
+test('configure PYTHONPATH with no existing env', function (t) {
+  t.plan(1)
+
+  delete process.env.PYTHONPATH
+
+  var prog = gyp()
+  prog.parseArgv([])
+  prog.spawn = function () {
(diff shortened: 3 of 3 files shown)


rvagg:
Does this actually work though? I was looking at this a while back because the dependency depth here bothers me too but I wasn't sure that the same functionality could be replicated quite as easily.

Perhaps it's really this easy but since we don't have test cases around this I'm not overly confident! We can certainly give it a try in v4 pre-releases which I'm planning on spinning up soon but tests would be nice. Any chance you'd like to write us some tests for this?

mhart:
Sure – I mean, I believe it was only added by @TooTallNate originally because he thought it was a good fit for the `path-array` module – not because the original code, written by @timbertson was faulty in any way: https://github.com/nodejs/node-gyp/pull/509/files#r20182682

I just reverted those lines to how they were when #509 was merged.

Kinda seems like overkill to add a module for just a single string join, no?

lib/configure.js
```diff
@@ -288,8 +287,9 @@ function configure (gyp, argv, callback) {
       argv.unshift(gyp_script)
 
       // make sure python uses files that came with this particular node package
-      var pypath = new PathArray(process.env, 'PYTHONPATH')
-      pypath.unshift(path.join(__dirname, '..', 'gyp', 'pylib'))
+      var pypath = process.env.PYTHONPATH ? [process.env.PYTHONPATH] : []
+      pypath.unshift(path.resolve(__dirname, '..', 'gyp', 'pylib'))
```
mhart: Ah, just noticed this should stay as `path.join` – will add that.

mhart:
I'll add some tests tomorrow if I can figure out how 👍

lib/configure.js
```diff
@@ -288,8 +287,9 @@ function configure (gyp, argv, callback) {
       argv.unshift(gyp_script)
 
       // make sure python uses files that came with this particular node package
-      var pypath = new PathArray(process.env, 'PYTHONPATH')
+      var pypath = process.env.PYTHONPATH ? [process.env.PYTHONPATH] : []
       pypath.unshift(path.join(__dirname, '..', 'gyp', 'pylib'))
+      process.env.PYTHONPATH = pypath.join(process.platform === 'win32' ? ';' : ':')
```
bnoordhuis: It's arguably a bit nicer to wrap it in a `if ('PYTHONPATH' in process.env) { ... }`.

Can you replace the `process.platform === 'win32' ? ';' : ':'` with `win ? ';' : ':'`?

TooTallNate:
It's only so big because of `es6-symbol`... I wonder if there's a smaller polyfill.

lib/configure.js
mhart: When you say wrap in an `if`, what did you have in mind?

I can certainly do:
```js
var pypath = 'PYTHONPATH' in process.env ? [process.env.PYTHONPATH] : []
```

Or would you prefer something like:

```js
var pypath = [path.join(__dirname, '..', 'gyp', 'pylib')]
if ('PYTHONPATH' in process.env) {
  pypath.push(process.env.PYTHONPATH)
}
process.env.PYTHONPATH = pypath.join(win ? ';' : ':')
```
(I kinda prefer that over an unshift anyway tbh)

bnoordhuis: I mean the second one.  I wrote `'PYTHONPATH' in process.env` but perhaps checking for truthiness / non-emptiness (i.e., just `process.env.PYTHONPATH`) is better.

mhart:
```js
var pypath = [path.join(__dirname, '..', 'gyp', 'pylib')]
if (process.env.PYTHONPATH) pypath.push(process.env.PYTHONPATH)
process.env.PYTHONPATH = pypath.join(win ? ';' : ':')
```
Preferences on indentation, braces?

bnoordhuis: node-gyp uses two spaces and (predominantly, I think) braces.

mhart:
@rvagg added some tests 👍

mhart:
@rvagg let me know if those tests look ok, or if there's anything you need further to get this through

mhart:
Hi all – any movement on this?

FWIW, I've been building a tool to track down size regressions (and other issues with sub-deps) – you can see an example of it coming to life here: http://pkgsize.com.s3.amazonaws.com/node-gyp.html (hovering over should show you what events caused certain size changes)

Obviously, `path-array` and `request` are the big offenders. Considering that `path-array` is literally just used for one string join here and has such a deep dep tree, it really will reduce the risk of even further bloat to remove it.

bnoordhuis:
LGTM.  CI: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/12/

This probably also needs a CI run with v0.10 and v0.12.

mhart:
@bnoordhuis would love to get this in – how do I get a CI run with v0.10 and v0.12?

bnoordhuis:
https://ci.nodejs.org/view/All/job/nodegyp-test-commit/13/ - v0.10.46
https://ci.nodejs.org/view/All/job/nodegyp-test-commit/14/ - v0.12.15

@nodejs/build Can someone take a look at why the win2012r2 bot complains about this:
```
c:\workspace\nodegyp-test-commit\nodes\win2012r2\local_node>mkdir npm-cache 

c:\workspace\nodegyp-test-commit\nodes\win2012r2\local_node>call "c:\workspace\nodegyp-test-commit\nodes\win2012r2\local_node\npm.cmd" config set cache "c:\workspace\nodegyp-test-commit\nodes\win2012r2\local_node\npm-cache" 
The system cannot execute the specified program.
```
From https://ci.nodejs.org/view/All/job/nodegyp-test-commit/13/nodes=win2012r2/console.

bnoordhuis:
The ubuntu buildbot seems pretty happy though so I think this is good to go.  @mhart Can you squash it into what you think are logical/reasonable commits?

mhart:
@bnoordhuis Sure thing – I'll keep the original "revert" commits separate and squash the rest

mhart:
@bnoordhuis done – let me know if you want me to squash further

joaocgreis:
The job was trying to download `node.exe` from the wrong place, so it was getting a html page instead of a exe, hence the "cannot execute" error. Also, npm was interfering with the CITGM job. Both fixed.

Started two jobs with similar parameters:
- v0.10: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/29/
- v0.12: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/30/

Seems good!

mhart:
Anything more need to be done on this? If so lemme know 👍

mhart:
Hey all, just checking in on this PR. Let me know if there are any more concerns?

bnoordhuis:
I think we were (and are) just waiting for the next node-gyp release.  I've landed it in 9c8d275...ddac348, thanks!

mhart:
Great, thanks @bnoordhuis 👍 

Once it makes it to Node.js it'll be the first drop in size we've seen in years – here's hoping!

jbergstroem:
This is pretty awesome. Thanks for the effort!

mhart:
Thanks for pushing this out – npm@4.1.2 has just included it and has reaped the benefits of an 18% reduction in size 👍

PR #999, opened 2016-07-27
Update shared library extension on AIX to be .a instead of .so

sxa555:
AIX generally uses .a instead of .so as the shared library suffix. Without this change, if you try to build a shared library version of Node on AIX (further patches required which I have working) then node-gyp tries to bolt on .so to the end of libnode.a which is not what we want. This changes the suffix if we're on AIX.


bnoordhuis:
Changes to files under gyp/ should be sent to the upstream GYP project first.

I don't quite understand your explanation though.  node-gyp is not involved in building libnode so why is this change necessary?

sxa555:
Fair point - I was assuming this repository was used for the gyp for building node under tools/gyp and not just the one for modules. I'll look at getting it into gyp then back into the main node codebase directly.

PR #1000, opened 2016-07-28
Update request dependency to version ≥ 2.74.0

evilaliv3:
This update is to address a security vulnerability allowing potential remote memory exposure on request<=2.68

Details at: https://snyk.io/vuln/npm:request:20160119

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "npmlog": "0 || 1 || 2 || 3",
     "osenv": "0",
     "path-array": "^1.0.0",
-    "request": "2",
+    "request": "^2.74.0",
     "rimraf": "2",
     "semver": "2.x || 3.x || 4 || 5",
     "tar": "^2.0.0",


Fishrock123:
cc @rvagg & @bnoordhuis

bnoordhuis:
LGTM, I guess, but it's not like we use request in a way that makes it susceptible to that vulnerability.

evilaliv3:
it is not you @bnoordhuis but in case the library is used in other projects (like fsevents) you end in suggesting the vulnerable version in the packages resolution and so possibly injecting that version in who uses you.

bnoordhuis:
You mean because of npm@3 de-duping dependencies?  node-gyp accepts a broad range so I don't think it matters as long as other modules don't explicitly depend on vulnerable versions.

The same is true of fsevents, any request@2.x satisfies its constraints.  Or rather, it satisfies node-pre-gyp's constraints - fsevents itself doesn't depend on request.

evilaliv3:
ah ok this is a better explanation that i think it is solved.

if there is no way to exclude a  version ok i think with this clarification any user would be able to check that the right version is used.

bnoordhuis:
I'll go ahead and close, I think we have consensus that no changes are necessary in node-gyp.

PR #1043, opened 2016-11-01
FIXUP: bcrypt cannot be installed with proxychain.

jinchizhong:
GYP should not test stderr output as failed condition.


rvagg:
I think take this up @ https://github.com/refack/GYP
In general I think I agree but I'm afraid of the many possible edge cases where this breaks down and I don't know why it was put there in the first place--perhaps for good reason.

PR #1061, opened 2016-11-28
Update link of gyp as Google code is shutting down

PeterDaveHello:


gibfahn:
Would https://chromium.googlesource.com/external/gyp or https://gyp.gsrc.io/ be a better starting point for someone who doesn't know what gyp is?

PeterDaveHello:
Oh cool, would you like me to update it?

gibfahn:
@PeterDaveHello Yes, sounds good to me. I guess you should use https://gyp.gsrc.io as that seems to be the homepage.

PeterDaveHello:
@gibfahn thanks, if you don't mind, I just directly updated and committed on GitHub, I think you can squash them directly on GitHub, too, if you don't like it, I can clone and squash them locally or just send a new one, thanks.

gibfahn:
@PeterDaveHello looks fine to me. This needs a `node-gyp` contributor to review though.

bnoordhuis:
Thanks Peter, landed in ff9a6fa.

PR #1069, opened 2016-12-10
Fix deprecated dependency warning

simonepri:
screenshot- 34
**node-gyp** installs **npmlog@3.1.2**
**npmlog@3.1.2** installs **gauge@2.6.0**

**gauge@2.6.0** uses **has-color@0.1.7** that is been ranamed to _supports-color_

The warning is not showed if we tell **node-gyp** to use **npmlog@4.0.1**
**npmlog@4.0.1** installs **gauge@2.7.2** that fix the dependency warnings. 

See https://github.com/iarna/gauge/pull/77

screenshot- 33

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "minimatch": "^3.0.2",
     "mkdirp": "^0.5.0",
     "nopt": "2 || 3",
-    "npmlog": "0 || 1 || 2 || 3",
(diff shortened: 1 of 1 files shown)


Fishrock123:
Thanks, landed in https://github.com/nodejs/node-gyp/commit/80fc5c3d313c54a190e4ed079bfd642dae329af5

PR #1071, opened 2016-12-13
Fixes #1070

### LINKED ISSUE #1070: Read IOJS_ORG_MIRROR and NODEJS_ORG_MIRROR also from npm config
Just like in [node-sass](https://github.com/sass/node-sass/blob/master/lib/extensions.js#L223), would be nice to read both `IOJS_ORG_MIRROR` and `NODEJS_ORG_MIRROR` (and their deprecated version) also from `npm config` settings. With this would be also possible to set its value in .npmrc file.

The change is really easy to do: just insert these variables in their respectives [if/else chains](https://github.com/nodejs/node-gyp/blob/master/lib/process-release.js#L66):
- `process.env.npm_config_iojs_org_mirror`
- `process.env.npm_config_nvm_iojs_org_mirror`
- `process.env.npm_config_nodejs_org_mirror`
- `process.env.npm_config_nvm_nodejs_org_mirror`

I can do a pull request if you will.

Thanks in advance!

fsmaia:
Fixes issue #1070 by adding support to npm config for mirror variables.

### DIFF
--- a/lib/process-release.js
+++ b/lib/process-release.js
@@ -54,19 +54,19 @@ function processRelease (argv, gyp, defaultVersion, defaultRelease) {
   // check for the nvm.sh standard mirror env variables
   if (!overrideDistUrl) {
     if (isIojs) {
-      if (process.env.IOJS_ORG_MIRROR) {
-        overrideDistUrl = process.env.IOJS_ORG_MIRROR
-      } else if (process.env.NVM_IOJS_ORG_MIRROR) {// remove on next semver-major
-        overrideDistUrl = process.env.NVM_IOJS_ORG_MIRROR
+      if (process.env.IOJS_ORG_MIRROR || process.env.npm_config_iojs_org_mirror) {
+        overrideDistUrl = process.env.IOJS_ORG_MIRROR || process.env.npm_config_iojs_org_mirror
+      } else if (process.env.NVM_IOJS_ORG_MIRROR || process.env.npm_config_nvm_iojs_org_mirror) {// remove on next semver-major
+        overrideDistUrl = process.env.NVM_IOJS_ORG_MIRROR || process.env.npm_config_nvm_iojs_org_mirror
         log.warn('download',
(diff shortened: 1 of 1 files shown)


fsmaia:
By adding them as command line flags you mean relying on a dependency like [command-line-args](https://www.npmjs.com/package/command-line-args)? And then reading the config either by environment variable and command-line (which include npm_config automatic inheritance)? I never worked in this config-to-cli inheritance before, but I can try with this library. Any other suggestion?

About the tests and nice commit log, I can work on that (linking to issue is just a personal preference).

About keeping lines <= 80 columns, did you consider writing an `.editorconfig` file to project?

Thanks in advance!

bnoordhuis:
> By adding them as command line flags you mean relying on a dependency like command-line-args?

We use nopt and some homegrown logic.  You can find the list of switches in lib/node-gyp.js, grep for `configDefs` and `parseArgv`.

bnoordhuis:
> About keeping lines <= 80 columns, did you consider writing an .editorconfig file to project?

There are existing style violations and I don't think it's worth the churn to fix them, I just try to keep new ones from sneaking in. :-)

fsmaia:
> We use nopt and some homegrown logic. You can find the list of switches in lib/node-gyp.js, grep for configDefs and parseArgv.

When switching to `parseArgv` instead of environment variable, should I refactor [this entry](https://github.com/nodejs/node-gyp/blob/master/lib/install.js#L441) too? By the way switching to `parseArgv` involves moving it from `node-gyp` and exposing it in a higher instance (to be used in `node-gyp` and `install`). Are you sure this is the moment to do this refactoring?

> There are existing style violations and I don't think it's worth the churn to fix them, I just try to keep new ones from sneaking in. :-)

In order to not reproduce these violations again, could you give me the specifications?

fsmaia:
Reviewing the code a little further I think `parseArgv` isn't a nice approach. It is useful to parse CLI arguments, but in this case we are dealing with environment configuration used in installation step, and at this time the CLI isn't even available.

Issue #903, opened 2016-04-12
setting cache-min to > 2^32 breaks node-gyp


dominictarr:
using node-gyp with some settings for cache-min crashes with an error.

I was setting `cache-min` to a large value, say, `9999999999` so that npm will install from the cache instead of going to the main registry, which slows down an install a lot if you have a slow internet connection, especially since most requests will just return 304 that your cached version is correct.

However, I get this error when installing a module that uses node-gyp

```
ValueError: Variable expansion in this context permits str and int only, found long for cache_min while trying to load binding.gyp
```

this just seemed like a weird error, and compilers do lots of weird things, but then I realized that my cache-min setting was larger than 2^32, although I don't know why node-gyp would care about that,
setting it to 999999999 (9 9's) that works.

how does node-gyp use cache-min? is this a node-gyp bug or an npm bug?

full trace:

```
> chloridedown@1.0.22 install /home/dominic/c/scuttlebot/node_modules/chloridedown
> node-gyp rebuild

Traceback (most recent call last):
  File "/home/dominic/.nvm/versions/node/v5.6.0/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py", line 16, in <module>
    sys.exit(gyp.script_main())
  File "/home/dominic/.nvm/versions/node/v5.6.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 545, in script_main
    return main(sys.argv[1:])
  File "/home/dominic/.nvm/versions/node/v5.6.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 538, in main
    return gyp_main(args)
  File "/home/dominic/.nvm/versions/node/v5.6.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 514, in gyp_main
    options.duplicate_basename_check)
  File "/home/dominic/.nvm/versions/node/v5.6.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 130, in Load
    params['parallel'], params['root_targets'])
  File "/home/dominic/.nvm/versions/node/v5.6.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 2775, in Load
    variables, includes, depth, check, True)
[...]

PR #1085, opened 2017-01-05
skip ValueError for long, list and dict type;

vany-egorov:
skip ValueError for long, list and dict type;
refs https://github.com/yarnpkg/yarn/issues/2286
refs https://github.com/yarnpkg/yarn/issues/2376

### DIFF
--- a/gyp/pylib/gyp/input.py
+++ b/gyp/pylib/gyp/input.py
@@ -1222,9 +1222,9 @@ def ProcessVariablesAndConditionsInDict(the_dict, phase, variables_in,
     # Skip "variables", which was already processed if present.
     if key != 'variables' and type(value) is str:
       expanded = ExpandVariables(value, phase, variables, build_file)
-      if type(expanded) not in (str, int):
+      if type(expanded) not in (str, int, long):
         raise ValueError(
-              'Variable expansion in this context permits str and int ' + \
+              'Variable expansion in this context permits str and int/long ' + \
               'only, found ' + expanded.__class__.__name__ + ' for ' + key)
       the_dict[key] = expanded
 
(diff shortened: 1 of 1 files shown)


gyp/pylib/gyp/input.py
```diff
@@ -1222,6 +1222,8 @@ def ProcessVariablesAndConditionsInDict(the_dict, phase, variables_in,
     # Skip "variables", which was already processed if present.
     if key != 'variables' and type(value) is str:
       expanded = ExpandVariables(value, phase, variables, build_file)
+      if type(expanded) in (long, list, dict):
+        continue
```
bnoordhuis: This doesn't look self-evidently correct to me.  What is yarn doing that npm doesn't that causes ExpandVariables to return unexpected types?  `long` is acceptable because `int` is accepted but why `list` and `dict`?  With what keys do you get those?

bnoordhuis: Seems I answered that question 8 months ago in https://github.com/nodejs/node-gyp/issues/903#issuecomment-209012970...  I suppose the question then becomes whether this is the only ExpandVariables call site that needs updating.

vany-egorov: I added extra debug lines.
I got only `long`. No `list` or `dict`.

```json
[
  {"type(expanded)": "<type 'long'>", "expanded": "1483986851164", "value": "1483986851164", "phase": "0", "build_file": "binding.gyp"}
  {"type(expanded)": "<type 'long'>", "expanded": "1483986851164", "value": "1483986851164", "phase": "0", "build_file": "deps\\libexpat\\libexpat.gyp"}
]
```

PR #1092, opened 2017-01-20
lib: add gyp.js support

indutny:
See: #960

Added initial `gyp.js` support with --gypjs command line option.
Environment variable `npm_config_gypjs` also turns this option on.

Update configure and build usage strings depending on `npm_config_gypjs`
environment variable.

Set `npm_config_gypjs` env variable if `--gypjs` command-line option
was set to affect usage text for `configure` and `build` commands.

Update usage strings if `--gypjs` command-line option was supplied

Trying to load gyp.js module only if --gypjs command-line option was supplied.

cc @bnoordhuis 

See #962

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,37 @@ you can require the `.node` file with Node and run your tests!
 __Note:__ To create a _Debug_ build of the bindings file, pass the `--debug` (or
 `-d`) switch when running either the `configure`, `build` or `rebuild` command.
 
+#### Usage without Python
+
+There is a [`gyp.js`](https://github.com/indutny/gyp.js/) project, a GYP
+implementation in JavaScript. It generates [`Ninja`](https://ninja-build.org/)
+build files and requires no Python installation.
+
+In this case you will need to install [`Ninja`](https://ninja-build.org/) build
+tool and a C/C++ compiler toolchain.
+
+To generate projects files with `gyp.js` instead of `gyp` and to build them with
+`ninja`, please supply a command-line option `--gypjs` in `node-gyp` command.
+
+``` bash
+$ node-gyp configure --gypjs
+```
+
+It is also possible to set `npm_config_gypjs` environment variable to turn on
+the `gyp.js` usage.
+
+``` bash
+$ npm_config_gypjs=1 node-gyp build
+```
+
+Path to an existing `ninja` installation can be set with a `--ninja` command-line
+option or in a `NINJA` environment variable.
+
+``` bash
+$ node-gyp configure --gypjs --ninja=/my/path/to/ninja
+$ npm_config_gypjs=1 NINJA=/my/path/to/ninja node-gyp build
+```
+
 
 The "binding.gyp" file
 ----------------------
@@ -184,6 +215,8 @@ Command Options
 | `--python=$path`                  | Set path to the python (2) binary
 | `--msvs_version=$version`         | Set Visual Studio version (win)
 | `--solution=$solution`            | Set Visual Studio Solution version (win)
+| `--gypjs`                         | Use gyp.js instead of gyp
+| `--ninja=$ninja`                  | Override ninja command (with --gypjs)
 
 
 License
--- a/lib/build.js
+++ b/lib/build.js
@@ -11,12 +11,12 @@ var fs = require('graceful-fs')
   , glob = require('glob')
   , log = require('npmlog')
   , which = require('which')
-  , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
 
-exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
+exports.usage = 'Invokes `' + (process.env.npm_config_gypjs ? 'ninja' :
+  (win ? 'msbuild' : 'make')) + '` and builds the module'
 
 function build (gyp, argv, callback) {
   var platformMake = 'make'
@@ -36,8 +36,10 @@ function build (gyp, argv, callback) {
     , config
     , arch
     , nodeDir
-    , copyDevLib
 
+  if (gyp.opts.gypjs) {
+    command = gyp.opts.ninja || process.env.NINJA || 'ninja'
+  }
   loadConfigGypi()
 
   /**
@@ -60,7 +62,6 @@ function build (gyp, argv, callback) {
       buildType = config.target_defaults.default_configuration
       arch = config.variables.target_arch
       nodeDir = config.variables.nodedir
-      copyDevLib = config.variables.copy_dev_lib == 'true'
 
       if ('debug' in gyp.opts) {
         buildType = gyp.opts.debug ? 'Debug' : 'Release'
@@ -73,7 +74,7 @@ function build (gyp, argv, callback) {
       log.verbose('architecture', arch)
       log.verbose('node dev dir', nodeDir)
 
-      if (win) {
+      if (win && !gyp.opts.gypjs) {
         findSolutionFile()
       } else {
         doWhich()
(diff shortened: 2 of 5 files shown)


indutny:
cc @pmed

pmed:
Hi Fedor @indutny 

I glad to see that gyp.js feature is again moving forward 👍

Is it a rebased version of #962? If so, please note that PR is based on #964, and those two PRs probably should be closed after merging of this one.

README.md
```diff
@@ -184,6 +215,8 @@ Command Options
 | `--python=$path`                  | Set path to the python (2) binary
 | `--msvs_version=$version`         | Set Visual Studio version (win)
 | `--solution=$solution`            | Set Visual Studio Solution version (win)
+| `--gypjs`                         | Use gyp.js instead of gyp
```
jbergstroem: Perhaps mention that this can go away at any stage until we can take a decision about whether or not we should keep it. Both in readme and some smaller note here.

lib/build.js
```diff
@@ -212,57 +189,72 @@ function build (gyp, argv, callback) {
...
-      argv.push('build')
+      // build with ninja
+      if (verbose) {
+        argv.push('-v')
+      }
+      // Specify the build type, Release by default
+      argv.push('-C', path.join('build', buildType))
       if (jobs) {
-        var j = parseInt(jobs, 10)
+        var j = jobs.toUpperCase() === 'MAX'? require('os').cpus().length : parseInt(jobs, 10)
```
jbergstroem: nit: space before `?`

Fishrock123:
Why does this require Ninja but default builds don't?

indutny:
@Fishrock123 it doesn't, default slow version of ninja is included. If it wouldn't be ninja, we'd have to use MSVS on Windows and make on Unixes, meaning that gyp.js will have to support two platform-specific generators instead of one cross-platform generator.

indutny:
@bnoordhuis PTAL

Fishrock123:
@indutny Oh, I was wondering why gyp.js required it when regular gyp didn't? Just maintenance cost of that?

indutny:
@Fishrock123 yep, maintenance cost and a single solution for all platforms.

mko-io:
any update?

jskorepa:
What is status of this?

README.md
```diff
@@ -108,6 +108,37 @@ you can require the `.node` file with Node and run your tests!
...
+tool and a C/C++ compiler toolchain.
+
+To generate projects files with `gyp.js` instead of `gyp` and to build them with
+`ninja`, please supply a command-line option `--gypjs` in `node-gyp` command.
+
+``` bash
+$ node-gyp configure --gypjs
+```
+
+It is also possible to set `npm_config_gypjs` environment variable to turn on
```
refack: IMHO no need for this after #1185

lib/build.js
```diff
@@ -11,12 +11,12 @@ var fs = require('graceful-fs')
   , glob = require('glob')
   , log = require('npmlog')
   , which = require('which')
-  , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
 
-exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
+exports.usage = 'Invokes `' + (process.env.npm_config_gypjs ? 'ninja' :
```
refack: Can move to ` templates

lib/configure.js
```diff
@@ -21,27 +21,39 @@ var fs = require('graceful-fs')
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
   , findNodeDirectory = require('./find-node-directory')
-  , msgFormat = require('util').format
+  , gypjs = undefined
 
-exports.usage = 'Generates ' + (win ? 'MSVC project files' : 'a Makefile') + ' for the current module'
+exports.usage = 'Generates ' + (process.env.npm_config_gypjs ? 'ninja build files' :
```
refack: template

refack (changes requested):

refack:
@indutny You up for this? I'd be happy to help/take over...

indutny:
@refack feel free to take over.

refack:
@indutny Ack

Croydon:
Any progress so far?

bnoordhuis:
I've been thinking about this for some time and I've come to the conclusion that I'd rather not have node-gyp be in the business of generating platform- and tooling-specific build files.

That's what gyp.js does, it writes ninja files tailored for the platform and the compiler.  As the upstream GYP project has demonstrated, keeping up with platforms and toolchains is a never-ending slog.  It takes a lot of time and energy that is better spent elsewhere.

I'm hoping we can leverage an existing tool.  My personal preference is [cmake](http://cmake.org/); it's stable and just works.  Now to write a tool that translates GYP files to CMakeLists...

@indutny WDYT?  Reasonable, unreasonable?

LinusU:
I've been using CMake.js for at least one project, and it works amazingly good! It only requires the cmake binary instead of relying on python which is a huge plus, I also prefer writing cmake files to gyp file to be honest...

https://github.com/cmake-js/cmake-js#why-cmake

jbergstroem:
I like `gyp.js` because of the vision being self-hosted with installing node.js. I'm totally fine if node.js itself uses something else; but a big community win would be having `gyp.js` Good Enough to build a majority of the `node-gyp` projects.

indutny:
@bnoordhuis I'm fine with compiling GYP files to CMake in addons, but I have no idea if this is possible in general.

IMO, gyp.js + ninja.js is more viable and as @jbergstroem said is self-hosted.

aduh95:
With Node.js v4 reaching End Of Life in two weeks, `gyp.js` will be compatible with all supported Node versions! Maybe we could consider getting rid of the Python dependency. What is the status of this? Is there something I can do to help?

bnoordhuis:
I wouldn't mind revisiting this if @indutny is up to it.

AlJohri:
@indutny any chance we could revive this PR?

indutny:
Of course. Anyone wants to pick it up from me? I haven't touched this code in months now.

rvagg:
I was thinking of picking this up again but only just realised that gyp.js requires ninja. I'm really not a fan of ninja simply because it's not commonly available. We're replacing a python dependency with a ninja dependency, that's arguably asking for more trouble because at least python is easy to fetch and install on every platform. We're going to have to get into the business of shipping ninja binaries like Chromium does.
Maybe getting it beyond ninja is a reasonable task, but that's not a trivial amount of work.

indutny:
There is [ninja.js](https://github.com/indutny/ninja.js) FWIW.

rvagg:
@indutny how mature is that? If I ship a node-gyp with --gypjs support, what's your rough % guess for how much of the current ecosystem is going to be able to successfully build _without_ requiring additional tools (python, ninja, just make and c++)? I'm wondering how much investment I'm going to have to make if I take this on.

indutny:
This is a very good question. Most addons have very simple (if not identical) configuration. gyp.js should handle it well out of box.

The only limiting factor is "python" support: https://github.com/indutny/gyp.js#reduced-python . gyp.js emulates some trivial python commands, but of course cannot do full python support.

rvagg:
Since we're the only remaining major user of GYP I suspect we'll be able to move the ecosystem to not depend on sophisticated python syntax in their gyp files. Maybe we could even move it to support some JavaScript if people really need advanced stuff in there.

Kreijstal:
does gyp.js support mingw builds with msys2?
We had to patch node_gyp because it assumes windows->msvc without considering using gcc.. aka mingw

cclauss:
22 issues mentioning `mingw`: https://github.com/search?q=repo%3Anodejs%2Fnode-gyp+mingw&type=issues

Kreijstal:
> 22 issues mentioning `mingw`: https://github.com/search?q=repo%3Anodejs%2Fnode-gyp+mingw&type=issues

I have managed to get mingw working with node-gyp, but it is nothing that I can upstream as it simply removes the msvc hardcoded version and hardcodes mingw on its place.. nothing dynamic

Issue #1056, opened 2016-11-21
Visual Studio 2017


felixfbecker:
VS 2017RC1 is out now and has a new installer that allows even finer cherry-picking of the wanted components. What is the minimal set of components that need to be installed for using node-gyp? Could this be added to README?

rvagg:
Yeah, I've been wondering the same thing, @nodejs/platform-windows, @joaocgreis, @felixrieseberg, @mousetraps can anyone fill us in on details on what's needed from VS 2017 to compile Node addons? Perhaps prepare a PR in anticipation of 2017 going gold?

felixfbecker:
I have tried getting this to work with VS 2017 for a few days now without success.
- Installing just "VC++2017 Compiler Toolset" is not enough and results in an error message from MSBuild that VCBuild.exe could not be found. A search inside the installation directories reveals that VCBuild.exe indeed wasn't installed anywhere. Maybe it is installed as part of .NET Framework?
- Setting `msvs_version` to `2017` results in an "undefined key" error from node-gyp.

rvagg:
Ah, it looks like they're not even prepared for it upstream @ gyp: https://chromium.googlesource.com/external/gyp/+/master/pylib/gyp/MSVSVersion.py#238

joshgav:
+ @digitalinfinity

felixrieseberg:
Heya! I'm at Slack now, but still pretty invested in things being "okay" on Windows. I have a feeling this will be a tough one, but once things are fixed upstream, we should try to get https://github.com/felixrieseberg/windows-build-tools up to par with what'll be needed for most tools.

felixfbecker:
Not so much interested in installing them from NPM, I just want to know what minimal components I need from the new (leightweight) VS installer

Fishrock123:
@felixrieseberg Status update?

felixfbecker:
Is there an issue open at gyp about this?

mousetraps:
Sorry, just saw this thread. Not surprising given that both the msbuild resolution scheme and the installer paths have been completely revamped in VS2017, so I suspect we'll, at the very least, need to make some modifications to node-gyp itself to properly locate the path when setting msvs_version. I'll investigate what we need to do to make this work again and report back.

joshgav:
@mousetraps good to see you :)

/cc @kunalspathak who I believe has been looking at VS 2017 support as well.

mousetraps:
[...]

PR #1101, opened 2017-01-31
Add support for Visual Studio 2017

bzoz:
Add support for Visual Studio 2017 to node-gyp

With this version VS does not use Windows registry any more, instead it installs a COM server that one can query to obtain list of installed modules and their locations - see this blogpost: https://blogs.msdn.microsoft.com/heaths/2016/09/15/changes-to-visual-studio-15-setup/. 

Python does not support communicating with COM servers directly. Chromium project when updating `GN` did not add extra library, instead assuming default installation path with environment variable override (see https://cs.chromium.org/chromium/src/build/vs_toolchain.py?q=vs_toolchain&sq=package:chromium&l=150). I've used the same method  to update `GYP` - default installation path and default Windows SDK version with override. PR for `GYP` has been also opened upstream: https://chromium-review.googlesource.com/#/c/433540/

To ease usage for node-gyp users, a Python script that can obtain VS installation path and installed SDK version is provided. It depends on `comtypes` Python library (https://github.com/enthought/comtypes) which adds support for COM to Python. This script is in turn used by `node-gyp` to set proper environment variables overrides.

For easier review this PR is split into 4 commits:

  * First commit adds `comtypes` Python library. It is needed to query VS COM server about installation paths.

  * Second commit adds support for VS2017 to `gyp`. Similar to the way it was implemented in `gn`, it assumes that VS2017 Professional was installed in the default location. User can use two environment variables to overwrite this: `VS2017_INSTALL` for installation path and `VS2017_SDK` for installed SDK version. This patch was also commited upstream: https://chromium-review.googlesource.com/#/c/433540/

  * Third commit moves `PythonFinder` to separate file. We will use Python in build stage to locate new `msbuild` installation location.

  * Fourth commit adds VS2017 support to `node-gyp`. It includes Python script that can communicate with VS2017 COM server, and necessary changes to configure and build stages.
 

Ref: https://github.com/nodejs/node/pull/11084,
https://github.com/nodejs/node-gyp/issues/1056


refack:
@bzoz Great job tracking down the COM api 👍 but isn't `comtypes` overkill, you can just trigger COM objects via powershell. Or I can compile a tiny tool for that.

refack:
(tiny tool)[https://drive.google.com/open?id=0Bz0LZMH4OpErR1JHZHZvSWI0NEE] (29KB) made from [Microsoft/vs-setup-samples](https://github.com/Microsoft/vs-setup-samples).
It has easy to parse text output, e.g.
```
C:\temp$ Setup.Configuration.CS.exe
InstanceId: 45ec1659 (Complete)
InstallationVersion: 15.0.26020.0 (4222126355906560)
InstallationPath: D:\bin\dev\VS\2017\BuildTools
Product: Microsoft.VisualStudio.Product.BuildTools
Workloads:
    Microsoft.VisualStudio.Workload.MSBuildTools
    Microsoft.VisualStudio.Workload.VCTools
```
and the env manipulation bat is at ```${InstallationPath}\Common7\Tools\VsMSBuildCmd.bat```

refack:
VS2017 build tools on Windows 7SP1:
```
gyp ERR! stack Error: Command failed: c:\bin\dev\python27\python.EXE \\REFAELUX\code\3party\node-gyp\find_vs2017.py
gyp ERR! stack Traceback (most recent call last):
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\find_vs2017.py", line 173, in <module>
gyp ERR! stack     main()
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\find_vs2017.py", line 158, in main
gyp ERR! stack     installs = GetVS2017CPPBasePath()
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\find_vs2017.py", line 106, in GetVS2017CPPBasePath
gyp ERR! stack     iface = CreateObject(GUID('{177F0C4A-1CD3-4DE7-A32C-71DBBB9FA36D}'))
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\comtypes\comtypes\client\__init__.py", line 238, in CreateObject
gyp ERR! stack     obj = comtypes.CoCreateInstance(clsid, clsctx=clsctx, interface=interface)
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\comtypes\comtypes\__init__.py", line 1225, in CoCreateInstance
gyp ERR! stack     _ole32.CoCreateInstance(byref(clsid), punkouter, clsctx, byref(iid), byref(p))
gyp ERR! stack   File "_ctypes/callproc.c", line 945, in GetResult
gyp ERR! stack WindowsError: [Error -2147221164] Class not registered
```
Apparently the COM interface doesn't get registered
(I get the same with the above COM calling EXE)

bzoz:
@refack sorry, I cannot reproduce. After installing "Build Tools for Visual Studio 2017 RC" on Windows 7 python script can locate it.

However, I can get this error for clean Windows installation, when VS2017 was never installed. Fixed the script.

refack:
I trying with some minimal component set and found that the COM server wasn't registered...

P.S. funny thing, the Installer is an `electron` app

heaths:
I own the COM APIs. They are registered when you install VS, though in some older previews there were some circumstances in which they would get prematurely unregistered. That was fixed, IIRC, in RC2.

I recommend using the native COM APIs with `IUnknown` bindings, which would be easier to use in C++. Or is this about setting up a node-gyp project rather than using in a Node native add on?

refack:
@heaths thank for the reply. Since this is a nodejs / python project, we're trying not to cloud it with other languages or precompiled EXEs. I just converted my C# code to be .NET 2.0 compatible so now we have a way to call it from powershell 2.0.
As for the registration, at one instance I installed the "Build Tools for Visual Studio 2017 RC" on windows 7 pro SP1, in a very minimal way and the COM wasn't registered... I'll try to repro.

heaths:
Sounds like the build tools are missing a dependency, then. I'll track this down.

refack:
@heaths moving to Microsoft/vs-setup-samples#5

bzoz:
Not needed since https://github.com/nodejs/node-gyp/pull/1130 landed.

AndrewPardoe:
The officially supported vswhere tool is now available as part of the install starting today with Visual Studio 15.2 preview 2: https://blogs.msdn.microsoft.com/heaths/2017/04/21/vswhere-is-now-installed-with-visual-studio-2017/. This doesn't fix all the existing installs of VS 2017 but going forward you can now detect the VS install locations and whether C++ tools are available from a scripted environment. I hope this helps your scenario.

PR #1103, opened 2017-02-02
Detect VS2017

refack:
Based on @bzoz's #1101 but without python `comtypes'

Also bumped `gyp` to latest version and updated out patch files

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -14,7 +14,9 @@ var fs = require('graceful-fs')
   , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
-  , win = process.platform === 'win32'
+  , win = process.platform === 'win32';
+if (win)
+  var findVS = require('windows-autoconf')
 
 exports.usage = 'Invokes `' + (win ? 'msbuild' : 'make') + '` and builds the module'
 
@@ -106,7 +108,7 @@ function build (gyp, argv, callback) {
     which(command, function (err, execPath) {
       if (err) {
         if (win && /not found/.test(err.message)) {
-          // On windows and no 'msbuild' found. Let's guess where it is
+          log.verbose('could not find "msbuild.exe" in PATH - finding location in registry')
           findMsbuild()
         } else {
           // Some other error or 'make' not found on Unix, report that to the user
@@ -122,68 +124,18 @@ function build (gyp, argv, callback) {
   /**
    * Search for the location of "msbuild.exe" file on Windows.
    */
-
   function findMsbuild () {
-    if (config.variables.msbuild_path) {
-      command = config.variables.msbuild_path
-      log.verbose('using MSBuild:', command)
-      copyNodeLib()
-      return
-    }
-
-    log.verbose('could not find "msbuild.exe" in PATH - finding location in registry')
-    var notfoundErr = 'Can\'t find "msbuild.exe". Do you have Microsoft Visual Studio C++ 2008+ installed?'
-    var cmd = 'reg query "HKLM\\Software\\Microsoft\\MSBuild\\ToolsVersions" /s'
-    if (process.arch !== 'ia32')
-      cmd += ' /reg:32'
-    exec(cmd, function (err, stdout, stderr) {
-      if (err) {
-        return callback(new Error(err.message + '\n' + notfoundErr))
+    var notfoundErr = 'Can\'t find "msbuild.exe". Do you have Microsoft Visual Studio installed?'
+    try {
+      var msbuild_path = findVS.locateMsbuild()
+      if (!msbuild_path) {
+        return callback(new Error(notfoundErr))
       }
-      var reVers = /ToolsVersions\\([^\\]+)$/i
-        , rePath = /\r\n[ \t]+MSBuildToolsPath[ \t]+REG_SZ[ \t]+([^\r]+)/i
-        , msbuilds = []
-        , r
-        , msbuildPath
-      stdout.split('\r\n\r\n').forEach(function(l) {
-        if (!l) return
-        l = l.trim()
-        if (r = reVers.exec(l.substring(0, l.indexOf('\r\n')))) {
-          var ver = parseFloat(r[1], 10)
-          if (ver >= 3.5) {
-            if (r = rePath.exec(l)) {
-              msbuilds.push({
-                version: ver,
-                path: r[1]
-              })
-            }
-          }
-        }
-      })
-      msbuilds.sort(function (x, y) {
-        return (x.version < y.version ? -1 : 1)
-      })
-      ;(function verifyMsbuild () {
-        if (!msbuilds.length) return callback(new Error(notfoundErr))
-        msbuildPath = path.resolve(msbuilds.pop().path, 'msbuild.exe')
-        fs.stat(msbuildPath, function (err, stat) {
-          if (err) {
-            if (err.code == 'ENOENT') {
-              if (msbuilds.length) {
-                return verifyMsbuild()
-              } else {
-                callback(new Error(notfoundErr))
-              }
-            } else {
-              callback(err)
-            }
-            return
-          }
-          command = msbuildPath
-          copyNodeLib()
-        })
-      })()
-    })
+    } catch (e) {
+      return callback(new Error(err.message + '\n' + notfoundErr))
+    }
+    command = msbuild_path
+    copyNodeLib()
   }
 
   /**
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -23,7 +23,7 @@ var fs = require('graceful-fs')
   , findNodeDirectory = require('./find-node-directory')
   , msgFormat = require('util').format
 if (win)
-  var findVS2017 = require('./find-vs2017')
+  var findVS = require('windows-autoconf')
 
 exports.usage = 'Generates ' + (win ? 'MSVC project files' : 'a Makefile') + ' for the current module'
 
@@ -41,15 +41,13 @@ function configure (gyp, argv, callback) {
       callback(err)
     } else {
       python = found
+      // 'python' should be set by now
+      process.env.PYTHON = python
       getNodeDir()
     }
   })
 
   function getNodeDir () {
-
-    // 'python' should be set by now
-    process.env.PYTHON = python
-
     if (gyp.opts.nodedir) {
       // --nodedir was specified. use that for the dev files
       nodeDir = gyp.opts.nodedir.replace(/^~/, osenv.home())
@@ -88,22 +86,29 @@ function configure (gyp, argv, callback) {
     mkdirp(buildDir, function (err, isNew) {
       if (err) return callback(err)
       log.verbose('build dir', '"build" dir needed to be created?', isNew)
-      if (win && (!gyp.opts.msvs_version || gyp.opts.msvs_version === '2017')) {
-        findVS2017(function (err, vsSetup) {
-          if (err) {
-            log.verbose('Not using VS2017:', err.message)
-            createConfigFile()
-          } else {
-            createConfigFile(null, vsSetup)
-          }
-        })
-      } else {
-        createConfigFile()
-      }
+      createConfigFile()
     })
   }
 
-  function createConfigFile (err, vsSetup) {
+  function findVisualStudio2017 (defaults) {
+    if (gyp.opts.msvs_version && gyp.opts.msvs_version !== '2017')
+      return
+
+    try {
+      var vsSetup = findVS.getVS2017Setup()
+      if (!vsSetup || !vsSetup.InstallationPath) return
+    } catch (_) {
+      return
+    }
+
+    gyp.opts.msvs_version = '2015'
+    process.env['GYP_MSVS_VERSION'] = 2015
+    process.env['GYP_MSVS_OVERRIDE_PATH'] = vsSetup.InstallationPath
+    defaults['msbuild_toolset'] = 'v141'
+    defaults['msvs_windows_target_platform_version'] = vsSetup.SDK
+  }
+
+  function createConfigFile (err) {
     if (err) return callback(err)
 
     var configFilename = 'config.gypi'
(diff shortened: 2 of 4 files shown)


refack:
(changed COM tool from an exe to a `powershell` script)

joaocgreis:
Node-gyp must support Windows 7/2008R2 while Node does, and those versions of Windows ship with PowerShell 2.0. The module you use here seems to require at least version 3. We could ask users to update PowerShell, but there have been issues in the past from users locked in corporate environments with no permissions to install anything, so this might be troublesome to most and impossible to some.

Using an executable opens an issue of trust in the committed file. We could get around this by setting up a job in node's release infrastructure to compile and sign an executable to use here, but only if must be.

@refack what issue do you see with `comtypes`? I'd rather not add the dependency, but so far it seems to be the easiest for all users. Either way we decide to go, thanks for submitting this, it's great to be able to broaden discussion by having more alternatives!

refack:
I'll check if I can back port the script...
Keep in mind, it's a fall back to try to detect VS2017 specificaly, doesn't it's installation upgrades `powershell` anyway?

* as for the exe, I get it.
** as for `comtypes` I think it's major overkill, and the "prevailing winds" is to reduce dependency on python, not increase it for this "tiny" COM query 🤔😕

bzoz:
@refack I just checked this on Windows 7 - after installing VS2017 and selecting Desktop Development with C++, powershell is still at version 2

refack:
Yep just did that too 😣 (also encountered the `ExecutionPolicy` issue)

refack:
P.S. same config (VS2017 build tools on Windows 7SP1):
```
gyp ERR! stack Error: Command failed: c:\bin\dev\python27\python.EXE \\REFAELUX\code\3party\node-gyp\find_vs2017.py
gyp ERR! stack Traceback (most recent call last):
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\find_vs2017.py", line 173, in <module>
gyp ERR! stack     main()
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\find_vs2017.py", line 158, in main
gyp ERR! stack     installs = GetVS2017CPPBasePath()
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\find_vs2017.py", line 106, in GetVS2017CPPBasePath
gyp ERR! stack     iface = CreateObject(GUID('{177F0C4A-1CD3-4DE7-A32C-71DBBB9FA36D}'))
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\comtypes\comtypes\client\__init__.py", line 238, in CreateObject
gyp ERR! stack     obj = comtypes.CoCreateInstance(clsid, clsctx=clsctx, interface=interface)
gyp ERR! stack   File "\\REFAELUX\code\3party\node-gyp\comtypes\comtypes\__init__.py", line 1225, in CoCreateInstance
gyp ERR! stack     _ole32.CoCreateInstance(byref(clsid), punkouter, clsctx, byref(iid), byref(p))
gyp ERR! stack   File "_ctypes/callproc.c", line 945, in GetResult
gyp ERR! stack WindowsError: [Error -2147221164] Class not registered
```
Apparently the COM interface doesn't get registered

refack:
Unless you install the .NET 4.6 SDK

joaocgreis:
@refack any luck with PowerShell v2?

refack:
@joaocgreis just cracked it (and added an undocumented Registry approach as backup)

joaocgreis:
I've tested, and I could not get this to work because of spaces in the path. Also, the registry key approach does not provide the SDK.

@heaths this looks for the path in the registry at `HKLM\Software\Microsoft\VisualStudio\SxS\VS7`. Can we trust the value in this registry key? Is it likely to be removed in the future?

heaths:
Paths could always have spaces. Make sure the path is quoted and you shouldn't have a problem. No registry values are supported for VS2017. Whatever you found is probably just a left-over that should've been removed. MSIs are singletons and may have nothing to do with where VS instances are installed.

refack:
Fixed path escaping issue, and now the code works with [Powershell "core"](https://github.com/PowerShell/PowerShell/).
Registry "hack" is a backup measure anyway ...

P.S. I refactored my code into a [package](https://github.com/node4good/get-vs2017-path) (it's has no dependecies of it's own) what might be the push back of adding it as a dependency?

refack:
As my [detection package](https://github.com/node4good/get-vs2017-path) is getting traction, and bug fixes, I believe it's worth the add.

@bnoordhuis @kunalspathak I think now it's mature enough to cc you :)

heaths:
While I appreciate the community working together to fill gaps - being a long-time open source contributor myself - I'm curious how using the batch scripts are easier than acquiring the PowerShell module:

```powershell
install-module vssetup -scope currentuser # only needed once
$installationPath = get-vssetupinstance | select-vssetupinstance -latest | select-object -expand InstallationPath
```

I will be keeping these up to date with new functionality as we bring it online.

AndrewPardoe:
First, I agree with Heath: The PowerShell solution is probably best for most users.

I had a discussion with the Boost folks that lays out the options. This thread may be interesting to you: https://github.com/boostorg/build/issues/157

refack:
@heaths It's a change to the underlying system. While reasonable for developers, `node-gyp` is an automation module that runs as part of setup of different systems.
Also one tenet of the `node.js` way is "Keep everything local /  Leave no trace" emblemized with the ubiquitous `node_molues` directory (with all it's pros and cons).
Second: it require access to the internet, and permission to install such things. Again, not an assumption `node-gyp` can take.
If you can escalate these concerns, that would be great. I believe the "C++ Build Tools" configuration is an example of helping these use cases.

heaths:
@refack FWIW, I'm working to finish up a single native binary that can return information in a number of formats. A prototype actually ships in the _tools_ subdirectory of the Microsoft.VisualStudio.Setup.Configuration.Native NuGet package, which is what several VC projects are using. You might take a look at that, as the eventual new project will also grow over time. These will be fully supported along with the other current projects.

refack:
@heaths That's sounds promising. Maybe also also consider implementing `IDispatch` so we could use it directly with Powershell's `New-Object -ComObject` and `cscript` / `JScript`?

refack:
@heaths Where's the best place to leave feedback on "Microsoft.VisualStudio.Setup.Configuration.Native"? I have one request; we need the version number for installed SDKs so we could populate the `<WindowsTargetPlatformVersion>10.0.14393.0</WindowsTargetPlatformVersion>` field in `vcxproj` files

refack:
P.S. IMHO `nuget` is very `node.js`y 😜

AndrewPardoe:
@refack, given that NuGet (formerly NuPack) and node.js were designed around the same time, that's not a surprise. They both reflect how the world was thinking about packages at the time.

heaths:
@refack re: PowerShell, people should just use the VSSetup module. Even with `IDispatch` support, enumerating instances through the `IEnumXXX`-style interface is a PITA and doesn't jive with PowerShell. The point of the module was to make something easily acquirable and easy to use. If you don't want to take a dependency on a network connection, it's license to redistribute.

We can consider adding `IDispatch` for scripting support, but PowerShell was meant to replace all that. It would be a lower priority item.

re: feedback, consider it heard. We can consider how to do that, but the data on which the query API works doesn't store that. The more data we store, we slow the query API down. That's only one type of question to ask, but I'm sure other people have others. We can't really store every conceivable answer. Instead, it would be better that the Windows SDK - which is actually a system singleton (i.e. it's still MSI-based and will be for the foreseeable future since it's so big and used by things outside VS) - write something like that. I will pass along feedback.

For now, you could do something like the following as long as their package ID pattern holds:

```powershell
get-vssetupinstance | select -expand Packages | where Id -match 'Win10SDK_\d' | select Version
```

refack:
@heaths Thanks again for listening. (as for figuring out the SDK [that's almost exaclty what I'm doing in my CS script](https://github.com/node4good/get-vs2017-path/blob/master/tools/Get-VS7.cs#L199))

joaocgreis:
@heaths we must support Windows 7 and 2008R2 as shipped, with Powershell v2. I did not manage to get the VSSetup module to work with it, is there a way? I'm not thrilled about landing a binary either - even if we can trust the binary's content we might run into issues with antivirus software because node-gyp must be usable as downloaded from the internet (e.g. after updating npm). Currently, the solution here, using the COM API, seems to be the best one.

@refack can you move the gyp bump to a different PR, or is it required for this? Also, the [commit with the Gyp floating patch](https://github.com/nodejs/node-gyp/commit/3f4be7aff1acdf1001e33e698eb4a4b3d1597498) that you had before looked good, can you bring it back (or is there a reason not to)?

About using a package: @refack can you commit to always supporting the same versions of Windows that node does? @nodejs/node-gyp are there other reasons why we might not want this as a package?

refack:
@joaocgreis RE floating patch: #1118 #1119 The `GYP` bump is needed as it adds some VS2017 logic, so tests that actually `build` would have failed. I tried to have them in separate commits (pure `GYP` bump, and our patches).
RE: package support, `Windows` versions yes! For now as for `VS2017`, either work or fail gracefully. for future versions I want to support more MSVS versions (a kind of `autoconf` for Windows)...

I Did want to ask which `node` versions to support?

joaocgreis:
Floating patches: Thanks for pointing that out, point taken.

Node versions: All current + LTS branches, that is v4, v6, v7 and master. I will give this some testing before landing, CitGM will test more when releasing. We must be careful though, because CitGM only runs in VS2015.

refack:
Versions: Noted.

Well for now CitGM will protect us from regressions. I could write a patch for CitGM that will [bootstrap VS2017](https://docs.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio) and run with it. I just need you to point me in the right direction...

joaocgreis:
Installing from CitGM would be a problem, the machines we use have fixed VS versions (and the job is not sandboxed). I'm looking into getting some more resources to run VS2017 or I'll reprovision some of what we have.

lib/configure.js
```diff
@@ -39,10 +41,23 @@ function configure (gyp, argv, callback) {
...
+      if (!vsSetup) return callback()
+      gyp.opts.msvs_version = '2017';
+      process.env['vs2017_install'] = vsSetup.InstallationPath;
+      if (vsSetup.SDK) process.env['vs2017_sdk'] = vsSetup.SDK.replace(/\d+$/, '0')
+      callback();
+    } else {
+      getNodeDir()
+    }
+  }
+
```
joaocgreis: `getNodeDir()` always needs to run, and this function should only run if there was no version of VS specified, or it was 2017. Errors finding VS2017 should be ignored and it should continue to find other versions. What about:

```js
  findPython(python, function (err, found) {
    if (err) {
      callback(err)
    } else {
      python = found
      if (win)
        findVisualStudio2017()
      getNodeDir()
    }
  })

  function findVisualStudio2017() {
    if (gyp.opts.msvs_version && gyp.opts.msvs_version !== '2017')
      return

    try {
      const vsSetup = findVS.getVS2017Setup()
    } catch (_) {}
    if (!vsSetup)
      return

    gyp.opts.msvs_version = '2017'
    process.env['vs2017_install'] = vsSetup.InstallationPath
    if (vsSetup.SDK)
      process.env['vs2017_sdk'] = vsSetup.SDK.replace(/\d+$/, '0')
  }
```

lib/build.js
```diff
@@ -122,6 +124,17 @@ function build (gyp, argv, callback) {
   /**
    * Search for the location of "msbuild.exe" file on Windows.
    */
+  function findMsbuild15() {
+    log.verbose('looking for VS2017 msbuild')
+    var msbuild_path = findVS.locateMsbuild()
```
joaocgreis: Errors finding VS2017 should be ignored, it should still be able to find other versions. Something like:

```js
    try {
      var msbuild_path = findVS.locateMsbuild()
    } catch (_) {}
```

refack:
`get-vs2017-path`@1.5 can replace almost all MSVS detecting logic... I'm testing on a plethora of systems (windows 7,8,10 vs MSVS 2010 / 2015 / 2017)

lib/build.js
refack: Done

refack:
Shrank it down to ~20 line (+ the `GYP` bump)

refack:
And now no `GYP` patch :)

joaocgreis:
I do not think I should sign this off using a module, I believe the code should be here where all collaborators control changes - at least until the module is proved stable and widely used. 

For now, we should move forward with a minimal approach using the PowerShell method that so far seems to work everywhere. @refack can you update using your code or can you review if I put together a PR?

refack:
I can do the minimal patch, give me a few minutes

refack:
Did a minimal patch, no package.
But I shall return
image
Muhhhhaaaaaaa

heaths:
I have published a release of vswhere - a single-file, native executable you can redistribute - at https://github.com/Microsoft/vswhere. It currently emits data in plan text and JSON and would be fairly easy to use. For example, to find the root path to the latest version installed for native tools:

```batch
for /r "usebackq tokens=1* delims=: " %%i in (`vswhere.exe -latest -requires Microsoft.VisualStudio.Workload.NativeDesktop`) do (
  if /i "%%i"=="installationPath" set dir=%%j
)

rem %dir% now contains the root installation path, if available
```

refack:
[I tweaked the `.bat` file](https://github.com/node4good/get-vs2017-path/blob/master/tools/try_powershell.cmd) to be more "polite" if it can't find the COM.
cc @joaocgreis

joaocgreis:
Reviewed and opened https://github.com/nodejs/node-gyp/pull/1130 with a slightly modified version.

refack:
I brough back [`windows-autoconf`](https://github.com/node4good/windows-autoconf), since it now has general MSVS detection logic, which enabled me to remove almost half of `build.js`.
I have it CI on a windows machine at AppVeyor [AppVeyor](https://ci.appveyor.com/project/refack/windows-autoconf)
I'm working on eliminating more of `node-gyp`'s VS decection code, and adding it's test suite as a dependency for passing the CI...

refack:
@bnoordhuis @joaocgreis 
Can we have vote this vs node-gyp#1130 ? See all the red lines, code you don't need to worry about ;)
cc @nodejs/node-gyp @nodejs/build

refack:
CI now passes `node-gyp` test suite [AppVeyor](https://ci.appveyor.com/project/refack/windows-autoconf)

refack:
Getting better coverage from [AppVeyor](https://ci.appveyor.com/project/refack/windows-autoconf) VS2015 and vs2017 pass `node-gyp`'s test suite

refack:
Great news we landed VS2017 support to `GYP` https://chromium-review.googlesource.com/c/453201/

AndrewPardoe:
The officially supported vswhere tool is now available as part of the install starting today with Visual Studio 15.2 preview 2: https://blogs.msdn.microsoft.com/heaths/2017/04/21/vswhere-is-now-installed-with-visual-studio-2017/. This doesn't fix all the existing installs of VS 2017 but going forward you can now detect the VS install locations and whether C++ tools are available from a scripted environment.

refack:
@AndrewPardoe Are you following my list from https://developercommunity.visualstudio.com/content/problem/35325/missing-vs150comntools.html?

AndrewPardoe:
@refack I wasn't, but I'm searching for all the places online where vswhere was mentioned and trying to link to this post : ) Thank you for the list!

Issue #1094, opened 2017-01-23
TypeError: Cannot read property 'length' of undefined in file semver/semver.js:239:14


AleksandrChernyavenko:
when I run it - parameter "version" === undefined

```

Error: gyp info it worked if it ends with ok
gyp info using node-gyp@3.5.0
gyp info using node@6.9.4 | linux | x64
gyp ERR! UNCAUGHT EXCEPTION 
gyp ERR! stack TypeError: Cannot read property 'length' of undefined
gyp ERR! stack     at parse (/home/alexander/www/electron/project-name/node_modules/semver/semver.js:239:14)
gyp ERR! stack     at Function.valid (/home/alexander/www/electron/project-name/node_modules/semver/semver.js:255:11)
gyp ERR! stack     at processRelease (/home/alexander/www/electron/project-name/node_modules/node-gyp/lib/process-release.js:16:25)
gyp ERR! stack     at configure (/home/alexander/www/electron/project-name/node_modules/node-gyp/lib/configure.js:35:17)
gyp ERR! stack     at Object.self.commands.(anonymous function) [as configure] (/home/alexander/www/electron/project-name/node_modules/node-gyp/lib/node-gyp.js:55:37)
gyp ERR! stack     at run (/home/alexander/www/electron/project-name/node_modules/node-gyp/bin/node-gyp.js:87:30)
gyp ERR! stack     at _combinedTickCallback (internal/process/next_tick.js:67:7)
gyp ERR! stack     at process._tickCallback (internal/process/next_tick.js:98:9)

```

For fix it I just add code above into function `function parse(version, loose) {`

```
  if(version === undefined) {
    return null;
  }
```

bnoordhuis:
Is there an easy way for me to reproduce?

Can you print the output of `node -p process.versions` and check what that "version" field in /home/alexander/www/electron/project-name/node_modules/semver/package.json says?

Is there a reason you have node-gyp in your node_modules directory instead of using the one that is bundled with node/npm?

TimHollies:
I'm encountering the same issue, and have come across this before. Last time I found that removing a particular package, which didn't require node-gyp compilation, fixed it for some reason. I had to find this package by trial and error, which takes a while with a large application. 

`node -p process.versions`

```
{ http_parser: '2.7.0',
  node: '7.5.0',
  v8: '5.4.500.48',
  uv: '1.10.2',
  zlib: '1.2.8',
  ares: '1.10.1-DEV',
  modules: '51',
[...]

PR #1107, opened 2017-02-06
Upgrade semver dependency.

bnoordhuis:
Should fix reports of "TypeError: Cannot read property 'length' of
undefined" runtime exceptions where node-gyp uses a buggy version
of semver due to our overly broad accepted version range.

npm updated to semver@5.3.0 recently so we can too now and still
share the module with npm, keeping the download small.

Fixes: https://github.com/nodejs/node-gyp/issues/1094

CI: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/35/

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "osenv": "0",
     "request": "2",
     "rimraf": "2",
-    "semver": "2.x || 3.x || 4 || 5",
+    "semver": "~5.3.0",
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Landed in 93d7fa8 with https://ci.nodejs.org/view/All/job/nodegyp-test-commit/38/ (all green) for good measure.

PR #1109, opened 2017-02-08
Added info about windows-build-tools and the python path to readme

Darkle:


README.md
```diff
@@ -42,7 +42,7 @@ You will also need to install:
       * You also need to install the `Command Line Tools` via Xcode. You can find this under the menu `Xcode -> Preferences -> Downloads`
       * This step will install `gcc` and the related toolchain containing `make`
   * On Windows:
-    * Option 1: Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
+    * Option 1: Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator). Note: although windows-build-tools will install Python, it will not add python to your PATH, so you will have to add it manually.
```
joshgav: Might be worth noting this PS command to add Python to your path (change `c:\python27` to wherever you installed Python):

```powershell
[System.Environment]::SetEnvironmentVariable('Path', "c:\python27;$env:Path", "Process")
```

The last param can be `Machine | User | Process`, but if you use only Machine or User it won't take effect till you relaunch PS or CMD sessions.

/cc @felixrieseberg - maybe you can add this to the scripts?

felixrieseberg: We're configuring npm to use python, but didn't configure `PATH` for fear that it might overwrite existing configurations. I'm super open to changing that though!

cc @A92hm: Thoughts?

A92hm: I think that should be optional. 

As @felixrieseberg said we didn't want to automatically add python to the`PATH` to prevent messing up users' configuration. But in some cases users don't have python installed and they want `windows-build-tools` to install and add python to their `PATH`. @joshgav Your proposed way looks good, but we need to add command line options to enable it.

This has been sitting in our backlog, but we haven't gotten into implementing it :)

joshgav: Cool, thanks for your work @A92hm and @felixrieseberg!

joaocgreis:
@Darkle Thanks for submitting this! While other options are being worked on for `windows-build-tools`, it doesn't hurt to have a warning here to clarify. However, in it's current form, I fear that this might transmit the idea that adding Python to the path is essential, which is not the case. What about something like:

> Note: although `windows-build-tools` will install Python, only npm will be configured to use it. If you want to use Python for other purposes, you can install it as you would normally since this will not conflict, or add this manually to the `PATH`.

felixrieseberg:
...and done :tada: You can now run `npm install --add-python-to-path --global windows-build-tools`.

danielsokolowski:
This should be merged in.

giuseppeferrante85:
Hi All

i had to unistall python 2.7 (from window-build-tool folder) forced by my company and use only python 3 (has been installed into another path) so i could link againg window-build-tool to the newone path?
thanks all

cclauss:
https://github.com/felixrieseberg/windows-build-tools installs Python 3.8

giuseppeferrante85:
@cclauss 

thank you Clauss but i Python it's already installed in my laptop .. i just need to bind windows-build-tool with the python folder.. i hope it's clear.. :-)

shamseer-ahammed:
+1

PR #1113, opened 2017-02-10
Add configuration for SSL verification when downloading content behind corporate proxy

slawomirbrys:
### Case
When trying to do `node-gyp install` then some packages are downloaded from internet resources. 

### Problem
If you trying to do that behind corporate proxy then additional option `--proxy` should be setup, but this option is not enough if SSL certificate is modified on this proxy, then below error occurs:
> Error: unable to verify the first certificate

### Solution
Turn off check of SSL certificate via setting `strictSSL` set to `false` on `request` object (default: `true`). Then get this value in command line via `--strictssl` option that accepts `true` or `false` value.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -443,10 +443,16 @@ function download (gyp, env, url) {
               || env.http_proxy
               || env.HTTP_PROXY
               || env.npm_config_proxy
+  
+  var strictSsl = true;
+  if (!gyp.opts.strictssl) {
+    strictSsl = gyp.opts.strictssl
+  }
   if (proxyUrl) {
     if (/^https?:\/\//i.test(proxyUrl)) {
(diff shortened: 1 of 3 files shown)


lib/install.js
```diff
@@ -443,10 +443,16 @@ function download (gyp, env, url) {
               || env.http_proxy
               || env.HTTP_PROXY
               || env.npm_config_proxy
+  
+  var strictSsl = true;
+  if (!gyp.opts.strictssl) {
+    strictSsl = gyp.opts.strictssl
+  }
```
bnoordhuis: That always disables SSL verification unless you explicitly pass `--strictssl=true`, doesn't it..?

rvagg:
closing due to age, feel free to reopen against the latest master if you still want to pursue this

PR #1120, opened 2017-02-15
ninja: Change VS variable to GYP variable

refack:
Causes ninja build on windows to fail


refack:
Discussion moved to #1121 anyway

Issue #1118, opened 2017-02-15
Don't include common.gypi from node source tree


bnoordhuis:
node-gyp includes `$nodedir/common.gypi` when creating build files for an add-on.

Per https://github.com/nodejs/CTC/issues/62#issuecomment-280108504, that probably isn't future-proof and it's a bad idea in general: it makes add-ons pick up compiler and linker settings that aren't intended for them and may be actively harmful.  People have filed bug reports about it.

Plan of action:

1. Merge `common.gypi` into `addon.gypi`.
2. Strip everything that isn't needed by add-ons.
3. Test the bejebus out of it with citgm.

murgatroid99:
I would support this change. I got burned by the fact that `common.gypi` explicitly disables `-Werror`

bnoordhuis:
Yep, but that flows on down from V8.  We'd still carry that change with a separate common.gypi.

rvagg:
still planning on making this happen @bnoordhuis? I fear this is just going to linger as a perpetually open issue unless you make it happen, I suspect there's too many dragons for most mortals to want to attempt this

bnoordhuis:
@rvagg It's one of those projects that I do want to get around to someday but I don't know when. If you want to close it, that's fine by me.

Good job on the cleanup action, by the way. :-)

rvagg:
I think I got throttled along the way with my cleanup, I stopped being able to close issues, a red error box from GitHub as if I was blocked. Seems to be working again now.

I don't mind this being left open if it's genuinely a TODO, your call. I'm just aiming to get rid of the clutter so this starts being manageable again.

PR #1121, opened 2017-02-15
fix: Making ninja output on windows to work

refack:


bnoordhuis:
I'm a little confused which PR I should review, this one or #1120?

Also, cc @kunalspathak.

0005-ninja-make-ninja-work-on-windows-again.patch
```diff
@@ -0,0 +1,38 @@
+From 06cea25b3af096b7fe82c25cc06ecba808ba9765 Mon Sep 17 00:00:00 2001
```
kunalspathak: We don't need this file i believe?

refack:
@bnoordhuis This one is more complete. But requires a change in `GYP`, so I understand any pushback.
#1120 should be a no brainer ;)

0005-ninja-make-ninja-work-on-windows-again.patch
refack: We're not doing that anymore?

kunalspathak: never mind, i thought this was added accidentally.

addon.gypi
```diff
@@ -109,7 +109,7 @@
           '-luuid.lib',
           '-lodbc32.lib',
           '-lDelayImp.lib',
-          '-l"<(node_root_dir)/$(ConfigurationName)/<(node_lib_file)"'
+          '-l<(node_root_dir)/<(CONFIGURATION_NAME)/<(node_lib_file)'
```
kunalspathak: won't this stop working for `MSBuild` ? Where is `(CONFIGURATION_NAME) defined ?

refack: <(XXXXX) is `GYP` variable resolution (i.e. `GYP` will resolve this)

kunalspathak: Sorry, may be i was not clear. What i mean is for `MsBuild`, this variable is declared [here](https://github.com/nodejs/node-gyp/blob/master/gyp/pylib/gyp/MSVSSettings.py#L426) which makes it available here as well as at other places. I am wondering, with this change if you build using `MsBuild`, what value will this variable resolve to?

refack: The process requires that `GYP(CONFIGURATION_NAME) === MsBuild(ConfigurationName)`
So now this resolution is a serpentius way, `GYP` sets it for `MsBuild`, and here `MsBuild` resolves it. And obviously it breaks for `ninja`/`win32`.

refack: Also #1118

rvagg:
closing due to age, hopefully this is taken up in refack/GYP and we can pull it in there?

Issue #1119, opened 2017-02-15
Two new commit patch `GYP` code but are missing .patch file


refack:
ec5fc36a80ea1e98eac1f490545ccb4e71fce846
a04ea3051a60bbfdabda3bafef213028cc099f6a

bnoordhuis:
I think I'll remove the .patch files, they just duplicate what's already in git.  Thanks for bringing it up though.

refack:
What if you want to rebase our changes onto a new `GYP` version?

PR #1122, opened 2017-02-16
Delete superfluous .patch files.

bnoordhuis:
The patch files are unnecessary, they also exist as commits in the
history.  Delete the files.

Refs: https://github.com/nodejs/node-gyp/issues/1119


refack:
@bnoordhuis but they make rebasing easier... 😞

bnoordhuis:
Why?  You can just `git cherry-pick` the commits after an update.

refack:
Makes it more error prone. Are those commits "sterile"?
I kinda liked this approach and the statement in makes, vendor in `GYP` as-is and in parallel version track our patches...

bnoordhuis:
I don't know what you mean by 'sterile.'  I also don't understand why you think it's more error prone.  Can you elaborate?

refack:
@bnoordhuis "Sterile" as in including only patches tn `GYP` code, and are they clearly marked.
Error prone, as in "less obvious to find needed steps" and "easier to miss one".

Assuming the procedure for a `GYP` update now is:
1. Copy all `GYP` code (except tests)
2. Apply patches
3. Maybe fix some conflicts, and update `.patch` files
4. Commit

New process will be:
1. Same as above
2. Find all _relevant_ commits that patch `GYP`. Which are relevant? maybe some are stale? what's the cut off date? Since last `GYP` Update?
3. Cherry pick those. Making sure not to miss any.
4. Maybe resolving conflicts.
5. Commit.

Both obviously works, IMHO 2nd one more fragile.

bnoordhuis:
I don't think it's all that different.  `git log gyp/` shows what patches we float and we have to review them after an upgrade anyway.  It's the same process we use in nodejs/node for V8.

All of this may be academic though because upstream gyp is clearly in maintenance mode.  I don't expect any big updates before upstream abandons it completely.

refack:
Cool 👍

Issue #1124, opened 2017-02-20
Be a good citizen on Windows and use appropriate directories.


MicahZoltu:
TL;DR: On Windows, node-gyp should install into `%LOCALAPPDATA%/<vendor>/<product>`.

In Linux, the accepted location to store :allthethings: is the home directory (`~`).  On Windows, this is *not* appropriate and almost nothing should go into the root of the `USERPROFILE` directory or `HOME` directory.  At the moment, node-gyp installs into `~` on Windows 10.  It should instead install to somewhere more appropriate such as:
* `%APPDATA%/<vendor>/<product>` for small user config of a particular product.  Note, this is synced over a network/internet so nothing in here should be particularly large.
* `%LOCALAPPDATA%/<vendor>/<product>` for large user config/extensions/per-user app installs/per-user caching.  This folder is not synced over a network/internet so it can be big.  Also, it is often redirected to a larger slower drive.
* `%PROGRAMDATA%/<vendor>/<product>` for local (not synced over network) multi-user configuration data.  This data is more permissive in that any user can read/write to it unlike `%PROGRAMFILES%` which is read only for everyone but admins.  Care should be taken with regards to security any executable written here.
* `%PROGRAMFILES%/<vendor>/<product>` for multi-user application data.  Only admins can write here, but everyone can read.  Good place to put installed software.

For the case of `node_gyp`, I assume that `%LOCALAPPDATA%/<vendor>/<product>` or `%PROGRAMFILES%/<vendor>/<product>` is the appropriate location, depending on whether there is an expectation of admin rights at install time.

Note: None of the above folders are `%USERPROFILE%` because the only thing that should use that folder are save dialogs.  😄

bnoordhuis:
Any change in the defaults is probably automatically semver-major.  That means that even if node-gyp makes the switch, it probably won't end up on user's systems for a long time.

So far no one has complained either so I'd be inclined to leave well enough alone.  Changes to the defaults are disruptive to users.  Unless there are strongly compelling reasons it's probably not worth doing.

MicahZoltu:
> So far no one has complained either so I'd be inclined to leave well enough alone

[...]

PR #1125, opened 2017-02-20
Improves Windows support.

MicahZoltu:
This is a naive fix for #1124.  On Windows, files should not be stored directly in `%USERPROFILE%` for a number of reasons.  Instead, store the files in an appropriate location on Windows such as `%LOCALAPPDATA%/<vendor>/<product>`.

### DIFF
--- a/bin/node-gyp.js
+++ b/bin/node-gyp.js
@@ -25,9 +25,8 @@ prog.parseArgv(process.argv)
 prog.devDir = prog.opts.devdir
 
 var homeDir = osenv.home()
-if (process.env.LOCALAPPDATA) {
-  // TODO: create intermediate directories
-  pro.devDir = path.resolve(process.env.LOCALAPPDATA, 'nodejs', 'node-gyp')
+if (process.env.LOCALAPPDATA && process.platform === 'win32') {
+  prog.devDir = path.resolve(process.env.LOCALAPPDATA, 'nodejs', 'node-gyp')
 } else if (prog.devDir) {
   prog.devDir = prog.devDir.replace(/^~/, homeDir)
 } else if (homeDir) {
--- a/bin/node-gyp.js
+++ b/bin/node-gyp.js
@@ -25,14 +25,18 @@ prog.parseArgv(process.argv)
 prog.devDir = prog.opts.devdir
 
 var homeDir = osenv.home()
-if (prog.devDir) {
+if (process.env.LOCALAPPDATA) {
+  // TODO: create intermediate directories
+  pro.devDir = path.resolve(process.env.LOCALAPPDATA, 'nodejs', 'node-gyp')
+} else if (prog.devDir) {
   prog.devDir = prog.devDir.replace(/^~/, homeDir)
 } else if (homeDir) {
   prog.devDir = path.resolve(homeDir, '.node-gyp')
 } else {
   throw new Error(
     "node-gyp requires that the user's home directory is specified " +
-    "in either of the environmental variables HOME or USERPROFILE. " +
+    "in either of the environmental variables HOME or USERPROFILE, " +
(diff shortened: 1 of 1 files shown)


bin/node-gyp.js
```diff
@@ -25,14 +25,18 @@ prog.parseArgv(process.argv)
 prog.devDir = prog.opts.devdir
 
 var homeDir = osenv.home()
-if (prog.devDir) {
+if (process.env.LOCALAPPDATA) {
+  // TODO: create intermediate directories
```
MicahZoltu: What is the best way to do this in node without any dependencies?  I found http://stackoverflow.com/a/40249973/879046, but I'm not sure if that is appropriate in this case or not.

bin/node-gyp.js
```diff
@@ -25,14 +25,18 @@ prog.parseArgv(process.argv)
...
+  pro.devDir = path.resolve(process.env.LOCALAPPDATA, 'nodejs', 'node-gyp')
```
MicahZoltu: Does `path.resolve` add the directory separators for me?  The documentation is a bit vague on this.

bin/node-gyp.js
MicahZoltu: Fixed and removed comment.

bin/node-gyp.js
MicahZoltu: Fixed.

bin/node-gyp.js
```diff
@@ -25,14 +25,17 @@ prog.parseArgv(process.argv)
 prog.devDir = prog.opts.devdir
 
 var homeDir = osenv.home()
-if (prog.devDir) {
+if (process.env.LOCALAPPDATA && process.platform === 'win32') {
+  prog.devDir = path.resolve(process.env.LOCALAPPDATA, 'nodejs', 'node-gyp')
+} else if (prog.devDir) {
```
joaocgreis: `prog.devDev` should still be the first one, so that the command line parameter `--devdir` has effect.
```
if (prog.devDir) {
  ...
} else if (process.env.LOCALAPPDATA && process.platform === 'win32') {
  ...
```

joaocgreis:
@MicahZoltu thanks for submitting this, I think this would be good to have. Your comment at https://github.com/nodejs/node-gyp/issues/1124#issuecomment-281478660 looks about right to me if you want to move forward with it and include it here.

digitalinfinity:
cc @kunalspathak for awareness in case changes are needed in node-chakracore from this change

MicahZoltu:
Sorry for the delay on getting back to this.

What is the proper way to do File IO in this context?  It doesn't appear that patterns are in place to do async IO, which is normally how I would go about doing any kind of disk operations.

refack:
@MicahZoltu are you interested in following up on this?
IMHO should move to `%TMP%` tough.

MicahZoltu:
I want it to get done, but I would be super happy if someone else took it over.  Are you volunteering?  ;)

I'm currently not sure what the right way to do file-io in this area of code is.  Normally I would expect full async but this code isn't setup to be async and changing the pattern feels wrong when I'm trying to do a minor change.

refack:
> I want it to get done, but I would be super happy if someone else took it over. Are you volunteering? ;)
> 
> I'm currently not sure what the right way to do file-io in this area of code is. Normally I would expect full async but this code isn't setup to be async and changing the pattern feels wrong when I'm trying to do a minor change.

I'll take over no-problem (we usually give the OP first dibs).
IMHO with the code downstream handling all the "real" io, just needs to give precedence to `prog.devDir` everything will fall into place.

rvagg:
closing due to age, anyone's welcome to pick this up again against the lastest master and open a new PR

PR #1130, opened 2017-02-25
win: find and setup for VS2017

joaocgreis:
Reviewed https://github.com/nodejs/node-gyp/pull/1103 :
- Filter VS2017 installations using the installed components, to make sure the installation found is usable.
- Added logging.

cc @refack @bzoz can you review?

### DIFF
--- a/lib/Find-VS2017.cs
+++ b/lib/Find-VS2017.cs
@@ -1,0 +1,262 @@
+// powershell -ExecutionPolicy Unrestricted -Version "2.0" -Command "&{Add-Type -Path Find-VS2017.cs; [VisualStudioConfiguration.Main]::Query()}"
+using System;
+using System.Text;
+using System.Runtime.InteropServices;
+
+namespace VisualStudioConfiguration
+{
+    [Flags]
+    public enum InstanceState : uint
+    {
+        None = 0,
+        Local = 1,
+        Registered = 2,
+        NoRebootRequired = 4,
+        NoErrors = 8,
+        Complete = 4294967295,
+    }
+
+    [Guid("6380BCFF-41D3-4B2E-8B2E-BF8A6810C848")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [ComImport]
+    public interface IEnumSetupInstances
+    {
+
+        void Next([MarshalAs(UnmanagedType.U4), In] int celt,
+            [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.Interface), Out] ISetupInstance[] rgelt,
+            [MarshalAs(UnmanagedType.U4)] out int pceltFetched);
+
+        void Skip([MarshalAs(UnmanagedType.U4), In] int celt);
+
+        void Reset();
+
+        [return: MarshalAs(UnmanagedType.Interface)]
+        IEnumSetupInstances Clone();
+    }
+
+    [Guid("42843719-DB4C-46C2-8E7C-64F1816EFD5B")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [ComImport]
+    public interface ISetupConfiguration
+    {
+    }
+
+    [Guid("26AAB78C-4A60-49D6-AF3B-3C35BC93365D")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [ComImport]
+    public interface ISetupConfiguration2 : ISetupConfiguration
+    {
+
+        [return: MarshalAs(UnmanagedType.Interface)]
+        IEnumSetupInstances EnumInstances();
+
+        [return: MarshalAs(UnmanagedType.Interface)]
+        ISetupInstance GetInstanceForCurrentProcess();
+
+        [return: MarshalAs(UnmanagedType.Interface)]
+        ISetupInstance GetInstanceForPath([MarshalAs(UnmanagedType.LPWStr), In] string path);
+
+        [return: MarshalAs(UnmanagedType.Interface)]
+        IEnumSetupInstances EnumAllInstances();
+    }
+
+    [Guid("B41463C3-8866-43B5-BC33-2B0676F7F42E")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [ComImport]
+    public interface ISetupInstance
+    {
+    }
+
+    [Guid("89143C9A-05AF-49B0-B717-72E218A2185C")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [ComImport]
+    public interface ISetupInstance2 : ISetupInstance
+    {
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetInstanceId();
+
+        [return: MarshalAs(UnmanagedType.Struct)]
+        System.Runtime.InteropServices.ComTypes.FILETIME GetInstallDate();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetInstallationName();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetInstallationPath();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetInstallationVersion();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetDisplayName([MarshalAs(UnmanagedType.U4), In] int lcid);
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetDescription([MarshalAs(UnmanagedType.U4), In] int lcid);
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string ResolvePath([MarshalAs(UnmanagedType.LPWStr), In] string pwszRelativePath);
+
+        [return: MarshalAs(UnmanagedType.U4)]
+        InstanceState GetState();
+
+        [return: MarshalAs(UnmanagedType.SafeArray, SafeArraySubType = VarEnum.VT_UNKNOWN)]
+        ISetupPackageReference[] GetPackages();
+
+        ISetupPackageReference GetProduct();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetProductPath();
+
+        [return: MarshalAs(UnmanagedType.VariantBool)]
+        bool IsLaunchable();
+
+        [return: MarshalAs(UnmanagedType.VariantBool)]
+        bool IsComplete();
+
+        ISetupPropertyStore GetProperties();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetEnginePath();
+    }
+
+    [Guid("DA8D8A16-B2B6-4487-A2F1-594CCCCD6BF5")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [ComImport]
+    public interface ISetupPackageReference
+    {
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetId();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetVersion();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetChip();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetLanguage();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetBranch();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetType();
+
+        [return: MarshalAs(UnmanagedType.BStr)]
+        string GetUniqueId();
+
+        [return: MarshalAs(UnmanagedType.VariantBool)]
+        bool GetIsExtension();
+    }
+
+    [Guid("c601c175-a3be-44bc-91f6-4568d230fc83")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    [ComImport]
+    public interface ISetupPropertyStore
+    {
+
+        [return: MarshalAs(UnmanagedType.SafeArray, SafeArraySubType = VarEnum.VT_BSTR)]
+        string[] GetNames();
+
+        object GetValue([MarshalAs(UnmanagedType.LPWStr), In] string pwszName);
+    }
+
+    [Guid("42843719-DB4C-46C2-8E7C-64F1816EFD5B")]
+    [CoClass(typeof(SetupConfigurationClass))]
+    [ComImport]
+    public interface SetupConfiguration : ISetupConfiguration2, ISetupConfiguration
+    {
+    }
+
+    [Guid("177F0C4A-1CD3-4DE7-A32C-71DBBB9FA36D")]
+    [ClassInterface(ClassInterfaceType.None)]
+    [ComImport]
+    public class SetupConfigurationClass
+    {
+    }
+
+    public static class Main
+    {
+        public static void Query()
+        {
+            ISetupConfiguration query = new SetupConfiguration();
+            ISetupConfiguration2 query2 = (ISetupConfiguration2)query;
+            IEnumSetupInstances e = query2.EnumAllInstances();
+
+            int pceltFetched;
+            ISetupInstance2[] rgelt = new ISetupInstance2[1];
+            StringBuilder log = new StringBuilder();
+            while (true)
+            {
+                e.Next(1, rgelt, out pceltFetched);
+                if (pceltFetched <= 0)
+                {
+                    Console.WriteLine(String.Format("{{\"log\":\"{0}\"}}", log.ToString()));
+                    return;
+                }
+                if (CheckInstance(rgelt[0], ref log))
+                    return;
+            }
+        }
+
+        private static bool CheckInstance(ISetupInstance2 setupInstance2, ref StringBuilder log)
+        {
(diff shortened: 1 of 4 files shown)


refack:
Personally I would move logic from the C# to JS. I like that you use more of the information available to us from the COM, but I would filter and choose in JS. I also like the idea of a log, but maybe if the C# prints everything, there will be no need.
I'll take some ideas to #1103 and to [windows-autoconf](https://github.com/node4good/windows-autoconf) 😉

joaocgreis:
Updated, fixed failure when there's no usable installation and store the MSBuild executable path in config so that it can respect `--msvs_version`. I considered filtering on the JS side, but it can be a huge list of components and is easy enough to do on the C# side. We can always improve if necessary.

Tested (with node v7 on Windows 7):
- Several installations of VS2017, where only one is usable
- VS2017 installed but not usable
- After uninstalling VS2017
- VSBT2015 and VS2017 side by side (`--msvs_version` works)
- Run npm as a normal user
- On Windows 10, node v4, running as a normal user

bnoordhuis:
Is this PR taking a dependency on powershell and the C# compiler?  That... seems unfortunate.

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/commit/b456dd0d46a143c3e6afd63116124800f3c1c18f

@bnoordhuis it does, but those two are installed by default everywhere I've looked. To make node-gyp work out of the box in the platforms we support, it's either this, `comtypes` as in https://github.com/nodejs/node-gyp/pull/1101, or have an [executable](https://github.com/nodejs/node-gyp/pull/1103#issuecomment-282196116) committed. This worked great everywhere I've tested, but it's not impossible that there is some uncommon Windows configuration that breaks. We can fix those based on actual issues, or decide for any of the other options.

bnoordhuis:
Merging is a little premature.  There has been no sign-off from a committer and you didn't run the CI either.

refack:
LGTM(ish) I'll make another PR with some more changes 😉

joaocgreis:
Took it back, I took @bzoz sign-off as enough, this has been under discussion in https://github.com/nodejs/node-gyp/pull/1101 and https://github.com/nodejs/node-gyp/pull/1103 for a long time. I'll run CI.

lib/build.js
```diff
@@ -14,7 +14,9 @@ var fs = require('graceful-fs')
   , mkdirp = require('mkdirp')
   , exec = require('child_process').exec
   , processRelease = require('./process-release')
-  , win = process.platform == 'win32'
+  , win = process.platform === 'win32'
+if (win)
```
refack: You don't need this if you already injected `config.variables.msbuild_path` (which is smart)

joaocgreis:
CI:
- v.7.6.0: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/39/
- v6.10.0: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/40/
- v4.8.0: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/41/

lib/build.js
joaocgreis: Thanks!

joaocgreis:
@bnoordhuis CI is green, do you want to review? I think this is important to have, what should I do to move this forward?

joaocgreis:
cc @nodejs/node-gyp @nodejs/build

joaocgreis:
Updated: changed the MSBuild dependency component. Installing only MSBuild does not make it usable for C++ if the workload is not installed. But installing "Visual Studio C++ core features" is enough, included MSBuild.exe and the necessary files to make it usable.

Fishrock123:
@bnoordhuis, @rvagg PTAL

rvagg:
working on this now FYI, had to get a few other pressing things out of the way but I'm reviewing and testing this out for myself now

lib/configure.js
```diff
@@ -137,6 +139,18 @@ function configure (gyp, argv, callback) {
     // disable -T "thin" static archives by default
     variables.standalone_static_library = gyp.opts.thin ? 0 : 1
 
+    if (win && !(gyp.opts.msvs_version && gyp.opts.msvs_version !== '2017')) {
```
rvagg: would you mind inverting this? reads really badly. `(win && (!gyp.opts.msvs_version || gyp.opts.msvs_version === '2017'))`

lib/configure.js
```diff
@@ -137,6 +139,18 @@ function configure (gyp, argv, callback) {
...
+      const vsSetup = findVS2017()
+      if (vsSetup) {
+        gyp.opts.msvs_version = '2015'
```
rvagg: are we forcing this to make GYP happy, or is this (and the next line) a mistake?

lib/find-vs2017.js
```diff
@@ -0,0 +1,47 @@
...
+
+  hasCache = true
+
+  const ps = 'PowerShell -ExecutionPolicy Unrestricted -Command '
+  const csFile = path.join(__dirname, 'Find-VS2017.cs')
+  const psQuery = ps + '"&{Add-Type -Path \'' + csFile + '\'; [VisualStudioConfiguration.Main]::Query()}" 2>&1'
+
+  var vsSetup
+  try {
+    const vsSetupRaw = execSync(psQuery, { encoding: 'utf8' })
```
rvagg: `var` for consistency, also given that node-gyp still works for very old versions of Node

rvagg: sorry, I mean, `s/var/const` in this entire file

lib/configure.js
```diff
@@ -137,6 +139,18 @@ function configure (gyp, argv, callback) {
     // disable -T "thin" static archives by default
     variables.standalone_static_library = gyp.opts.thin ? 0 : 1
 
+    if (win && !(gyp.opts.msvs_version && gyp.opts.msvs_version !== '2017')) {
+      const vsSetup = findVS2017()
```
rvagg: `s/const/var`

lib/Find-VS2017.cs
```diff
@@ -0,0 +1,262 @@
+// powershell -ExecutionPolicy Unrestricted -Version "2.0" -Command "&{Add-Type -Path Find-VS2017.cs; [VisualStudioConfiguration.Main]::Query()}"
```
rvagg: does this file need source attribution and/or copyright? or is it ours?

lib/configure.js
joaocgreis: We are forcing this to make GYP happy, and to avoid pulling https://chromium-review.googlesource.com/#/c/433540/ (which did not land yet) as a floating patch, since floating patches seem to be avoided here.

lib/Find-VS2017.cs
joaocgreis: This was submitted by @refack in https://github.com/nodejs/node-gyp/pull/1103 , I believe it was adapted from the python COM API by @bzoz in https://github.com/nodejs/node-gyp/pull/1101 , which in turn got the GUIDs and API from the nuget package upstream. @refack can you confim?

joaocgreis:
@rvagg updated, PTAL. About the 2015->2017 GYP hack, my comment above is hidden by GitHub:

> We are forcing this to make GYP happy, and to avoid pulling https://chromium-review.googlesource.com/#/c/433540/ (which did not land yet) as a floating patch, since floating patches seem to be avoided here.

rvagg:
@joaocgreis would you mind putting that comment about the gyp hack into the code so we have record of it? I can see someone coming along to "fix" it later without that context. I could even imagine myself doing that!

joaocgreis:
Updated with the comment.

@refack can you confirm that you wrote the `.cs` file (or let us know where you got it from)? Thanks!

lib/configure.js
```diff
@@ -137,6 +139,21 @@ function configure (gyp, argv, callback) {
...
+      if (vsSetup) {
+        // GYP doesn't (yet) have support for VS2017, so we force it to VS2015
+        // to avoid pulling a floating patch that has not landed upstream.
+        // Ref: https://chromium-review.googlesource.com/#/c/433540/
+        gyp.opts.msvs_version = '2015'
+        process.env['GYP_MSVS_VERSION'] = 2015
+        process.env['GYP_MSVS_OVERRIDE_PATH'] = vsSetup.path
+        defaults['msbuild_toolset'] = 'v141'
+        defaults['msvs_windows_target_platform_version'] = vsSetup.sdk
+        variables['msbuild_path'] = path.join(vsSetup.path, 'MSBuild', '15.0', 'Bin', 'MSBuild.exe')
```
bnoordhuis: Can you wrap this at 80 columns?

lib/find-vs2017.js
```diff
@@ -0,0 +1,47 @@
...
+
+function findVS2017() {
+  if (hasCache)
+    return cache
+
+  hasCache = true
+
+  var ps = 'PowerShell -ExecutionPolicy Unrestricted -Command '
+  var csFile = path.join(__dirname, 'Find-VS2017.cs')
+  var psQuery = ps + '"&{Add-Type -Path \'' + csFile + '\'; [VisualStudioConfiguration.Main]::Query()}" 2>&1'
```
bnoordhuis: Likewise.

lib/find-vs2017.js
```diff
@@ -0,0 +1,47 @@
...
+
+  var vsSetup
+  try {
+    var vsSetupRaw = execSync(psQuery, { encoding: 'utf8' })
```
bnoordhuis: Policy decision: do we still support v0.10?  If yes, you can't use execSync(), it was added in v0.12.

That aside, is there a reason to use execSync() instead of spawnSync()?  If `__dirname` has spaces in it, execSync() will probably fail.

Lastly, can powershell always be expected to be on the path?

joaocgreis: I kept the sync version from the original PR because making it async would have required deeper changes in `build.js` and `configure.js`. This is no longer true after the change to how MSBuild is found, so I'll change to async.

I made sure this works when node-gyp is installed to a location with spaces in its path.

Powershell is always in the path, unless is has been tampered with. The location of the binary seems to be stable, I'll change this to use it directly to avoid issues.

lib/Find-VS2017.cs
refack: I wrote it myself. GUIDs extracted from [interop](https://www.nuget.org/packages/Microsoft.VisualStudio.Setup.Configuration.Interop/)

joaocgreis:
@bnoordhuis updated, PTAL. This now works with v0.10.

Tested, with node_gyp replaced in a node installation under `Program Files` with versions v0.10.48, v0.12.18, v4.8.0, v6.10.0 and v7.7.2. Also tested without a usable VS2017 installation and without any VS installed.

lib/Find-VS2017.cs
refack: Please add
```
// Copyright 2017 - Refael Ackermann
// Distributed under MIT style license
// See accompanying file LICENSE at https://github.com/node4good/windows-autoconf
```

refack:
Please see https://github.com/nodejs/node-gyp/pull/1130/files/034a33078ce30007b74ec7b94156d4270b67e1c4#r106001050

lib/Find-VS2017.cs
refack: I recommend using updated file https://github.com/node4good/windows-autoconf/blob/master/tools/GetVS2017Configuration.cs

refack:
I recommand using updated file https://github.com/node4good/windows-autoconf/blob/master/tools/GetVS2017Configuration.cs
Same as one that landed in [`boost/build`](https://github.com/boostorg/build/blob/develop/src/tools/vc141helper/GetVS2017Configuration.cs)
And gets proper MS style CI at [AppVeyor](https://ci.appveyor.com/project/refack/windows-autoconf), which BTW runs `node-gyp` test suite

joaocgreis:
Updated the COM API part of the file. I'd rather keep the code part as is here, with the filtering and the logging.

Started CI:
- v7.7.2: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/42/ (green)
- v0.10.48: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/43/ (red on ubuntu, confirmed the same happens on master: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/44/)

@bnoordhuis should we run other versions?

bnoordhuis:
Maybe do v4.x and v6.x just in case?

joaocgreis:
- v6.10.0: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/45/ (green)
- v4.8.0: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/46/ (green)
- v0.12.18: https://ci.nodejs.org/view/All/job/nodegyp-test-commit/47/ (green)

rvagg:
@joaocgreis let's get this published today. Clean it up and get it merged and we'll publish and then PR npm to get it there too. Ping me on IRC when you're around and want to move forward.

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/commit/ae141e19060ecb1c596a96acf9510d66b7376c18

rvagg:
Attempted PRs to npm (I always get seem to get this process wrong):

* lts: https://github.com/npm/npm/pull/16049
* latest: https://github.com/npm/npm/pull/16050

rvagg:
OK, looks like we might not be getting v4 compatibility for VS 2017 from npm unless we patch it ourselves I guess @nodejs/lts https://github.com/npm/npm/pull/16049

refack:
What's needed for v4 compat? Can I help?

gibfahn:
@refack issue is that Node 4 is going into maintenance mode soon, so there might not be another update to npm before that time. The ping to @nodejs/lts is for the LTS team to decide whether it's worth doing another v4 release to pick up the npm update.

FWIW, I think it's worth updating to support VS2017 in Node 4 (but then I won't have to do the release).

refack:
@gibfahn, thanks
Best I can do is make sure VS2017 changes to `node-gyp` are fully compatible with v4 (I'll add it to my CI matrix)

Fishrock123:
> OK, looks like we might not be getting v4 compatibility for VS 2017 from npm unless we patch it ourselves I guess

I'm not sure that was read from the comment correctly but it *could* be the case.

gibfahn:
>Unless Node.js is planning on doing a new v4 release for this(?), I don't see us doing a new npm@2 release.

Based on that comment, if we can get it into Node v4, we can get it into npm. 

Node v4 goes into maintenance mode on 2017-04-01, i.e. in ten days. Whichever way we go, we need to make a decision ASAP.

refack:
👎 
#1151 `v3.6.0` might break a lot of `actions` out there 😟

refack:
@joaocgreis, @rvagg Let's fast track #1153

refack:
**"Bring back VS150COMNTOOLS"**
I urge everyone else to up vote at https://developercommunity.visualstudio.com/content/problem/35325/missing-vs150comntools.html

lib/Find-VS2017.cs
```diff
@@ -0,0 +1,268 @@
...
+            {
+                const string Win10SDKPrefix = "Microsoft.VisualStudio.Component.Windows10SDK.";
+
+                string id = package.GetId();
+                if (id == "Microsoft.VisualStudio.VC.MSBuild.Base")
+                    hasMSBuild = true;
+                else if (id == "Microsoft.VisualStudio.Component.VC.Tools.x86.x64")
+                    hasVCTools = true;
+                else if (id.StartsWith(Win10SDKPrefix))
+                    Win10SDKVer = Math.Max(Win10SDKVer, UInt32.Parse(id.Substring(Win10SDKPrefix.Length)));
```
refack: Bug found with new SDK #1144

AndrewPardoe:
The officially supported vswhere tool is now available as part of the install starting today with Visual Studio 15.2 preview 2: https://blogs.msdn.microsoft.com/heaths/2017/04/21/vswhere-is-now-installed-with-visual-studio-2017/. This doesn't fix all the existing installs of VS 2017 but going forward you can now detect the VS install locations and whether C++ tools are available from a scripted environment. I hope this helps your scenario.

am11:
One approach is to download VSWhere as a fallback if not available and completely rely on it to find VS2017:

```powershell
# powershell
$vswherePath = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe";
if (!(Test-Path $vswherePath)) {
    $latestReleaseTag = $(curl https://api.github.com/repos/Microsoft/vswhere/releases/latest | `
                          ConvertFrom-Json) | % { $_.tag_name };

    $vswherePath = "$pwd\vswhere.exe";
    curl -O $vswherePath `
        https://github.com/Microsoft/vswhere/releases/download/$latestReleaseTag/vswhere.exe;
}

$vsBase = &$vswherePath -latest -products * `
              -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 `
              -property installationPath;

if (!$vsBase) {
    throw "VS or MSVC component is not installed";
}

&"$vsBase\vc\Auxiliary\Build\vcvars64.bat";
# now link.exe, cl.exe etc. are in PATH

# or write vsvars path to stdout for the caller (MSBuild Exec task)
echo $vsBase\vc\Auxiliary\Build\vcvars64.bat;
```

However, in practice, vswhere is available in a deterministic location when we have VS or buildtools installed, then downloading is sort of moot if we don't want to support non-ProgramFiles installs (e.g. https://github.com/dotnet/corert/blob/378ad6a/src/BuildIntegration/findvcvarsall.bat).

ElectronicLove:
So please can you help me did you find out how to solve this problem? I've tried everything still this error.

PR #1158, opened 2017-03-27
Consider using micromatch

doowb:
Hi,

This PR is doing two things:

- Replace minimatch with [micromatch](https://www.npmjs.com/package/micromatch).
- Update the glob pattern for matching valid header files so only one pass is made.

When the glob pattern is updated to `'*.{h,gypi}'`, `micromatch` will create one regexp pattern that will only do one pass over the file being checked. This differs from `minimatch` that creates 2 different regexp patterns and will check the file twice if the pattern doesn't match `'*.h'`.

Thanks for considering this.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -18,7 +18,6 @@ var fs = require('graceful-fs')
   , log = require('npmlog')
   , semver = require('semver')
   , request = require('request')
-  , micromatch = require('micromatch')
   , mkdir = require('mkdirp')
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Thanks for the PR.  The reason we use minimatch is because npm does, meaning we can share a copy with it and make the installer a little smaller.

I wouldn't object to removing minimatch altogether if size of the standalone install is your motivation for doing this.  It's not like we are doing anything fancy with it; a simple regex probably suffices.

doowb:
> if size of the standalone install is your motivation for doing this

No, it was more that I use micromatch often and I recognized a quick optimization that may be useful since almost any time I do an `npm install` I see something using `node-gyp`.

> a simple regex probably suffices

It probably does. I can update this PR with a regex change if you'd like. I know that micromatch expands the glob pattern into a more thorough regex than I would probably think of and there may be something additional done because of the `matchBase: true` option, but I'll look into it.

doowb:
@bnoordhuis I updated this to just use `path.extname` since the patterns are just checking for the file extensions.

bnoordhuis:
Thanks Brian, landed in d70513a.

PR #1162, opened 2017-04-03
FIX bug, Add strictSSL(Boolean) option

dkhjmu:
Added strictSSL sption


rvagg:
if you'd still like to pursue this could you explain the bug and use-case and also fix up your code spacing to match the rest of the code please?

PR #1167, opened 2017-04-07
Generator: msvs: compile .mm files

jracle:
Hi guys,

  in a native node extension I want to compile .mm file with C/C++ compiler on Windows.

It does not appear in vcxproj in CLCompile group. I did the modification to fix it.

Cheers

### DIFF
--- a/gyp/pylib/gyp/generator/msvs.py
+++ b/gyp/pylib/gyp/generator/msvs.py
@@ -2120,7 +2120,7 @@ def _MapFileToMsBuildSourceType(source, rule_dependencies,
   if ext in extension_to_rule_name:
     group = 'rule'
     element = extension_to_rule_name[ext]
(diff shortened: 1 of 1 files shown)


bnoordhuis:
.mm is normally used for Objective-C++ files, isn't it?  How would that work exactly?

jracle:
Yes indeed, this file compiles as Objective-C++ on mac, and as C++ on windows.
I happen in several projects to need to put .mm extension so as clang not to crash if I import an objective-C header. This is BTW consistent with the rest of the file I modified, see:

```
def _ValidateSourcesForMSVSProject(spec, version):
    ....
    is_compiled_file = ext in [
        '.c', '.cc', '.cpp', '.cxx', '.m', '.mm', '.s', '.S']
```

which already considers '.mm' as a valid extension for a vcxproj.

bnoordhuis:
Landed in 206309e, thanks.

PR #1171, opened 2017-04-11
Replace tape with tap

### LINKED ISSUE #1123: Stop using tape for tests
Tape runs all tests in the same process.  There doesn't seem to be a way to request process-per-test isolation.

Monkey-patching in tests ends up affecting other tests.  I'm quite sure at least one or two bugs (hopefully only in tests) went unnoticed because of it so I propose we move away from tape.

NatalieWolfe:
The [tap](https://www.npmjs.com/package/tap) module is a more fully featured TAP framework that executes its tests in separate child processes.

Resolves #1123

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,12 +40,12 @@
     "node": ">= 0.8.0"
   },
   "devDependencies": {
-    "tape": "~4.2.0",
     "bindings": "~1.2.1",
     "nan": "^2.0.0",
-    "require-inject": "~1.3.0"
+    "require-inject": "~1.3.0",
+    "tap": "^10.3.2"
   },
   "scripts": {
-    "test": "tape test/test-*"
+    "test": "tap test/test-*"
   }
 }
--- a/test/test-addon.js
+++ b/test/test-addon.js
@@ -1,12 +1,12 @@
 'use strict'
 
-var test = require('tape')
+var tap = require('tap')
 var execFile = require('child_process').execFile
 var path = require('path')
 var addonPath = path.resolve(__dirname, 'node_modules', 'hello_world')
 var nodeGyp = path.resolve(__dirname, '..', 'bin', 'node-gyp.js')
 
-test('build simple addon', function (t) {
+tap.test('build simple addon', function (t) {
   t.plan(3)
 
   // Set the loglevel otherwise the output disappears when run via 'npm test'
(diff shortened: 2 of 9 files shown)


bnoordhuis:
Thanks, this is great!  I wonder though, does `tap.test()` actually run tests in a separate process?  I glean from node-tap's source that only `tap.spawn()` does that.

NatalieWolfe:
Correct, `tap.test` doesn't run each in their own process, but `$ tap test/test-*` will spawn a separate child process for each file that matched. Was it per-individual-test that needed process isolation?

bnoordhuis:
Yes.

NatalieWolfe:
Is it possible to move the tests which need isolation into individual files? Which tests need this? I can look into doing that step as part of this ticket.

bnoordhuis:
I'm going to say 'all of them' - it's not really true but isolation never hurts.  It's my original plan of attack for #1123.

NatalieWolfe:
Is that necessary up front? That seems like the kind of thing that could be resolved on a case-by-case basis when colliding tests are discovered. Moving individual tests into separate files isn't that hard, but seems like overkill if you don't actually need them isolated.

bnoordhuis:
Such things tend to go unnoticed, though.  I raised #1123 after I found two unrelated tests had been clobbering each other undetected for some time.

Another anecdote: a project I was involved in had a bug caused by lack of isolation that stopped half the tests from running but it went undiscovered for a year because the CI was green all the time.

package.json
```diff
@@ -40,12 +40,12 @@
...
-    "tape": "~4.2.0",
     "bindings": "~1.2.1",
     "nan": "^2.0.0",
-    "require-inject": "~1.3.0"
+    "require-inject": "~1.3.0",
+    "tap": "^10.3.2"
   },
   "scripts": {
-    "test": "tape test/test-*"
+    "test": "tap test/test-*"
```
refack: IMHO need to add `-Rtap`

refack:
@NatalieWolfe do you need help? I can help...

maclover7:
ping @NatalieWolfe

rvagg:
needs to be updated, either @NatalieWolfe needs to revisit or someone else should open a PR, I'd +1 on merging this if it was ready

cclauss:
@NatalieWolfe Is this already done elsewhere or is there more to do here?

richardlau:
This was done in https://github.com/nodejs/node-gyp/pull/1795.

PR #1174, opened 2017-04-12
Check definedness of argv[0] first

marcust:
I'm not really sure what the `&& argv[0]` is for, but if that is a safegard for `argv[0]` being defined it makes more sense to check that before `semver.valid()` is called on it.
At least I had the case where this broke because that was actually undefined.


bnoordhuis:
Looks okay to me but maybe @rvagg can weigh in in case I miss some subtlety.

marcust:
@rvagg @bnoordhuis any chance to get this merged? (Cleaning up my PRs)

maclover7:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/78/

PR #1176, opened 2017-04-14
no_proxy support

ballwood:
Ported code across from the request library, which I know to work correctly - https://github.com/request/request.

Unsure of where to put the unit/integration tests, if someone could guide me I'll write them. There were 11 tests failing when running npm test when I forked and the same amount after I made the code change.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -22,6 +22,7 @@ var fs = require('graceful-fs')
   , request = require('request')
   , minimatch = require('minimatch')
   , mkdir = require('mkdirp')
+  , url = require('url')
   , processRelease = require('./process-release')
   , win = process.platform == 'win32'
 
@@ -423,6 +424,44 @@ function install (gyp, argv, callback) {
 
 }
 
+function formatHostname (hostname) {
+  // canonicalize the hostname, so that 'oogle.com' won't match 'google.com'
+  return hostname.replace(/^\.*/, '.').toLowerCase()
+}
+
+function parseNoProxyZone (zone) {
+  zone = zone.trim().toLowerCase()
+
+  var zoneParts = zone.split(':', 2)
+  var zoneHost = formatHostname(zoneParts[0])
+  var zonePort = zoneParts[1]
+  var hasPort = zone.indexOf(':') > -1
+
+  return {hostname: zoneHost, port: zonePort, hasPort: hasPort}
+}
+
+function uriInNoProxy (uri, noProxy) {
+  var parsedUri = url.parse(uri);
+  var port = parsedUri.port || (parsedUri.protocol === 'https:' ? '443' : '80')
+  var hostname = formatHostname(parsedUri.hostname)
+  var noProxyList = noProxy.split(',')
+
+  // iterate through the noProxyList until it finds a match.
+  return noProxyList.map(parseNoProxyZone).some(function (noProxyZone) {
+    var isMatchedAt = hostname.indexOf(noProxyZone.hostname)
+    var hostnameMatched = (
+      isMatchedAt > -1 &&
+      (isMatchedAt === hostname.length - noProxyZone.hostname.length)
+    )
+
+    if (noProxyZone.hasPort) {
+      return (port === noProxyZone.port) && hostnameMatched
+    }
+
+    return hostnameMatched
+  })
+}
+
 function download (gyp, env, url) {
   log.http('GET', url)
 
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Sorry for the delay.  This needs some tests but I'm kind of surprised at how complex it still is.  Is `npm_config_no_proxy` a thing?

vbfox:
I'd really like to see that PR get merged.

For npm config exporting `npm_config_no_proxy` there is a PR open but I don't know if it'll ever get merged: https://github.com/npm/npm/pull/14248 (Currently it can't be specified in npm config)

ballwood:
Hi, sorry I thought I replied to this last week, must have not pressed send. Yeah I noticed that PR was open and included it to future proof it but their work on it seems to have gone by the wayside. 

I can have a look at the tests I'm just trying to understand them at the moment and work out something that works correctly.

ballwood:
@bnoordhuis wrote some small unit tests as I got some free time

lib/install.js
```diff
@@ -423,6 +424,44 @@ function install (gyp, argv, callback) {
 
 }
 
+function formatHostname (hostname) {
+  // canonicalize the hostname, so that 'oogle.com' won't match 'google.com'
+  return hostname.replace(/^\.*/, '.').toLowerCase()
```
bnoordhuis: This transforms `google.com` into `.google.com`.  I infer that is intentional but can you update the comment to make that more explicit?  This behavior also seems only lightly tested.

Aside: can you capitalize and punctuate comments?

lib/install.js
```diff
@@ -423,6 +424,44 @@ function install (gyp, argv, callback) {
...
+}
+
+function parseNoProxyZone (zone) {
+  zone = zone.trim().toLowerCase()
+
+  var zoneParts = zone.split(':', 2)
+  var zoneHost = formatHostname(zoneParts[0])
+  var zonePort = zoneParts[1]
+  var hasPort = zone.indexOf(':') > -1
```
bnoordhuis: Not wrong per se but can also be derived without re-scanning from `zonePort !== undefined` or `zoneParts.length === 2`.

lib/install.js
```diff
@@ -423,6 +424,44 @@ function install (gyp, argv, callback) {
...
+  var port = parsedUri.port || (parsedUri.protocol === 'https:' ? '443' : '80')
+  var hostname = formatHostname(parsedUri.hostname)
+  var noProxyList = noProxy.split(',')
+
+  // iterate through the noProxyList until it finds a match.
+  return noProxyList.map(parseNoProxyZone).some(function (noProxyZone) {
+    var isMatchedAt = hostname.indexOf(noProxyZone.hostname)
+    var hostnameMatched = (
+      isMatchedAt > -1 &&
+      (isMatchedAt === hostname.length - noProxyZone.hostname.length)
```
bnoordhuis: Correct me if I'm wrong but isn't this just a circuitous way of saying `hostname === noProxyZone.hostname`?

lib/install.js
```diff
@@ -438,18 +477,32 @@ function download (gyp, env, url) {
...
-  if (proxyUrl) {
-    if (/^https?:\/\//i.test(proxyUrl)) {
-      log.verbose('download', 'using proxy url: "%s"', proxyUrl)
-      requestOpts.proxy = proxyUrl
-    } else {
-      log.warn('download', 'ignoring invalid "proxy" config setting: "%s"', proxyUrl)
+  var noProxyUrl = gyp.opts.no_proxy
+    || env.no_proxy
+    || env.NO_PROXY
+    || env.npm_config_no_proxy
```
bnoordhuis: This should probably go; the npm pull request wasn't merged.

lib/install.js
```diff
@@ -438,18 +477,32 @@ function download (gyp, env, url) {
...
+    || ''
+
+  if (noProxyUrl === '*' || (noProxyUrl !== '' && uriInNoProxy(url, noProxyUrl))) {
```
bnoordhuis: Long line; can you keep it under 80 columns?  The parens around the `&&` aren't necessary.

lib/install.js
```diff
@@ -438,18 +477,32 @@ function download (gyp, env, url) {
...
+      || env.http_proxy
+      || env.HTTP_PROXY
+      || env.npm_config_proxy
+
+    if (proxyUrl) {
+      if (/^https?:\/\//i.test(proxyUrl)) {
+        log.verbose('download', 'using proxy url: "%s"', proxyUrl)
+        requestOpts.proxy = proxyUrl
+      } else {
+        log.warn('download', 'ignoring invalid "proxy" config setting: "%s"', proxyUrl)
```
bnoordhuis: Ditto, long line.

test/test-download-options.js
```diff
@@ -0,0 +1,63 @@
+var test = require('tape')
```
bnoordhuis: Can you add `'use strict';`?

test/test-download-options.js
```diff
@@ -0,0 +1,63 @@
...
+    opts: {
+      proxy: 'http://www.test.com'
+    }
+  }
+
+  var req = install.test.download(gyp, {}, url)
+
+  t.equal(req.opts.proxy, gyp.opts.proxy)
+  t.equal(req.opts.uri, url)
+
```
bnoordhuis: Can you drop the blank line here and below?

dlangerenken:
@bnoordhuis is someone taking care of this PR? Seems like @ballwood is no longer working on it. It would be really nice to have in master.

rvagg:
This looks pretty good and we have `HTTPS_PROXY` support about to land too but it's going to need a new champion. If someone wants this in, please take this code, clean it up as per the suggestions in here and make it work against master. Tests would be good too if you can come up with some that are meaningful.

lib/install.js
alsotang: IMO it's a suffix match

imatlopez:
@rvagg I was looking at doing just this given this is quite a pain point inside our corporate network, however is this approach to make it independent of requests, or would it be simpler to set the environment variables from the cli flags as needed and let requests do its job?

rvagg:
@imatlopez you mean `request` not `requests` right? Apparently obeys no_proxy these days but I don't know what we would need to do here, if anything, for that support to flow through. If it's obeying environment variables then I'd assume we don't have to do anything on our end, just set the env var before calling. If you need gyp.opts to be able to carry it then I guess some additional work could go in here and hopefully that'd be simple. You're welcome to open a PR if you'd like it added.

PR #1181, opened 2017-04-23
Change source path for cmake

karboom:
If use relative path for source file, it will give an not found error when I include CMakeLists from other CMakeLists
``` cmake
include(build/Release/CMakeLists.txt)
```

I'm using CLion for dev


bnoordhuis:
I don't quite follow.  node-gyp doesn't use GYP's cmake generator.  What (or who) is this change for?

PR #1152, opened 2017-03-21
doc: add a note for using `configure` on Windows

vsemozhetbyt:
Fixes: https://github.com/nodejs/node-gyp/issues/1149

PR #1182, opened 2017-04-23
doc: make Windows command look more Windowsy

refack:
Follow up to #1152

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,8 @@ $ node-gyp configure
 
 Auto-detection fails for Visual C++ Build Tools 2015, so `--msvs_version=2015`
 needs to be added (not needed when run by npm as configured above):
-``` bash
-$ node-gyp configure --msvs_version=2015
+```ShellSession
+c:\code\my-addon> node-gyp configure --msvs_version=2015
 ```
 
 __Note__: The `configure` step looks for the `binding.gyp` file in the current


gibfahn:
Would `bat` or `cmd` make more sense [see language options](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L348)?

refack:
I thought about it but it's not a script it's a `ShellSession` ...

gibfahn:
>I thought about it but it's not a script it's a ShellSession ...

Sure, but `ShellSession` [applies to bash as well](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L4063), and is there a difference between interactively running commands and running them in a batch file (again, genuine question, there isn't in Bash, which is why we can use `bash`).

refack:
If I had used `cmd` keywords ```cmd would have been more informative...

Yes there are differences:
1. First of all there's the prompt (`$` or `xxx>`) showing in each line, for both environments.
2. `cmd` has a weird this with variables. In an interactive session there are late revoled, and in a batch they are early resolved. so in an interactive session you would write:
```cmd
FOR %i IN (1,2,3,4) DO ECHO %i
```
but in a batch you have to escape them:
```cmd
FOR %%i IN (1,2,3,4) DO ECHO %%i
```
3. `cmd` doesn't have function so there's a weird thing about multiline statements, but I'm not getting into that...

refack:
Not worth the bother.

PR #1198, opened 2017-05-13
win: more robust parsing of SDK version

refack:
Fixes: https://github.com/nodejs/node-gyp/issues/1179

Personally I prefer #1178, but this is smaller.

Issue #1179, opened 2017-04-18
v3.6.0 incompatible with super new SDKs 10.0.15063.Desktop


himadrinath:
see this docs: https://github.com/nodejs/node-gyp/issues/1144

refack:
If you're seeing errors similar to this with version `v3.6.1`

```
C:\Users\himad\Desktop\testProject\node_modules\sharp>if not defined npm_config_node_gyp (node "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\bin\node-gyp-bin\....\node_modules\node-gyp\bin\node-gyp.js" rebuild ) else (node "" rebuild )
Traceback (most recent call last):
File "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\gyp\gyp_main.py", line 16, in 
sys.exit(gyp.script_main())
File "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\gyp\pylib\gyp_init_.py", line 545, in script_main
return main(sys.argv[1:])
File "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\gyp\pylib\gyp_init_.py", line 538, in main
return gyp_main(args)
File "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\gyp\pylib\gyp_init_.py", line 514, in gyp_main
options.duplicate_basename_check)
File "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\gyp\pylib\gyp_init_.py", line 98, in Load
generator.CalculateVariables(default_variables, params)
File "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\gyp\pylib\gyp\generator\msvs.py", line 1918, in CalculateVariables
generator_flags.get('msvs_version', 'auto'))
File "C:\Users\himad\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\gyp\pylib\gyp\MSVSVersion.py", line 434, in SelectVisualStudioVersion
versions = _DetectVisualStudioVersions(version_map[version], 'e' in version)
KeyError: '2017'
gyp ERR! configure error
gyp ERR! stack Error: gyp failed with exit code: 1
[...]

PR #1205, opened 2017-05-22
Detect visual studio 2017

danielgindi:
This should have been done long ago...

### DIFF
--- a/gyp/pylib/gyp/MSVSVersion.py
+++ b/gyp/pylib/gyp/MSVSVersion.py
@@ -226,6 +226,15 @@ def _CreateVersion(name, path, sdk_based=False):
   if path:
     path = os.path.normpath(path)
   versions = {
+      '2017': VisualStudioVersion('2017',
+                                  'Visual Studio 2017',
+                                  solution_version='13.00',
+                                  project_version='15.0',
+                                  flat_sln=False,
+                                  uses_vcxproj=True,
+                                  path=path,
+                                  sdk_based=sdk_based,
+                                  default_toolset='v141'),
       '2015': VisualStudioVersion('2015',
                                   'Visual Studio 2015',
                                   solution_version='12.00',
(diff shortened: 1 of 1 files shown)


refack:
@danielgindi thanks for the contribution! 🥇 
As of `node-gyp@3.6.0` detection of VS2017 is already in.
(if you don't have > `v3.6.0` you'll need to update `npm`)

danielgindi:
Actually 3.6.1 is installed, a fresh install - and do not have the VS2017 detection in...
It does use it somehow with an env called `GYP_MSVS_OVERRIDE_PATH`, which does not exist on my system so I have no idea where it comes from.
But on another system - also 3.6.1 just installed fresh - no VS2017 detection...

refack:
> It does use it somehow with an env called GYP_MSVS_OVERRIDE_PATH

https://github.com/nodejs/node-gyp/blob/master/lib/configure.js#L156

We wanted to have VS2017 support before it landed in upstream `GYP` so we "lie" to it to make it work.
If you can't build with `v3.6.1` please supply the error output. Also look if it's related to https://github.com/nodejs/node-gyp/issues/1179

danielgindi:
Okay so I've debugged the `configure.js`, and `findVS2017` says that `"Error: Could not use PowerShell to find VS2017"`
Looking further into it...

refack:
See if this https://github.com/nodejs/node-gyp/pull/1198 helps

danielgindi:
`'Exception calling "Query" with "0" argument(s): "Input string was not in a correct format."\r\nAt line:1 char:95\r\n+ ... e-gyp\\lib\\Find-VS2017.cs\'; [VisualStudioConfiguration.Main]::Query()}\r\n+                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\r\n    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException\r\n    + FullyQualifiedErrorId : FormatException\r\n \r\n'`

danielgindi:
Yes, #1198 solves it. So we're waiting for v3.6.2, and for it to reach npm upstream...

danielgindi:
Seems like Windows Creators Update has this side effect of making vs2017 harder to detect by the old code.
This really needs to be released soon...

refack:
Will be in `npm@5.0.2`

refack:
Will be in `npm@5.0.2` 
Ref: https://github.com/npm/npm/pull/16894

PR #1206, opened 2017-05-22
build: rename copyNodeLib() to doBuild()

lc-soft:
@refack 
I use the `npm test` command to run the test, unfortunately, the test does not pass, in which there is a error message:

``` text
...
gyp verb node dev dir C:\Users\鍒樿秴\.node-gyp\6.3.1
gyp verb found first Solution file build/binding.sln
gyp verb using MSBuild: D:\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MSBuild.exe
gyp ERR! UNCAUGHT EXCEPTION
gyp ERR! stack ReferenceError: copyNodeLib is not defined
gyp ERR! stack     at findMsbuild (F:\浠ｇ爜搴揬GitHub\node-gyp\lib\build.js:127:7)
gyp ERR! stack     at F:\浠ｇ爜搴揬GitHub\node-gyp\lib\build.js:107:11
gyp ERR! stack     at F (F:\浠ｇ爜搴揬GitHub\node-gyp\node_modules\.1.2.14@which\which.js:68:16)
gyp ERR! stack     at E (F:\浠ｇ爜搴揬GitHub\node-gyp\node_modules\.1.2.14@which\which.js:80:29)
gyp ERR! stack     at F:\浠ｇ爜搴揬GitHub\node-gyp\node_modules\.1.2.14@which\which.js:89:16
gyp ERR! stack     at F:\浠ｇ爜搴揬GitHub\node-gyp\node_modules\.2.0.0@isexe\index.js:42:5
gyp ERR! stack     at F:\浠ｇ爜搴揬GitHub\node-gyp\node_modules\.2.0.0@isexe\windows.js:36:5
gyp ERR! stack     at FSReqWrap.oncomplete (fs.js:123:15)
gyp ERR! System Windows_NT 10.0.14393
gyp ERR! command "D:\\nodejs\\node.exe" "F:\\浠ｇ爜搴揬\GitHub\\node-gyp\\bin\\node-gyp.js" "rebuild" "-C" "F:\\浠ｇ爜搴揬\GitHub\\node-gyp\\test\\node_modules\\hello_world" "--loglevel=verbose"
gyp ERR! cwd F:\浠ｇ爜搴揬GitHub\node-gyp\test\node_modules\hello_world
gyp ERR! node -v v6.3.1
gyp ERR! node-gyp -v v3.6.1
gyp ERR! This is a bug in `node-gyp`.
gyp ERR! Try to update node-gyp and file an Issue if it does not help:
[...]

PR #1212, opened 2017-05-31
Upgrade to tar v3

isaacs:
Tar version 3 performs better and is more well tested than its
predecessor.  npm will be using this in the near future, so there is no
benefit in shipping a node-gyp that uses the slower and less reliable
fstream-based tar.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -15,10 +15,8 @@ var fs = require('graceful-fs')
   , rm = require('rimraf')
   , path = require('path')
   , crypto = require('crypto')
-  , zlib = require('zlib')
   , log = require('npmlog')
   , semver = require('semver')
-  , fstream = require('fstream')
   , request = require('request')
   , minimatch = require('minimatch')
   , mkdir = require('mkdirp')
@@ -144,41 +142,33 @@ function install (gyp, argv, callback) {
       var tarPath = gyp.opts.tarball
       var badDownload = false
         , extractCount = 0
-        , gunzip = zlib.createGunzip()
-        , extracter = tar.Extract({ path: devDir, strip: 1, filter: isValid })
 
       var contentShasums = {}
       var expectShasums = {}
 
       // checks if a file to be extracted from the tarball is valid.
       // only .h header files and the gyp files get extracted
-      function isValid () {
-        var name = this.path.substring(devDir.length + 1)
-        var isValid = valid(name)
-        if (name === '' && this.type === 'Directory') {
-          // the first directory entry is ok
-          return true
-        }
+      function isValid (path, entry) {
+        var isValid = valid(path)
         if (isValid) {
-          log.verbose('extracted file from tarball', name)
+          log.verbose('extracted file from tarball', path)
           extractCount++
         } else {
           // invalid
-          log.silly('ignoring from tarball', name)
+          log.silly('ignoring from tarball', path)
         }
         return isValid
       }
 
-      gunzip.on('error', cb)
-      extracter.on('error', cb)
-      extracter.on('end', afterTarball)
-
-      // download the tarball, gunzip and extract!
+      // download the tarball and extract!
 
       if (tarPath) {
-        var input = fs.createReadStream(tarPath)
-        input.pipe(gunzip).pipe(extracter)
-        return
+        return tar.extract({
+          file: tarPath,
+          strip: 1,
+          filter: isValid,
+          cwd: devDir
+        }).then(afterTarball, cb)
       }
 
       try {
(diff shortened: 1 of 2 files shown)


lib/install.js
```diff
@@ -218,7 +208,11 @@ function install (gyp, argv, callback) {
         })
 
         // start unzipping and untaring
-        req.pipe(gunzip).pipe(extracter)
+        res.pipe(tar.extract({
+          strip: 1,
+          cwd: devDir,
+          filter: isValid
+        }).on('close', _ => setTimeout(afterTarball, 1000)).on('error', cb))
```
bnoordhuis: Can you use a non-arrow function here?  We still nominally support v0.10 and v0.12.  Does tar@3 support pre-ES6 node?

Side question: why the 1s delay?

isaacs: tar@3 does not support node 0.x.  Neither does npm@5.

I thought they'd fallen off the LTS window.  How long are they going to be nominally supported?

The timeout is a bit of debugging scaffolding that I neglected to remove.  Good catch, thanks.  I'll rebase that out right now.

isaacs:
Note: This drops support for node 0.x, and thus should be considered a breaking semver-major change, and not landed until those node versions are no longer supported.

isaacs:
@bnoordhuis I see at the [LTS schedule](https://github.com/nodejs/LTS) that 0.12 was only in "maintenance" through 2016-12-31.  I think that means it _isn't_ nominally supported?  Still a major version bump, of course.

Fishrock123:
0.12 is no longer supported by Node.js, whether or not 0.12 is still supported by the current version of node-gyp, I am not really sure.

addaleax:
I’d say you can’t just remove 0.x support in a non-major release; too many CI configurations that use `node-gyp` still have those in them.

bnoordhuis:
I suppose that now that node 8 is out, we might as well bump major and drop support for node < 4.  There still seem to be quite a few v0.10.x users but those people probably don't upgrade frequently in general and node-gyp in particular.

refack:
Is there an LST/maintenance plan for `node-gyp`?
FWIW there seem to be [very low code traffic](https://github.com/nodejs/node-gyp/commits/master) and IMHO maintaining a parallel `v3.x` branch should be very easy.

BTW: to the package owners we need a `v3.6.2` for #1206, and #1205

refack:
@isaacs IMHO this PR should come with a `"engines" : { "node" : ">=4" },` clause in `package.json`. Or make this PR dependent on a PR with this change.

Fishrock123:
Ehhhh, if people really want a 3.x maintenance we'll probably only know if they complain about it so I'd say leave it be and consider doing something if people post issues about that specifically.

How do we manage these around here? Merge everything into `master` or another branch or?

gibfahn:
>Ehhhh, if people really want a 3.x maintenance we'll probably only know if they complain about it so I'd say leave it be and consider doing something if people post issues about that specifically.

Agreed

>Merge everything into master or another branch or?

Might as well stay on master, we could create a `3.x` branch if we need to do more 3.x releases going forward.

refack:
> How do we manage these around here? Merge everything into master or another branch or?

I don't know of a president here. I agree with @gibfahn work on `master`, and if needed create a branch from last known good v3, and backport any requested fixes.

isaacs:
You all could probably check the nodejs.org log files to see how many 0.x users are downloading gyp headers, but here at npm, we see that 0.10 and 0.12 together account for a little less than 6% of npm registry traffic, and they're almost all using older versions of npm (and thus, won't get a major version bump of node-gyp).  All other 0.x node versions make up less than 0.1% of npm registry traffic.

In other words, it's almost certainly safe to drop support for 0.x in a major version bump.

refack:
@isaacs if I read this thread correctly there's consensus on dropping < 4 support in a major version bump.

AFAICT We just don't have an established procedure on how to maintain two version lines of a package (only personal experience). IMHO it does not need to be as strict as `node core`'s. Especial since `node-gyp` if very low in code traffic.

I say we just wing it 😉 since collectively we have plenty of prior experience in maintaining packages.

isaacs:
@refack updated package.json engines field to specify >= 4

refack:
> @refack updated package.json engines field to specify >= 4

10x, lets see how much the CI is broken

refack:
https://ci.nodejs.org/job/nodegyp-test-pull-request/21/

refack:
Landed in 5f924ce62c9bca9ab9c2e547bfb87b9a391271ed & 75cfae290fee1791a23fa68820ae5dd841e93e14

Issue #1216, opened 2017-05-31
Opposite of --ensure, and is --ensure still useful?


gibfahn:
I was looking for a way to force `node-gyp` to re-download the headers even if the directory already exists in `$npm_config_devdir/$node_version`, and I found `--ensure`, which seems to do the opposite.

As I understand it, `node-gyp` doesn't download the headers unless the directory is missing, so `--ensure` is the default behaviour. Am I missing something? Is there a `--always-redownload` option (and should there be)?

Obviously as a workaround you can do this before the `node-gyp rebuild` or `npm install`:

```bash
node "$(npm config get prefix)"/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js remove $(node -p process.versions.node) 
```

gibfahn:
Thinking about this further, if the `--tarball` option is passed, I think it makes sense to delete the header directory and recreate it.

Thoughts?

bnoordhuis:
> As I understand it, node-gyp doesn't download the headers unless the directory is missing, so --ensure is the default behaviour. Am I missing something?

That's correct.  It's unconditionally set in lib/configure.js.

> Is there a --always-redownload option (and should there be)?

I don't think so.  What good would it do?

> Thinking about this further, if the --tarball option is passed, I think it makes sense to delete the header directory and recreate it.

I suppose.  It could piggyback on the remove command.

gibfahn:
>I don't think so. What good would it do?

I ran into this because we were testing `v8.0.0-pre`, and `node-gyp` didn't redownload the headers, which meant that we got lots of this error when the `NODE_MODULE_VERSION` was updated:

```bash
module.js:598
  return process.dlopen(module, path._makeLong(filename));
                 ^

Error: The module '/space/jenkins-local/workspace/ibm-node8-appmetrics/ARCH/x64/CC/gcc48/OS/linux/label/build/node-v8.0.0-linux-x64/lib/node_modules/appmetrics/appmetrics.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 55. This version of Node.js requires
NODE_MODULE_VERSION 57. Please try re-compiling or re-installing
```

When running in CI I always want to redownload the headers.

However if the "`--tarball` always reinstalls the headers" option is acceptable I'm happy with that.

PR #1220, opened 2017-06-02
configure: don't set ensure if tarball is set

gibfahn:
If you're providing a path to a header tarball to install, you probably
want it to always be re-installed.

Fixes: https://github.com/nodejs/node-gyp/issues/1216

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -72,8 +72,10 @@ function configure (gyp, argv, callback) {
         return callback(new Error('Invalid version number: ' + release.version))
       }
 
-      // ensure that the target node version's dev files are installed
-      gyp.opts.ensure = true
+      // If the tarball option is set, always remove and reinstall the headers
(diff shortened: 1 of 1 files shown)


refack:
Dry-run CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/18/

gibfahn:
I think that CI was for the wrong PR number.

CI 2: https://ci.nodejs.org/job/nodegyp-test-pull-request/20/

gibfahn:
>A test would be nice but it's probably hard to test in isolation.

Yeah, I was thinking about how to add a test, but I'm not sure of the best way to do it. To do a unit test I'd want to rewire [`getNodeDir()`](https://github.com/nodejs/node-gyp/blob/master/lib/configure.js#L48) without dealing with `createBuildDir()`, but that doesn't seem straightforward.

PR #1256, opened 2017-07-27
build,win: use config.gypi MSBuild path over env

jaimecbernardo:
When vs2017 and vs2015 are both installed and vs2015's MSBuild.exe is in the Path environment variable, node-gyp fails to build if it configured the project to build with vs2017, because it tries to use vs2015's MSBuild.exe, which is unable to find the v141 (vs2017) Platform Toolset build tools.

Change build behaviour to prefer using the MSBuild.exe path if it is defined in config.gypi instead of using the MSBuild.exe found in Path by "which".

Ref: https://github.com/nodejs/node/issues/13641

##### Checklist

##### Affected core subsystems(s)
gyp,win,build


cjhoward92:
Any news on this? I see this as a helpful enhancement. It's an old PR but still useful IMO.

joaocgreis:
@jaimecbernardo sorry this went unnoticed for so long. Looks good at first sight, but I'd like to run CI. Can you rebase this commit on the current master?

jaimecbernardo:
Hi @joaocgreis ,
Sure. I've just rebased it.

joaocgreis:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/97/

joaocgreis:
This no longer fixes the original issue, the VS2015 command prompt has the environment variable `VCTargetsPath` that makes MSBuild from VS2017 fail.

rvagg:
I'm going to close this due to age, lack of sync with master, and per @joaocgreis' comment above that this might not even address the original problem. New PR would be welcome if this is still needed!

Issue #1217, opened 2017-06-01
Build error when run from npm script


jan-j:
```
Exit code: 1
Command: sh
Arguments: -c node-gyp rebuild
Directory: /Users/john/projects/photo-booth/app/node_modules/printer
Output:
gyp info it worked if it ends with ok
gyp info using node-gyp@3.6.1
gyp info using node@7.9.0 | darwin | x64
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp info spawn /usr/local/bin/python2
gyp info spawn args [ '/Users/john/.nvm/versions/node/v7.9.0/lib/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'make',
gyp info spawn args   '-I',
gyp info spawn args   '/Users/john/projects/photo-booth/app/node_modules/printer/build/config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/Users/john/.nvm/versions/node/v7.9.0/lib/node_modules/node-gyp/addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/Users/john/.node-gyp/7.9.0/include/node/common.gypi',
gyp info spawn args   '-Dlibrary=shared_library',
gyp info spawn args   '-Dvisibility=default',
gyp info spawn args   '-Dnode_root_dir=/Users/john/.node-gyp/7.9.0',
gyp info spawn args   '-Dnode_gyp_dir=/Users/john/.nvm/versions/node/v7.9.0/lib/node_modules/node-gyp',
gyp info spawn args   '-Dnode_lib_file=node.lib',
gyp info spawn args   '-Dmodule_root_dir=/Users/john/projects/photo-booth/app/node_modules/printer',
gyp info spawn args   '-Dnode_engine=v8',
gyp info spawn args   '--depth=.',
gyp info spawn args   '--no-parallel',
gyp info spawn args   '--generator-output',
gyp info spawn args   'build',
gyp info spawn args   '-Goutput_dir=.' ]
Traceback (most recent call last):
[...]

PR #1183, opened 2017-04-23
Bump gyp 19495aa

refack:
* Bumped `GYP` to latest version
* rebased all floating patched since last bump

##### Checklist

- [x] `npm test` passes
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-guidelines)

##### Affected core subsystem(s)
gyp

PR #1267, opened 2017-08-19
Fix IndexError when parsing GYP files.

bnoordhuis:
GYP automatically turns variables ending in _dir, _file or _path into
absolute paths but didn't check for empty strings.

It interacted badly with variables inherited through the environment
from npm, the `scripts-prepend-node-path=false` setting in particular
because it is turned into `npm_config_script_prepend_node_path=`.

Fixes: https://github.com/nodejs/node-gyp/issues/1217
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/32/

### DIFF
--- a/gyp/pylib/gyp/input.py
+++ b/gyp/pylib/gyp/input.py
@@ -2033,7 +2033,7 @@ def MakePathRelative(to_file, fro_file, item):
         gyp.common.RelativePath(os.path.dirname(fro_file),
                                 os.path.dirname(to_file)),
                                 item)).replace('\\', '/')
-    if item[-1] == '/':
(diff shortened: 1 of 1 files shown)


refack:
@bnoordhuis you want to upstream it?

bnoordhuis:
> you want to upstream it?

Do you think it's useful?  It's not like we pull in changes from upstream anymore.

refack:
>> you want to upstream it?
>
> Do you think it's useful? It's not like we pull in changes from upstream anymore.

I try to for node core. And there's https://github.com/nodejs/node-gyp/pull/1183.
(Plus they have great try-bots)

Issue #1299, opened 2017-09-26
Unable to install node-gyp under node 0.10.x SyntaxError: Use of const in strict mode


j4y:
* **Node Version**:  node 0.10.36 and npm 1.4.28
* **Platform**: Linux (CentOS 7)
* **Compiler**:  gcc version 4.8.5 20150623 (Red Hat 4.8.5-16) (GCC) 
* **Module**: node-gyp@0.10.6

<details><summary>Verbose output (from npm or node-gyp):</summary>

```
gyp info it worked if it ends with ok
gyp info using node-gyp@0.10.6
gyp info using node@0.10.36 | linux | x64

/usr/local/lib/node_modules/node-gyp/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
const Hoek = require('hoek');
^^^^^
gyp ERR! UNCAUGHT EXCEPTION 
gyp ERR! stack SyntaxError: Use of const in strict mode.
gyp ERR! stack     at Module._compile (module.js:439:25)
gyp ERR! stack     at Object.Module._extensions..js (module.js:474:10)
gyp ERR! stack     at Module.load (module.js:356:32)
gyp ERR! stack     at Function.Module._load (module.js:312:12)
gyp ERR! stack     at Module.require (module.js:364:17)
gyp ERR! stack     at require (module.js:380:17)
gyp ERR! stack     at Object.<anonymous> (/usr/local/lib/node_modules/node-gyp/node_modules/request/node_modules/hawk/lib/index.js:5:33)
gyp ERR! stack     at Module._compile (module.js:456:26)
gyp ERR! stack     at Object.Module._extensions..js (module.js:474:10)
gyp ERR! stack     at Module.load (module.js:356:32)
gyp ERR! System Linux 3.10.0-514.26.2.el7.x86_64
gyp ERR! command "node" "/usr/local/bin/node-gyp" "rebuild"
gyp ERR! cwd /usr/local/licode/erizoAPI
gyp ERR! node -v v0.10.36
gyp ERR! node-gyp -v v0.10.6
gyp ERR! This is a bug in `node-gyp`.
gyp ERR! Try to update node-gyp and file an Issue if it does not help:
gyp ERR!     <https://github.com/TooTallNate/node-gyp/issues>

```

</details>

This used to work, but is now failing for us.  Any advice on how to proceed would be appreciated.  We are trying to run a legacy system, which uses node 0.10.x

[...]

PR #1300, opened 2017-09-26
deps: pin `request` to ~2.81.0

refack:
##### Checklist

- [X] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-message-guidelines)

##### Description of change
Required for "node < 4" compatibility and is congruent with `npm`

Fixes: https://github.com/nodejs/node-gyp/issues/1299


bnoordhuis:
Another thing, if we ever decide to upgrade the node-gyp in the v4.x and v6.x branches, that too would result in duplication: their copies of npm are stuck on request@2.75.0.

Not the end of the world, just something to keep in mind.

refack:
> Another thing, if we ever decide to upgrade the node-gyp in the v4.x and v6.x branches, that too would result in duplication: their copies of npm are stuck on request@2.75.0.

I'll test if `<2.82.0` get npm to dedup...

refack:
Indeed setting to `<2.82.0` allows deduping:
```
$ npm ls | grep req
npm@2.15.12 d:\code\4node\test-npm-request\npm
| +-- request@2.74.0 deduped
| +-- request@2.74.0 deduped
+-- request@2.74.0
```

addaleax:
@refack can you also give the minimum version in that specifier if you use `<`? I think right now you’d match *any* version of request before 2.82.0, up from `0.8.3` (which I assume does not work :))

refack:
> I think right now you’d match any version of request before 2.82.0, up from 0.8.3 (which I assume does not work :))

Ack.
Did some bisecting - minimal version to pass `npm test` is `2.9.0`.

addaleax:
thanks!

bnoordhuis:
Ping @refack.

edi9999:
@refack Could you please release of the 3.x with this change ?

PR #1269, opened 2017-08-20
python: more informative error

refack:
##### Checklist

- [X] `npm install && npm test` passes
- [X] tests are included 
- [X] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-message-guidelines)

##### Description of change
more informative error when failing to find a suitable python:
```
gyp info it worked if it ends with ok
gyp verb cli [ 'C:\\bin\\dev\\node\\node4.exe',
gyp verb cli   'D:\\code\\4node\\node-gyp\\bin\\node-gyp.js',
gyp verb cli   'configure',
gyp verb cli   '--loglevel=verbose' ]
gyp info using node-gyp@4.0.0
gyp info using node@4.8.2 | win32 | x64
gyp verb command configure []
gyp verb check python checking for Python executable "python2" in the PATH
gyp verb `which` failed Error: not found: python2
gyp verb `which` failed     at getNotFoundError (D:\code\4node\node-gyp\node_modules\which\which.js:13:12)
gyp verb `which` failed     at F (D:\code\4node\node-gyp\node_modules\which\which.js:68:19)
gyp verb `which` failed     at E (D:\code\4node\node-gyp\node_modules\which\which.js:80:29)
gyp verb `which` failed     at D:\code\4node\node-gyp\node_modules\which\which.js:89:16
gyp verb `which` failed     at D:\code\4node\node-gyp\node_modules\isexe\index.js:42:5
gyp verb `which` failed     at D:\code\4node\node-gyp\node_modules\isexe\windows.js:36:5
gyp verb `which` failed     at FSReqWrap.oncomplete (fs.js:82:15)
gyp verb `which` failed  python2 { [Error: not found: python2]
gyp verb `which` failed   stack: 'Error: not found: python2\n    at getNotFoundError (D:\\code\\4node\\node-gyp\\node_modules\\which\\which.js:13:12)\n    at F (D:\\code\\4node\\node-gyp\\node_modules\\which\\which.js:68:19)\n    at E (D:\\code\\4node\\node-gyp\\node_modules\\which\\which.js:80:29)\n    at D:\\code\\4node\\node-gyp\\node_modules\\which\\which.js:89:16\n    at D:\\code\\4node\\node-gyp\\node_modules\\isexe\\index.js:42:5\n    at D:\\code\\4node\\node-gyp\\node_modules\\isexe\\windows.js:36:5\n    at FSReqWrap.oncomplete (fs.js:82:15)',
[...]

PR #1325, opened 2017-11-04
Fixed handling of missing python on Windows. 

mvidmar:
##### Checklist
- [ ] tests are included

##### Description of change
configure: fixing to better handle python badVersion on windows platform

- invalid error message when python is found but is invalid version 
error returned: file not found but with valid file path which is 
confusing. error expected: invalid python version.
- in case of invalid version on guessed path (python27) it goes into
an endless loop (unlikely but an ugly error)

refs https://github.com/nodejs/node-gyp/pull/1269

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,26 @@
+<!--
+Thank you for reporting an issue. The more information you can give us, the
+better the chance we can fix your problem.
+
+This issue tracker is for issues with node-gyp,
+if you have an issue installing a specific module, please file an issue on
+that module's issue tracker (`npm issues modulename`).
+-->
+
+* **Node Version**: <!-- `node -v` and `npm -v` -->
+* **Platform**: <!-- `uname -a` (UNIX), or `systeminfo | findstr /B /C:"OS Name" /C:"OS Version" /C:"System Type"` (Windows) -->
+* **Compiler**: <!-- `cc -v` (UNIX) or `msbuild /version & cl` (Windows) -->
+* **Module**: <!-- what you tried to build/install -->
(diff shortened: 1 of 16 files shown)


refack:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/41/

refack:
Hello @mvidmar and welcome! Thank you for the fix 🥇 

Could you describe the situation that causes the endless recursion (I want to try to design a test around that)?

Also If you could adjust the commit message to our style guide (which is roughly based on [node-core's](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-message-guidelines))

mvidmar:
Tried to fix the commit message according to the style guide, let me know if I missed something.

The loop happens on Windows when c:\Python27\python.exe is not a valid version (like I mentioned - unlikely but possible. Happened to me during trying to get an understanding of this code). The loop happens from checkPythonVersion, call to checkPythonLauncher, then guessPython, which finds the executable and then reruns checkPythonVersion. That's why I added an increase of  checkPythonLauncherDepth within guessPython so it skips the fallback to checkPythonLauncher.

I also merged our fixes in case you want to pull that instead of this fix. Since I noticed you modified some other stuff as well. Here is the merged fix with PR 1269 https://github.com/mvidmar/node-gyp/commit/9c7ad90d5caeb809580ec1620ef3853a677cb503

rvagg:
hi @mvidmar do you still want to pursue this? if so, rebase against master and let's sort out a few style nits and get some windows people involved again

joaocgreis:
The current code to find Python already includes both these features, I think this can be closed. If I missed something, please rebase on current master. Thanks!

PR #1340, opened 2017-11-16
Fix: module compilation within termux-app on android

arrkiin:
##### Description of change
To solve the problem of linking a compiled module against a preinstalled library of termux-app on android I added a special condition branch to add the required -fPIC flag for the c and c++ compiler. After that everything compiled, linked and works as expected.

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -52,6 +52,14 @@
         'standalone_static_library': '<(standalone_static_library)'
       }],
 
+      ['_type!="executable"', {
+        'conditions': [
+          [ 'OS=="android"', {
+            'cflags!': [ '-fPIE' ],
+          }]
+        ]
+      }],
+
       ['_win_delay_load_hook=="true"', {
         # If the addon specifies `'win_delay_load_hook': 'true'` in its
         # binding.gyp, link a delay-load hook into the DLL. This hook ensures
@@ -135,10 +143,10 @@
           '_FILE_OFFSET_BITS=64'
         ],
       }],
-      [ 'OS in "freebsd openbsd netbsd solaris" or \
+      [ 'OS in "freebsd openbsd netbsd solaris android" or \
          (OS=="linux" and target_arch!="ia32")', {
         'cflags': [ '-fPIC' ],
-      }]
+      }],
     ]
   }
 }


addon.gypi
```diff
@@ -138,7 +138,11 @@
       [ 'OS in "freebsd openbsd netbsd solaris" or \
          (OS=="linux" and target_arch!="ia32")', {
         'cflags': [ '-fPIC' ],
-      }]
+      }],
+      [ 'OS == "android"', {
+        'cflags': [ '-fPIC' ],
+        'cflags_cc': ['-fPIC' ],
+      }],
```
bnoordhuis: You can fold this into the previous condition: `OS in "android freebsd ..."`.  You don't need to add it to `cflags_cc` because `cflags` is for both C and C++ code.

arrkiin: @bnoordhuis thanks for your comment. Your proposal was my first attemp to fix it, but it didn't. To verify if I didn't make any mistake I just did some testing:
- with 'cflags_cc': ['-fPIC' ] is compiles
- without 'cflags_cc': ['-fPIC' ] is does not
I think I have to compare the verbose outputs in more detail to figure out what happens. Any ideas?

bnoordhuis: `env V=1 node-gyp rebuild` turns on verbose compilation mode (i.e., prints the compile commands.)

arrkiin: This is the compiler command for the variant **without ccflags_cc**. It seems that the `-fPIE` after the first `-fPIC` creates and unlinkable shared library on android.
`...  -fPIC -Wall -Wextra -Wno-unused-parameter -O3 -fno-omit-frame-pointer -fPIE  -fno-rtti ...`

**With ccflags_cc** the command is build like that. It seems, that the later `-fPIC` overrides the preceding `-fPIE` so that it produces the right shared library.
`...  -fPIC -Wall -Wextra -Wno-unused-parameter -O3 -fno-omit-frame-pointer -fPIE -fPIC -fno-rtti ...`

With this result I would rather try to remove the `-fPIE` while compiling shared libraries on android. I added an additional entry in `target_condition` for removing `-fPIE` like shown below. That did the trick and produces the following compiler command.
`... -fPIC -Wall -Wextra -Wno-unused-parameter -O3 -fno-omit-frame-pointer -fno-rtti ...`

Excerpt of my local `addon.gypi`:
```
'target_conditions': [
...
      ['library=="shared_library"', {
        'conditions': [
          [ 'OS=="android"', {
            'cflags!': [ '-fPIE' ],
          }]
        ]
      }],
...
]
```
@bnoordhuis, what is your comment on this?

bnoordhuis: Ah, you're probably getting hit by #1118 - that `-fPIE` flag is not set by node-gyp, it's inherited from node's common.gypi.

The right thing to do until #1118 is resolved is probably to suppress the flag in node-gyp's addon.gypi.

It must have been broken for quite a while because that flag was added 20 months ago in https://github.com/nodejs/node/pull/5544.

addon.gypi
```diff
@@ -52,6 +52,14 @@
         'standalone_static_library': '<(standalone_static_library)'
       }],
 
+      ['library=="shared_library"', {
```
bnoordhuis: I _think_ you want to unset `-fPIE` unconditionally, otherwise it's still going to be problematic with add-ons that also build static libraries.

arrkiin: Ok. So we only want to remove the flag if we don't want to compile an executable. Am I right?

bnoordhuis:
LGTM, thanks.  CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/42/

You don't even need to special-case on `OS=="android"` when removing `-fPIE` but this is good enough.

arrkiin:
@bnoordhuis, when can I expect it to be merged and published?

bnoordhuis:
We don't release that often and it also takes a while for new versions of node-gyp make it into npm and node.

I can merge it but can you squash your commits and write up a nice commit log?  If you want to show up under your real name instead of arrkiin, please do `git commit --amend --author="..."`.  Cheers.

pietvanzoen:
Hey folks. Would love to see this merged. Running into this issue currently. :)

arrkiin:
Hey @bnoordhuis, sorry for the long delay. After your message I asked myself how to proceed and forgot to continue. In the meantime I used my own fork to do the experiments under termux. Until now I didn't see any problems. Tody I just found some time to squash the commits and add my forename.

maclover7:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/80/

johnnyvf24:
> Hey folks. Would love to see this merged. Running into this issue currently. :)
Make that two.

refack:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/100/

PR #1057, opened 2016-11-23
remove the need for xcodebuild

gdams:
xcodebuild now requires full xcode and not just developer tools
I have removed all commands that use xcode build and there seems to be no issues. This also means that we don't get the error message anymore.
What is the reason we actually need to use xcodebuild?

I know that this is something that should probably be raised with gyp upstream, but I wanted to make sure that this seems reasonable first.

PR #1370, opened 2018-01-21
gyp: don't print xcodebuild not found errors

gibfahn:
As node-gyp rebuild doesn't seem to need xcodebuild, we don't need to be printing the error every time GYP is run.

Refs: https://github.com/nodejs/node-gyp/pull/1057
Refs: https://chromium-review.googlesource.com/c/492046/
Refs: https://github.com/nodejs/node-gyp/issues/569

As the previous attempts haven't landed, I thought I'd aim for a minimal diff. This seems to remove all `node-gyp build` warnings during `npm install` of a native module, and errors from running configure in node core, although for some reason the first line in Node's built-in gyp [uses `xcrun` rather than `xcodebuild`](https://github.com/nodejs/node/blob/8803b69c724e7b03ec0fc4c82b71575262487624/tools/gyp/pylib/gyp/xcode_emulation.py#L498).

This is caused by me running a "getting started with node" course, and having half the audience stuck because "there was an error in the build".

##### Checklist

### DIFF
--- a/gyp/pylib/gyp/xcode_emulation.py
+++ b/gyp/pylib/gyp/xcode_emulation.py
@@ -429,7 +429,7 @@ class XcodeSettings(object):
     # Since the CLT has no SDK paths anyway, returning None is the
     # most sensible route and should still do the right thing.
     try:
-      return GetStdout(['xcodebuild', '-version', '-sdk', sdk, infoitem])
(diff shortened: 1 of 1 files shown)


gibfahn:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/45/

PR #1392, opened 2018-03-15
I had trouble finding the Windows build tools.  Fixed broken link in README.md

rpfister51:
##### Checklist

- [ ] tests are included 
- [ ] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-message-guidelines)

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -1,7 +1,12 @@
+module.exports = exports = function (gyp, argv, callback) {
+  return install(fs, gyp, argv, callback)
+}
 
-module.exports = exports = install
-
-module.exports.test = { download: download, readCAFile: readCAFile }
+module.exports.test = {
+  download: download,
+  install: install,
+  readCAFile: readCAFile,
+}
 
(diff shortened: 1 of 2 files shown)


bzoz:
Current link works on my box.

rpfister51:
Okay.  The link is not "broken".  But it leads to a non-specific page.  By non-specific I mean that it leads to a page that has no "Default Install" option, but instead a link "Download Build Tools for Visual Studio 2017".  That link leads to a page with 22 download buttons, non of which say "Default Install."  One of these does say "Build Tools for Visual Studio 2017" which probably works; that doesn't happen to be the route I took in the end.  I was trying to be helpful.  I think the current link, and it's description in README.md should be better.

maclover7:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/52/

rvagg:
https://github.com/nodejs/node-gyp/pull/1490

PR #1395, opened 2018-03-21
Add node-gyp builtin config file

akihikodaki:
##### Checklist

- [X] tests are included 
- [X] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#commit-message-guidelines)

##### Description of change

--nodedir is distribution-dependent. For example, some distribution may
ship Node.js with OpenSSL 1.1.0. ts maintainer would want to use OpenSSL
1.1.0 headers rahter than prebuilt OpenSSL 1.0.2 headers in such a case.

This change allows to provide such an option with builtin config file,
as npm does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,10 @@ __`node-gyp` responds to environment variables or `npm` configuration__
  This way only works when `node-gyp` is executed by `npm`:  
  `$ npm config set [--global] devdir /tmp/.gyp`  
  `$ npm i buffertools`
+3. `node-gyp` builtin config file at `/path/to/node-gyp/node-gyp.rc`. This is an
+ unchangeable "builtin" configuration file. Set fields in INI format here. This
+ is primarily for distribution maintainers to override default configs such
+ as `--nodedir` option in a standard and consistent manner.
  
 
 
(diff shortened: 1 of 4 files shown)


bnoordhuis:
Sorry for the delay.  How necessary is this?  I don't really like the idea of taking on another dependency, and - more importantly - introducing another source of truth.

I imagine that if you're a distro maintainer, you're probably served just as well by a shell script wrapper that passes `--nodedir` as an extra arg or through an environment variable.

akihikodaki:
> I imagine that if you're a distro maintainer, you're probably served just as well by a shell script wrapper that passes `--nodedir` as an extra arg or through an environment variable.

Indeed I did so. I'm just wondering why it does not have this kind of configuration file while NPM does.

bnoordhuis:
People usually configure node-gyp through npm; npm passes settings from the command line and .npmrc to node-gyp when it executes it.

bnoordhuis:
I don't think we'll be merging this so I'm going to close it out.  Thanks anyway for the PR.

PR #1436, opened 2018-05-10
Fix space in deps

jeffsenn:
##### Checklist

xxx- [ ] tests are included 
(Note: I'm unsure how to test this case without a lengthy exercise)
- [x] commit message follows [commit guidelines]

##### Description of change
Fixes a narrow issue when spaces are found in paths in the computed dependencies of the Makefile. This occurs in the unusual circumstance that the username and/or home directory has a space


etisdew:
Spaces are not an unusual case, will this be merged?

bnoordhuis:
I'm not 100% positive this is going to catch all cases but it's still a step up.  Landed in 1cf636b, thanks Jeff!

PR #1451, opened 2018-05-30
Zos fixes

jBarz:
##### Checklist

- [ ] tests are included 

##### Description of change

On z/OS, we fixed up node-gyp that comes packaged with node 6.x.
This PR transports those fixes here so that node-gyp works as a standalone.

### DIFF
--- a/gyp/pylib/gyp/input.py
+++ b/gyp/pylib/gyp/input.py
@@ -233,7 +233,12 @@ def LoadOneBuildFile(build_file_path, data, aux_data, includes,
     # Open the build file for read ('r') with universal-newlines mode ('U')
     # to make sure platform specific newlines ('\r\n' or '\r') are converted to '\n'
     # which otherwise will fail eval()
-    build_file_contents = open(build_file_path, 'rU').read()
+    if sys.platform == 'zos':
+      # On z/OS, universal-newlines mode treats the file as an ascii file. But since
+      # node-gyp produces ebcdic files, do not use that mode.
+      build_file_contents = open(build_file_path, 'r').read()
+    else:
+      build_file_contents = open(build_file_path, 'rU').read()
   else:
     raise GypError("%s not found (cwd: %s)" % (build_file_path, os.getcwd()))
 
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -244,25 +244,29 @@ function configure (gyp, argv, callback) {
       argv.push('-I', config)
     })
 
-    // for AIX we need to set up the path to the exp file
-    // which contains the symbols needed for linking.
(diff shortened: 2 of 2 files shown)


lib/configure.js
```diff
@@ -260,7 +260,10 @@ function configure (gyp, argv, callback) {
       var candidates = ['include/node/node',
                         'out/Release/node',
                         'out/Debug/node',
-                        'node'].map(function(file) {
+                        'node',
+                        'out/Release/obj.target/libnode',
+                        'out/Debug/obj.target/libnode',
+                        'lib/libnode'].map(function(file) {
```
richardlau: Might be an idea to rewrite the comment just prior to this block as it's no longer AIX specific. We can probably drop the candidate locations from the comment as it's a duplicate of the code. Something like: 
```
// For AIX and z/OS we need to set up the path to the exports file
// which contains the symbols needed for linking. 
```

Also will the exports file ever be `include/node/node.x` on z/OS? If not, we could slightly speed up by using a different set of candidates per platform.

richardlau:
New CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/47/

lib/configure.js
jBarz: Thanks @richardlau 
I will rewrite that comment.
Also the exports file on z/OS will never be `include/node/node.x`
So I will do as you suggested.

jBarz:
new CI after I addressed nits.

refack:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/51/

gyp/pylib/gyp/input.py
```diff
@@ -233,7 +233,12 @@ def LoadOneBuildFile(build_file_path, data, aux_data, includes,
     # Open the build file for read ('r') with universal-newlines mode ('U')
     # to make sure platform specific newlines ('\r\n' or '\r') are converted to '\n'
     # which otherwise will fail eval()
-    build_file_contents = open(build_file_path, 'rU').read()
+    if sys.platform == 'zos':
```
refack: Did you try to upstream it. Or in other words, will it be a problem for `node` (if we have non ASCII chars, possibly in `config.gypi` via `node-prefix` for example) ?

Or will it be better that `node-gyp` will generate UTF-8-BOM files, and skip patching GYP?

jBarz: @refack I tried looking for what `node-prefix` does. but couldn't find anything. Can you point me to some documentation about it?

bnoordhuis:
Cheers, landed in 293092c...eaedc12.

Issue #1454, opened 2018-06-05
Builds fail on macOS with XCode 10


jhermsmeier:
* **Node Version**: Node v10.3.0, npm v6.1.0
* **Platform**: Darwin R2D2 17.6.0 Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64 x86_64 i386 MacBookAir5,2 Darwin
* **Compiler**:
```
Apple LLVM version 10.0.0 (clang-1000.10.25.5)
Target: x86_64-apple-darwin17.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```
* **Module**: https://github.com/resin-io-modules/mountutils
(but happens with anything with native bindings)

<details><summary>Verbose output (from local npm install):</summary>

```
~/W/mountutils (master|ø) $ npm i

> mountutils@1.3.12 install /Users/Jonas/Work/mountutils
> prebuild-install || node-gyp rebuild

prebuild-install info begin Prebuild-install version 2.5.1
prebuild-install info install installing standalone, skipping download.
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance

xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance

Traceback (most recent call last):
  File "/usr/local/n/versions/node/10.3.0/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py", line 16, in <module>
    sys.exit(gyp.script_main())
  File "/usr/local/n/versions/node/10.3.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 545, in script_main
    return main(sys.argv[1:])
  File "/usr/local/n/versions/node/10.3.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 538, in main
    return gyp_main(args)
  File "/usr/local/n/versions/node/10.3.0/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 523, in gyp_main
    generator.GenerateOutput(flat_list, targets, data, params)
[...]

PR #1455, opened 2018-06-05
gyp: fix regex to match multi-digit versions

jhermsmeier:
##### Checklist

- [ ] tests are included 

##### Description of change

This fixes the regular expression matching in `xcode_emulation`
to also handle version numbers with multiple-digit major versions
which would otherwise break under use of XCode 10

Fixes: https://github.com/nodejs/node-gyp/issues/1454

### DIFF
--- a/gyp/pylib/gyp/xcode_emulation.py
+++ b/gyp/pylib/gyp/xcode_emulation.py
@@ -1262,7 +1262,7 @@ def XcodeVersion():
   except:
     version = CLTVersion()
     if version:
-      version = re.match(r'(\d\.\d\.?\d*)', version).groups()[0]
+      version = re.match(r'(\d+\.\d+\.?\d*)', version).groups()[0]
     else:
       raise GypError("No Xcode or CLT version detected!")
     # The CLT has no build information, so we return an empty string.


jhermsmeier:
There currently aren't any tests for this particular part of the code – should I add some? If so, I think I could do with some guidance on that

jhermsmeier:
Oh, I see, of course – not sure where my head was. I guess it would be preferable to land this upstream (if it hasn't been patched there yet)?

bnoordhuis:
Ideally, yes, but Google has effectively abandoned gyp so we're in this awkward situation where we're the de facto (but not de jure) upstream.

Trott:
I've confirmed that this change along unbreaks things on macOS High Sierra with CLI tools. Can we expedite landing and publishing?

bnoordhuis:
Merged in 293092c (commit 40f24e4 on the v3.x branch) and released as v3.6.3.  Thanks again, Jonas.

jhermsmeier:
Sweet, thank you for the swift response!

questsin:
still an issue

nornagon:
Hm, I'm not sure this has fully addressed the issue. On my system, `XcodeVersion()` is returning `"0101"`, which causes gyp to assume we want to target i386 (among other poor assertions). It seems like this only fixed the `CLTVersion` path, not the `xcodebuild -version` path?

cclauss:
@nornagon Would it be possible to add a new Travis CI run which mimics your Mac setup?
https://github.com/nodejs/node-gyp/blob/master/.travis.yml#L10-L14

PR #1458, opened 2018-06-08
Remove unused gyp test scripts.

bnoordhuis:
Shrinks node-gyp's size by about 100 kB.


bnoordhuis:
> Do we think we'll ever take another upstream gyp?

Unlikely in the extreme.

PR #1471, opened 2018-06-14
update request to 2.87.0

Rohithzr:
updated request to include security fixes and make it same as npm current
Passes all tests.


refack:
Hello @Rohithzr and thank you for the contribution!

The only issue this might have is that it's incompatible with node <= 4. We planned on releasing a `node-gyp@4.0.0` that will require node@4, so generally this might need to wait before we land this in `master`.
Or bump the node version as well in:
https://github.com/nodejs/node-gyp/blob/50b8734a7f79c57603d4d48cb3337713de404384/package.json#L37-L39

Let's see if our CI is properly setup - https://ci.nodejs.org/job/nodegyp-test-pull-request/54/

refack:
> The only issue this might have is that it's incompatible with node <= 4.

I guess I was wrong and `request>=2.82.1` is only incompatible with node __<__ 4, so we're all good here.

richardlau:
@refack Ben split off a a [`v3.x`](https://github.com/nodejs/node-gyp/tree/v3.x) branch to continue v3.x releases. For `node-gyp@4.0.0` we might as well drop Node.js v4 as well given that it is EOL.

Rohithzr:
Ah! I actually tested it with node 4.x which is still used somewhere in our legacy code. didn't give me any issues. Thanks for confirming it.

package.json
```diff
@@ -28,7 +28,7 @@
     "nopt": "2 || 3",
     "npmlog": "0 || 1 || 2 || 3 || 4",
     "osenv": "0",
-    "request": "2",
+    "request": "^2.86.0",
```
nschonni: Although the version should match, maybe use 2.87.0 since they removed the heok/hawk issues https://github.com/request/request/blob/master/CHANGELOG.md#v2870-20180521

xzyfer:
FWIW we did this update in node-sass@4.9.1 to resolve the security warnings and maintains support for Node 0.10. It was a smooth update.

Fishrock123:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/65/

Rohithzr:
@Fishrock123 so it works here without a doubt. How about we merge this and also figure out the 3.x.

PR #1481, opened 2018-06-21
escaping Dependency FilePaths to allow spaces

MylesBorins:
it is possible that there will be spaces in the path name
and that currently causes things to not work

can work out a test if this is considered a reasonable fix

/cc @fde31 who authored the patch

### DIFF
--- a/gyp/pylib/gyp/generator/make.py
+++ b/gyp/pylib/gyp/generator/make.py
@@ -601,6 +601,11 @@ def EscapeShellArgument(s):
   return "'" + s.replace("'", "'\\''") + "'"
 
 
+def EscapeFilePath(s):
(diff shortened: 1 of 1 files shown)


fde31:
thanks for following up on this! If I'm not mistaken that has been actually fixed in the meantime by introducing `SourceifyAndQuoteSpaces` in 3.7.0 of node-gyp via the following commit:

https://github.com/nodejs/node-gyp/commit/0540e4ec63257f56b218ee3c5dcd864aa8300dd0

which looks very similar to the patch of this PR.

MylesBorins:
Ah cool. closing 😊

MylesBorins:
Sent a patch upstream to node to float this patch on the npm tree

https://github.com/nodejs/node/pull/21448

PR #1484, opened 2018-06-25
deps: unpin `request` version range

iarna:
Pinning makes deduping in npm impossible.


bnoordhuis:
See #1300, we're not quite ready to let go of old Node.js versions just yet in v3.x.  I could do a v4.0.0 release from the master branch that you could bundle with npm, that has an unpinned request.

Issue #1439, opened 2018-05-14
Security issue in dependency


florianbrinkmann:
Hi,

there is a security issue ([CVE-2018-3728](https://nvd.nist.gov/vuln/detail/CVE-2018-3728)) in hoek, that is needed by hawk, and hawk is a dependency of request (it is the same issue like node-sass has in the current version https://github.com/sass/node-sass/issues/2355).

Could you increase the version of the `request` dependency accordingly?

Thanks in advance,
Florian

bnoordhuis:
> Could you increase the version of the request dependency accordingly?

In a word: no.  node-gyp depends on `"request": "2"` in order to share a copy with npm.  The common case is that node-gyp is bundled with npm.

florianbrinkmann:
Did not notice that, thanks for clarification!

cyrilchapon:
Hey. Landed here after going upward the chain hoek hawk request.

What do you mean "share a copy with npm" ?

Why was this closed, since there is still a potential vulnarability ?

cyrilchapon:
One should get inspiration on that kind of issue. https://github.com/sass/node-sass/issues/2355

Notice the "leave open" stuff, and great documentation about node.js version breaking

cyrilchapon:
Found https://github.com/sass/node-sass/issues/2355 that should be referenced here.

This issue is poorly documenting.

bnoordhuis:
You could have done a quick search, you know.  This isn't the first issue about it.

> What do you mean "share a copy with npm" ?

What exactly is unclear?

cyrilchapon:
> You could have done a quick search, you know. This isn't the first issue about it.

Yes I did. I think I can argue this is the one and only issue in this repository about this.

But if you found another one, I think any information about a "duplicate" would be way more useful than a silent close.

The whole point of `npm audit` is to track down security issues on a dependency chain. When running `npm audit`, node-gyp pops out, and it says "it has a security issue".

Well I think leaving an open issue for a security issue sounds fair, precisely since a security issue is actually remaining in the package. Event if it's waiting for an obscure deep dependency update.

> > What do you mean "share a copy with npm" ?

> What exactly is unclear?

I can't understand what you don't find "unclear" with it.
- A *copy* of what ?
- Why do you *need* to share this ?
- What do you mean, *share* ?

bnoordhuis:
> Yes I did. I think I can argue this is the one and only issue in this repository about this.

[...]

PR #1488, opened 2018-06-29
Getting Prototype Pollution errors and Upgrading dependencies seems to help

jrgleason:
Simple edits to the package.json but seem to help some of the node warnings in v11


bnoordhuis:
See #1439, we can't make this change in node-gyp@v3.x but we'll upgrade once we release node-gyp@4.  Thanks anyway.

Issue #1501, opened 2018-07-14
<(target_arch) not resolving, causing npm install to fail


yeerkkiller1:
* **Node Version**: 
node: v10.6.0, npm: 6.1.0
* **Platform**: 
MSYS_NT-10.0 DESKTOP-E7DLHGV 2.8.2(0.313/5/3) 2017-07-12 15:35 x86_64 Msys
* **Compiler**: 
Microsoft (R) Build Engine version 15.6.85.37198 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.
15.6.85.37198
Microsoft (R) C/C++ Optimizing Compiler Version 19.13.26131.1 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]
* **Module**: 
https://github.com/nodejs/abi-stable-node-addon-examples/blob/master/1_hello_world/napi/hello.cc

<details>
<summary>
npm install
</summary>

```
C:\Users\quent\Dropbox\compile-test>if not defined npm_config_node_gyp (node "C:\Program Files\nodejs\node_modules\npm\node_modules\npm-lifecycle\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild )  else (node "C:\Program Files\nodejs\node_modules\npm\node_modules\node-gyp\bin\node-gyp.js" rebuild )
The system cannot find the path specified.
gyp ERR! configure error
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (C:\Program Files\nodejs\node_modules\npm\node_modules\node-gyp\lib\configure.js:336:16)
gyp ERR! stack     at ChildProcess.emit (events.js:182:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:237:12)
gyp ERR! System Windows_NT 10.0.17134
gyp ERR! command "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\node_modules\\node-gyp\\bin\\node-gyp.js" "rebuild"
gyp ERR! cwd C:\Users\quent\Dropbox\compile-test
gyp ERR! node -v v10.6.0
gyp ERR! node-gyp -v v3.6.2
gyp ERR! not ok
npm ERR! code ELIFECYCLE
npm ERR! errno 1
[...]

PR #1506, opened 2018-07-19
configure.js: escape '<'

yeerkkiller1:
Fixes: https://github.com/nodejs/node-gyp/issues/1501

On windows invoking spawn on a batch file results in additional
argument processing. Special characters need to be escaped (twice).

##### Checklist

- [ ] tests are included 

##### Description of change

Fixes issue where npm install fails if python is a batch file. This occurs because the batch file tries to interpret the '<' specially.

I tested this change on my windows machine (with and without python.bat in my path). I don't think this change will impact other platforms, as it only applies changes when the platform is win32, and when python is a batch file (which would have broken anyway).

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -273,9 +273,16 @@ function configure (gyp, argv, callback) {
         // Windows expects an absolute path
         output_dir = buildDir
       }
+      
+      var target_arch_gyp_variable = '<(target_arch)'
+      if (process.platform === 'win32' && python.toLowerCase().endsWith('.bat')) {
+        // Windows will interpret python as a batch file, so we need to escape the '<' character
+        target_arch_gyp_variable = '^^^<(target_arch)'
+      }
+      
       var nodeGypDir = path.resolve(__dirname, '..')
       var nodeLibFile = path.join(nodeDir,
-        !gyp.opts.nodedir ? '<(target_arch)' : '$(Configuration)',
+        !gyp.opts.nodedir ? target_arch_gyp_variable : '$(Configuration)',
         release.name + '.lib')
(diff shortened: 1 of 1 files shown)


Fishrock123:
CI https://ci.nodejs.org/job/nodegyp-test-pull-request/66/

rvagg:
I'm not confident enough to pull this in atm, does someone else with deep enough expertise want to confirm this is the right thing to do?

rvagg:
@yeerkkiller1 if you're still interested, would you mind rebasing and squashing your two commits?

yeerkkiller1:
@rvagg, no problem, it's squashed and pushed now.

rvagg:
SGTM but I would love a +1 from @bzoz, @joaocgreis, @refack or one of our other core Windows experts, I've never had experience escaping in .bat with `^` (I had to Google it just now to see it was a thing!).

jlennox:
@rvagg This is indeed the correct fix.

I just spent an hour debugging this issue myself. It would be great if this fix would be merged.

rvagg:
@bzoz, @joaocgreis, please confirm

bzoz:
I think this is not needed anymore. https://github.com/nodejs/node-gyp/pull/1582 made a change that extracts the Python .exe filename even from batch files and then uses it. We are no longer using the batch file to spawn Python executables.

I've also was not able to trigger the original issue. @yeerkkiller1, @jlennox could you verify that the issue is still present in master?

rvagg:
#1582 went out with v5.0.0 so should be in npm already, maybe update npm: `npm install npm -g`

PR #1534, opened 2018-08-27
lib: use print() for python version detection

greenaddress:
This fixes an issue introduced by commit https://github.com/nodejs/node-gyp/commit/da5db4059f9d3628a9fcace262c482d585460002 specifically this line https://github.com/nodejs/node-gyp/commit/da5db4059f9d3628a9fcace262c482d585460002#diff-7ad6309831a48a234780b58baca65710R440

Namely, the bit that checks for python versions was changed and went from supporting both python2 and python3 to supporting only python2.

I understand there are other open issues/PRs in regards to python3 but this change specifically is a regression that broke windows builders with python3 in the path as python.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -437,7 +437,7 @@ PythonFinder.prototype = {
   },
 
   checkPythonVersion: function checkPythonVersion () {
-    var args = ['-c', 'import sys; print "%s.%s.%s" % sys.version_info[:3];']
+    var args = ['-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);']
     var env = extend({}, this.env)
     env.TERM = 'dumb'
 


greenaddress:
@refack done, thank you for review!

refack:
@greenaddress thank you for following up. We'll land this soon, then we need to release a new version and get that landed in `npm` so it might take some time for this to get to the public.

jkoorts:
Still happening 5 months later..

Fylipp:
Since this seems to be a very common error (countless forum posts and issues / who has Python 2 installed nowadays?) this should have been integrated into a package release by now. The last NPM release of `node-gyp` was 6 months ago. A small release containing this change would increase user friendliness considerably.

mittalyashu:
I am still facing this issue, when this bug fix will be released.

mittalyashu:
image

The last time an update was release 7 months ago, can you release the a new update with all the bug fixes

/cc @refack @hashseed @richardlau

richardlau:
I don't have permissions to publish this to npm.

mittalyashu:
😮 Oh! Okay

steveoh:
What versions did this go into? I'm on 10.16 and 6.10 and seeing this message?

richardlau:
> 
> 
> What versions did this go into? I'm on 10.16 and 6.10 and seeing this message?

`node-gyp` 4: https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#v400-2019-04-24

PR #1164, opened 2017-04-04
Prefix build targets with /t: on Windows

NatalieWolfe:
Currently, on non-Windows platforms, it is possible to have a multi-target native module and specify building just some of the targets using `node-gyp build my_target`. On Windows, however, specifying the target(s) to build requires the [`/t:` or `/target:` flag](https://msdn.microsoft.com/en-us/library/ms164311.aspx). Without this change you will get the error `MSB1008` because MSBuild things you are trying to specify multiple solutions.

PR #1540, opened 2018-09-12
Allow passing defines such as -- -DFOO=bar

ilatypov:
##### Checklist

- [ ] tests are included 

##### Description of change

My small change in node-pre-gyp hit a wall because `node-gyp configure .. -- -DFOO=bar` results in creating an msbuild option `/t:-DFOO=bar`, thank you very much.

https://github.com/mapbox/node-pre-gyp/pull/416


ilatypov:
Sorry I do not know how to run tests.  I see some added in the change that exposed my issue.

https://github.com/nodejs/node-gyp/pull/1164

refack:
Hello @ilatypov and thank you for the contribution.
Could you write a test for this change? An example can be:
https://github.com/nodejs/node-gyp/blob/aebdd7a710862c13592a9d2c8a28aadf661b0542/test/test-addon.js#L34-L48
(BTW: you run the test suite with a simple `npm install; npm test`)

kewde:
Hmm, this is an issue that does not only affect the the configure options but also overloading of properties.
```
/p:FOO=bar
```
becomes
```
/t:/p:FOO=bar
```

This fix wouldn't apply in all cases :/

rvagg:
@ilatypov would still like to pursue this? if so, please rebase to the current master and let's see if we can figure out a test for it.

cclauss:
Too many conflicts.

PR #1548, opened 2018-09-19
Corrected npm set msvs version

synthstow:
corrected * Launch cmd, `npm config set msvs_version 2017`
to: * Launch cmd, `npm config set msvs_version=2017`
Because that '=' sets the version to "2017" rather than ""

##### Checklist

- [ X] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

Corrected an incorrect settings command in the readme


README.md
```diff
@@ -51,7 +51,7 @@ Install tools and configuration manually:
    (using "Visual C++ build tools" workload) or [Visual Studio 2017 Community](https://visualstudio.microsoft.com/pl/thank-you-downloading-visual-studio/?sku=Community)
    (using the "Desktop development with C++" workload)
    * Install [Python 2.7](https://www.python.org/downloads/) (`v3.x.x` is not supported), and run `npm config set python python2.7` (or see below for further instructions on specifying the proper Python version and path.)
-   * Launch cmd, `npm config set msvs_version 2017`
+   * Launch cmd, `npm config set msvs_version=2017`
```
refack: If I'm not mistaken current version is right https://docs.npmjs.com/cli/config

synthstow:
Maybe throw in a note because it didn't work for me and after a bunch of
Google searches I found I needed the modification

Anyway it's working well now. Cheers for the work

rvagg:
both versions of that command should be doing the same thing, if not then this is a problem for npm

check your ~/.npmrc (or \Users\whatever\.npmrc or wherever it's found on Windows) when performing both commands

PR #1549, opened 2018-09-21
Allows GYP_MSVS_VERSION=2017

corylulu:
I am pretty ignorant to node-gyp, but I couldn't get a project to build because it was using 2017 build tools, despite everyone saying that 2017 tools were supported in 3.6.0. I kept getting an error thrown here. So I modified the source and got it to work. 

If there was a reason this shouldn't have been updated, then ignore this PR, but otherwise, I thought I'd share.


richardlau:
@refack It seems as though these changes are part of https://github.com/nodejs/node/commit/3a334b12807eae369d487e0dce594681661fdcc5#diff-644e93cb65b1ec5c94e03651ef10be26 over in the main Node.js repository. Do we want to do a similar base gyp update here in node-gyp?

joaocgreis:
@corylulu thanks for submitting this! With the current code in node-gyp, gyp should never see a `GYP_MSVS_VERSION=2017` because if a usable VS2017 installation is found it is always set to 2015.

However, it seems your installation works but node-gyp does not detect it. Can you run the released version of node-gyp with `--verbose` (or npm with `-dd`) and past the output here? At least the lines where it looks for VS2017.

@richardlau @refack the method we use here to detect VS2017 is better than what landed in gyp or node-core. This allows for only the exact needed components to be installed (not the entire workload), and does it using the recommended method. I'd rather continue to use this, or land this in gyp if we start maintaining gyp in the org.

refack:
> I'd rather continue to use this, or land this in gyp if we start maintaining gyp in the org.

@joaocgreis ack. I was referring to the changes in GYP that allow it to understand `GYP_MSVS_VERSION=2017`. I agree that the detection logic is a different issue.

bpasero:
I (and everyone trying to build VSCode) is also suffering from this. Steps to reproduce:
* install node.js `8.12.0`
* install `windows-build-tools` (from a PowerShell session with admin rights: `npm install --global windows-build-tools`)
* have a `package.json` that depends on `vscode-sqlite3`
* `npm install`

=> `error MSB8020: The build tools for v140 (Platform Toolset = 'v140') cannot be found. To build using the v140 build tools, please install v140 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution"`

Ideally I can teach `npm install` to use my existing 2017 installation.

rvagg:
this doesn't seem to be needed .. ? if I'm incorrect, can a collaborator who groks this better than me please reopen it. It also needs a rebase if @corylulu wants to continue pursuing it.

PR #1553, opened 2018-09-25
Update configure.js

killgxlin:
If i set msvs_version to number other than 2017 in my npm config. It will try to use vs2017.

##### Checklist

- [ ] tests are included 

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -83,7 +83,7 @@ function configure (gyp, argv, callback) {
     mkdirp(buildDir, function (err, isNew) {
       if (err) return callback(err)
       log.verbose('build dir', '"build" dir needed to be created?', isNew)
-      if (win && (!gyp.opts.msvs_version || gyp.opts.msvs_version === '2017')) {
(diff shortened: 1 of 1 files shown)


refack:
Hello @killgxlin and thank you for the PR.

I'm not sure I understand your fix.
In the current code:
```
(!gyp.opts.msvs_version || gyp.opts.msvs_version === '2017')
```
if `gyp.opts.msvs_version=2015` then
`!gyp.opts.msvs_version === false` and `(gyp.opts.msvs_version === '2017') === false`
So the if branch should not be taken

The test you suggest:
```
(!gyp.opts.msvs_version && gyp.opts.msvs_version === '2017')
```
Can never be true, since if `(gyp.opts.msvs_version === '2017') === true` then `Boolean(gyp.opts.msvs_version) === true` so the conjunction can only be false.

killgxlin:
@refack Sorry, for my carelessness:-p

refack:
@killgxlin no worries. Thank you for the effort, it is appreciated!

PR #1566, opened 2018-10-12
win: fix delay-load hook for electron 4

adill:
##### Checklist

##### Description of change
Electron, starting with major version 4, ships a `node.lib` import library referencing `electron.exe` (`node` is statically linked) so it is no longer possible to freely rename `electron.exe` and have native modules work out of the box.

Is `node-gyp` interested in a change like this? or is this be something that the Electron team should work around themselves?

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -1,6 +1,7 @@
 {
   'variables' : {
     'node_engine_include_dir%': 'deps/v8/include',
+    'node_host_binary%': 'node'
   },
   'target_defaults': {
     'type': 'loadable_module',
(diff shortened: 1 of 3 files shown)


refack:
> Is `node-gyp` interested in a change like this? or is this be something that the Electron team should work around themselves?

@adill seems like a valid use case to me.

If you are up to the challenge, could you parameterize this? It could be something like `<(node_core_target_name)<(EXECUTABLE_SUFFIX)` in the .gyp files. Hence
```python
  'defines': [ 'CORE_EXE_NAME=electron.exe', ],
```
and then the change is the .cc file would be:
```c++
  if (_stricmp(info->szDll, CORE_EXE_NAME) != 0)
    return NULL;
```

refack:
@adill yes. Just need to verify two things:
1. the this works as expected https://github.com/nodejs/node-gyp/blob/7a53da634341996716a9ac5e27859b0bdb4f6993/addon.gypi#L66
  I'm need to verify GYP will expanad the variable in this case
2. That this in fact solves the binary-was-renamed issue

richardlau:
> 
> 
> note: As implemented anyone still using `iojs.exe` would be broken by this change.

I don't think that will be an issue if we release in node-gyp@4 (where we're dropping support for EOL versions of Node.js/io.js).

refack:
> I've tested locally and verified that it works as expected. (I used `cld` as my test native module.)

@adill Could you write a test for this? I'm thinking something like:
1. There's a sample addon in https://github.com/nodejs/node-gyp/tree/master/test/node_modules/hello_world.
2. Copy `process.execPath` to a temp directory, then call `tmp1exe nodeGyp` on it and rename the exe
https://github.com/nodejs/node-gyp/blob/2394a1eca27bfc99e03f899034783287f17c6b08/test/test-addon.js#L8

adill:
@refack Added a test. Seems to work -- passes per usual, fails when I make the delay load hook return `nullptr`.

refack:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/102/

@khanhas, if this passes CI I'll land this, but a release is not ready yet. Hopefully this will be release soon, and quickly adopted by `npm` for an update.

PR #1573, opened 2018-10-14
meta: add to .gitignore

refack:
##### Checklist

- [X] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)


.gitignore
```diff
@@ -1,3 +1,5 @@
 gyp/test
 node_modules
 test/.node-gyp
+.ncu
+package-lock.json
```
marpme: I would actually commit the package-lock.json
see: https://stackoverflow.com/a/44210813

rvagg:
landed in d3478d7

.gitignore
cinderblock: Why is this ignored?

doronbehar: > Why is this ignored?

I too wonder why is this ignored... cc @refack ...

PR #1582, opened 2018-10-26
Clean-up python detection

joaocgreis:
This changes Python detection to make it try a sequence of possibilities. Every possibility will be tried discarding errors along the way, only giving up when there's nothing else to try. This should make Python detection more robust.

This stops using `which` to find the executable, instead relying on the mechanism that was already used for the py launcher: run Python and print `sys.executable`. This means that BAT files can be used, making node-gyp work when depot_tools is the first thing in the path.

image

@refack I based this on your commit from https://github.com/nodejs/node-gyp/pull/1269 . Ended up changing much of it, but I used some of the info text so I include your commit for proper credit. Let me know if I should remove it or if you want to land https://github.com/nodejs/node-gyp/pull/1269 separately. I rebased it on master and included `es6: true` in `.eslintrc` to make linting work (since ES6 was already there for `.jshintrc`).

Error output when no Python is installed:
image

##### Checklist

- [x] tests are included 

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,7 +1,7 @@
 {
   "asi": true,
   "laxcomma": true,
-  "es5": true,
+  "esversion": 6,
   "node": true,
   "strict": false
 }
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,5 +1,6 @@
 ---
   env:
     node: true
+    es6: true
   rules:
     no-unused-vars: error
--- a/bin/node-gyp.js
+++ b/bin/node-gyp.js
@@ -79,8 +79,12 @@ function run () {
   prog.commands[command.name](command.args, function (err) {
     if (err) {
       log.error(command.name + ' error')
-      log.error('stack', err.stack)
-      errorMessage()
+      if (err.noPython) {
+        log.error(err.message)
+      } else {
+        log.error('stack', err.stack)
+        errorMessage()
+      }
       log.error('not ok')
       return process.exit(1)
     }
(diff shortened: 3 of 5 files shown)


joaocgreis:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/112/

joaocgreis:
The code here will try all the mechanisms for detecting Python. It will produce log messages at `verbose` level as it goes. However, it will also save the most relevant ones in an array. If detecting Python fails in the end, it will print those messages at `error` level. This should make issues finding Python much easier to understand, both for the users and for us here.

cc @nodejs/node-gyp

.jshintrc
```diff
@@ -1,7 +1,7 @@
 {
   "asi": true,
   "laxcomma": true,
-  "es5": true,
+  "esversion": 6,
```
richardlau: This is fine for master but shouldn't be backported to 3.x (do we have the equivalent of core's dont-land labels)?

joaocgreis: I don't see labels for this. Didn't we already do the final v3 release? I was expecting the next one to be v4 only, though we could do v3 if there is a pressing need of some sort.

This would always be semver-major defensively. Because this is written using new javascript features and removes workarounds for old node versions, this is effectively semver-major.

joaocgreis:
Thanks @rvagg! I will land this soon if there is no objection.

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/commit/a6990ada1f1147cb6da7ceaa0bb7c8333c9c7ae6

PR #1603, opened 2018-11-13
Raise the lower limit on Python version to 2.7.9

cclauss:
Python 2.6 end of life was more than five years ago and Python < 2.7.9 have ___known security flaws___ and therefore should not be trusted.

##### Checklist

- [ ] tests are included 

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -372,7 +372,7 @@ PythonFinder.prototype = {
   log: logWithPrefix(log, 'find Python'),
   argsExecutable: ['-c', 'import sys; print(sys.executable);'],
   argsVersion: ['-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);'],
-  semverRange: '>=2.6.0 <3.0.0',
+  semverRange: '>=2.7.9 <3.0.0',
 
   // These can be overridden for testing:
   execFile: cp.execFile,


refack:
@cclauss thanks for the submission.

The trouble is we still support platforms that don't have python2.7 readily available i.e. centos https://gist.github.com/dalegaspi/dec44117fa5e7597a559

cclauss:
@refack The tread rightly points to https://gist.github.com/crayzeewulf/bcb5dcec03cd0dd03f3622abf7b6cec4 as the modern approach.

joaocgreis:
While we shouldn't recommend anything but the latest version of Python, I don't see a reason to break compatibility for end users.

It's not so much about what we use in CI or what systems we support (though I don't know all in detail and there might be a case to be made there somewhere), but about end users running Python 2.6.0 at their own risk. I don't think we have a strong case for breaking that.

That said, if there is a good reason to break compatibility, now is a good time as node-gyp is about to get a semver-major bump. However, it makes more sense to me to wait for Python 3 and break Python 2 all in one go.

cclauss:
We ___need___ Python 2 and Python 3 to be running in parallel.  This is best practice for all significant codebases so that A/B comparisons can be done to ensure equivalent functionality.  

Python 2.6 end of life was more than five years ago and https://snarky.ca/stop-using-python-2-6 was written three years ago.  It contains strong reasons for breaking compatibility and Python < 2.7.9 have known security flaws and therefore should not be trusted.

cclauss:
Python 2.6 is an attack vector...  If we really care about _our_ users we should not encourage them to continue being easy targets for cyber attacks.  🎄

PR #1572, opened 2018-10-14
Fix the remaining undefined names

cclauss:
@refack This gets the codebase to pass:
* python2 -m --count --show-source --statistics --select=E901,E999,F821,F822,F823
* python3 -m --count --show-source --statistics --select=E901,E999,F821,F822,F823

@refack Your review please?

PR #1612, opened 2018-11-26
Define basestring, cmp(), unicode in xcodeproj_file.py

cclauss:
__basestring__, __cmp()__, and __unicode__ were all removed in Python 3 so this PR conditionally defines these terms to provide equivalent functionality in both Python 2 and Python 3.

Was part of #1572


gyp/pylib/gyp/xcodeproj_file.py
```diff
@@ -155,6 +155,23 @@
   _new_sha1 = sha.new
 
 
+try:
+    basestring     # Python 2
+except NameError:  # Python 3
+    basestring = (str, )
```
browniebroke: I was going to suggest to re-implement it as in the [six library](https://github.com/benjaminp/six/blob/master/six.py#L35-L53) but it requires a lot more changes to this file, wherever `basestring` is used. After a second thought, I think it's fine like this.

gyp/pylib/gyp/xcodeproj_file.py
```diff
@@ -324,8 +341,7 @@ def Copy(self):
           that._properties[key] = new_value
         else:
           that._properties[key] = value
-      elif isinstance(value, str) or isinstance(value, unicode) or \
-           isinstance(value, int):
+      elif isinstance(value, (basestring, int)):
```
browniebroke: I was worried that it might break if `basestring` is a tuple, but I just tested in Python 3.6 and it works 👍

PR #1614, opened 2018-11-26
reload() and sys.setdefaultencoding() were removed in Python 3

cclauss:
__reload()__ was [moved](https://docs.python.org/3/library/importlib.html#importlib.reload) and __sys.setdefaultencoding()__ was removed in Python 3.

Was part of #1572


test/fixtures/test-charmap.py
```diff
@@ -2,14 +2,21 @@
 import sys
 import locale
 
-reload(sys)
+try:
+  reload(sys)
+except NameError:
+  pass
```
browniebroke: I believe this `pass` statement isn't the right equivalent. It looks like it has been moved to `imp` in Python 3, and then from `imp` to `importlib` in Python 3.4, according to [this page](https://docs.python.org/3/whatsnew/3.4.html#importlib):

> The `reload()` function has been moved from `imp` to `importlib` as part of the `imp` module deprecation.

Given that [Python 3.3 has reached EOL more than a year ago](https://www.python.org/downloads/release/python-337/), I think it's fine to target 3.4+ here:

```python
from importlib import reload
reload(sys)
```

test/fixtures/test-charmap.py
```diff
@@ -2,14 +2,21 @@
...
   if not encoding:
     return False
 
-  sys.setdefaultencoding(encoding)
+  try:
+    sys.setdefaultencoding(encoding)
+  except NameError:
+    pass
   textmap = {
     'cp936': u'\u4e2d\u6587',
```
browniebroke: Could we replace these `u'...'` strings by regular strings & use the `unicode_literals` future import instead?

test/fixtures/test-charmap.py
cclauss: Line 6 is for line 17...  Given that __sys.setdefaultencoding()__ was removed in Python 3 (because __utf-8__ is already the default), line 6 does not really help us.

PR #1615, opened 2018-11-26
Builtin cmp() was removed in Python 3

cclauss:
__cmp()__ was removed in Python 3 so this PR conditionally defines __cmp()__ to provide equivalent functionality in both Python 2 and Python 3.  Also removed a legacy Python <= 2.5 workaround.

Was part of #1572


cclauss:
@refack Could I please get some reviews on https://github.com/nodejs/node-gyp/pulls/cclauss

gyp/pylib/gyp/MSVSNew.py
```diff
@@ -4,22 +4,17 @@
...
-  import hashlib
-  _new_md5 = hashlib.md5
-except ImportError:
-  import md5
-  _new_md5 = md5.new
-
+    cmp             # Python 2
+except NameError:
+    def cmp(x, y):  # Python 3
+        return (x > y) - (x < y)
```
bnoordhuis: Since `cmp` is only used in one place in this file, I'd just open-code it there. Easier to understand and maintain than version-specific branching.

gyp/pylib/gyp/MSVSNew.py
```diff
@@ -68,8 +63,7 @@ def __cmp__(self, other):
 class MSVSFolder(MSVSSolutionEntry):
   """Folder in a Visual Studio project or solution."""
 
-  def __init__(self, path, name = None, entries = None,
-               guid = None, items = None):
+  def __init__(self, path, name=None, entries=None, guid=None, items=None):
```
bnoordhuis: Stylistic change, please undo. Same goes for the whitespace changes.

gyp/pylib/gyp/MSVSNew.py
```diff
@@ -59,10 +49,11 @@ def MakeGuid(name, seed='msvs_new'):
 #------------------------------------------------------------------------------
 
 
+@total_ordering
 class MSVSSolutionEntry(object):
-  def __cmp__(self, other):
+  def __lt__(self, other):
```
browniebroke: Sorry if that's a dumb question, but should we also define `__eq__` and `__ne__` as per the page you linked to? 

> To avoid the hassle of providing all six functions, you can implement `__eq__`, `__ne__`, and only one of the ordering operators, and use the `functools.total_ordering()` decorator to fill in the rest.

Issue #1653, opened 2019-01-22
node-gyp gets confused by MSBuild on the %PATH%


twoi:
* **Node-gyp Version**: v3.8.0
* **Node Version**: v11.7.0
* **Npm Version**: 6.5.0
* **Platform**: Windows 8.1 Enterprise / 6.3.9600 N/A ビルド 9600 / x64-based PC

I tried to build build a simple module on msvs 2017, but that failed as long as an old MSBuild
c:\Program Files (x86)\MSBuild\14.0\Bin\amd64\
was on the %PATH%

When I removed the path, `node-gyp build` built just fine. It also worked, when I added the latest MSBuild to the %PATH%
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\amd64

So `node-gyp configure` doesn't seem to care if there is an MSBuild.exe on the path, but `node-gyp build` stumbles upon an insufficient MSBuild on the path, although it does work fine if there is none at all on the path.

As this took me hours to figure out, I would suggest `node-gyp build` just ignore any MSBuild on the path. (Otherwise, what's the point of `node-gyp configure` going to great pains to find a suitable MSBuild in the first place?)

Judging from stackoverflow and github issues, i'm not the only one stumbling on this. In fact, several other developers had "success" after uninstalling (!!) their old Visual Studios - I'm assuming what cured it was that the uninstall also fixed their %PATH%.

<details><summary>Verbose output:</summary>

```
C:\Projects\node>node-gyp configure --verbose --target=v0.12.18 --msvs_version=2017
[...]
gyp verb find vs2017 Found installation at: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise
gyp verb find vs2017   - Found Microsoft.VisualStudio.Component.Windows10SDK.14393
gyp verb find vs2017   - Found Microsoft.VisualStudio.Component.Windows10SDK.17763
gyp verb find vs2017   - Found Microsoft.VisualStudio.Component.VC.Tools.x86.x64
gyp verb find vs2017   - Found Microsoft.VisualStudio.VC.MSBuild.Base
gyp verb find vs2017   - Using this installation with Windows 10 SDK
gyp verb find vs2017 using installation: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise
[...]
gyp info ok

C:\Projects\node>node-gyp build
gyp info it worked if it ends with ok
gyp info using node-gyp@3.8.0
[...]

PR #1654, opened 2019-01-22
win: use msbuild from the configure stage

bzoz:
##### Checklist

##### Description of change
If node-gyp configure has set up MSBuild location use it instead the one that happens to be first on the PATH.

Fixes: https://github.com/nodejs/node-gyp/issues/1653

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -95,6 +95,13 @@ function build (gyp, argv, callback) {
    */
 
   function doWhich () {
+    // On Windows use msbuild provided by node-gyp configure
+    if (win && config.variables.msbuild_path) {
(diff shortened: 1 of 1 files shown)


refack:
@bzoz thanks for the fix. Could you add a test? (maybe run `node-gyp configure` then add a fake `msbuild.exe` to the path and run `node-gyp build`)

joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/115/

@refack do you mind if I land this even without the test? Change here is pretty straightforward and there was no test before.

refack:
@joaocgreis, done.

PR #1657, opened 2019-01-30
doc: add fix for msbuild errors on windows

bertyhell:
##### Checklist

##### Description of change
Add an extra step to fix an issue with node-gyp rebuild on windows. After trying 2 days to fix it any other way.
* tried installing build tools
* tried reinstalling node-gyp
* tried env variables for python and msbuild
* tried installing visual studio 2017
* tried reinstalling build tools
* tried removing node and all node_modules and reinstalling everything

The step that I added was a last resort effort before I was going to format my pc to see if that fixes it.

I kept getting error messages like:
* error MSB4019: The imported project "C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.Cpp.Default.props" was not found. Confirm 
that the path in the <Import> declaration is correct, and that the file exists on disk
* C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.Cpp.Platform.targets(57,5): error MSB8020: The build tools for v141 (Platform Toolset = 'v141') cannot be found
* error MSB4062: The "SetEnv" task could not be loaded from the assembly C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.Build.CppTasks.Common.dll
*  [SetupUpdaterImpl]: Stream was closed
* node-gyp rebuild failed with: Error: Command failed: node-gyp rebuild


joaocgreis:
This should not be fixed by adding instructions to install an outdated version of Visual Studio, but rather by understanding what went wrong and adding a better error message for it. Visual Studio 2017 is fully supported (provided there is no incompatibility in the modules' code) and should be used.

@bertyhell can you paste the full output of node-gyp (or npm) when run with `--verbose`? It should indicate clearly why it is not using Visual Studio 2017, or provide clues to what the error is. Also, thanks for trying to make node-gyp better and actually submitting this PR!

rvagg:
@bzoz @joaocgreis see fa8b5c840be756b80a2adc17c78f00d4e7619465, it's just a doc change, is that a valid recommendation and worth putting in here?

PR #1670, opened 2019-02-14
Remove deprecated compatibility code.

bnoordhuis:
* Remove support for the IOJS_ORG_MIRROR, NVM_IOJS_ORG_MIRROR
  and NVM_NODEJS_ORG_MIRROR enviroment variables.

* Remove obsolete support for io.js. It's been out of support
  for over three years now.


test/docker.sh
```diff
@@ -3,7 +3,6 @@
 #set -e
 
 test_node_versions="0.8.28 0.10.40 0.12.7 4.3.0 5.6.0"
```
richardlau: Should we be updating these (perhaps in a different PR) to in support versions of Node.js?

bnoordhuis: Yep, good idea. I could do that as a separate commit or a separate PR.

richardlau: > 
> 
> Yep, good idea. I could do that as a separate commit or a separate PR.

Your choice 😄

bnoordhuis:
Updated after some delay... CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/116/

PR #1655, opened 2019-01-25
src: add arm64 to Windows native module compilation

jkunkee:
##### Checklist
- [ ] `npm install && npm test` passes
    - I'm still working on how to bootstrap NPM on a new architecture.
- [ ] tests are included
    - I don't see any relevant tests to update.
- [x] documentation is changed or added
- [x] commit message follows guidelines

##### Description
Node.js ships with a number of native modules used for testing. This change is the minimum required for those modules to node-gyp to be able to build them. All Node.js unit tests pass.

PR #1678, opened 2019-02-25
configure: Enable ARM64 single-toolchain build on Windows

kaadam:
Adds ARM64 targets to generated .sln/.vcxproj files for cross-compiling
node modules to Windows on Arm, 64-bit. To use, run
`set npm_config_arch=arm64` on the command prompt before `npm install`.

Change-Id: Idbe401437e58d13c815c02f7b5b45247daabbe65

##### Checklist

- [ ] tests are included

##### Description of change
This change is intent to enable cross-compiling for ARM64 Windows.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -134,6 +134,9 @@ function configure (gyp, argv, callback) {
 
     // set the target_arch variable
     variables.target_arch = gyp.opts.arch || process.arch || 'ia32'
+    if (variables.target_arch == 'arm64') {
+      defaults['msvs_configuration_platform'] = 'ARM64'
+    }
 
     // set the node development directory
     variables.nodedir = nodeDir


kaadam:
Related to: https://github.com/nodejs/node/issues/25998

gyp/pylib/gyp/generator/msvs.py
```diff
@@ -2651,9 +2653,17 @@ def _GetMSBuildGlobalProperties(spec, guid, gyp_file_name):
...
 
+  preferred_tool_architecture = None
+
   if os.environ.get('PROCESSOR_ARCHITECTURE') == 'AMD64' or \
      os.environ.get('PROCESSOR_ARCHITEW6432') == 'AMD64':
-    properties[0].append(['PreferredToolArchitecture', 'x64'])
+     preferred_tool_architecture = 'x64'
+
+  if os.environ.get('npm_config_arch') == 'arm64':
+    preferred_tool_architecture = 'ARM64'
```
jkunkee: This sets the PreferredToolArchitecture property in the vcxproj files. This property is [used to select either the x64 or x86 tool exe files to use during the build](https://docs.microsoft.com/en-us/cpp/build/msbuild-visual-cpp-overview?view=vs-2017#preferredtoolarchitecture-property).

For example, in a standard VS2017 Community installation with the ARM64 tools installed, there are six versions of cl.exe:

```cmd
c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\vc\Tools\MSVC\14.16.27023\bin\Hostx64\arm64\cl.exe
c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\vc\Tools\MSVC\14.16.27023\bin\Hostx64\x64\cl.exe
c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\vc\Tools\MSVC\14.16.27023\bin\Hostx64\x86\cl.exe
c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\vc\Tools\MSVC\14.16.27023\bin\Hostx86\arm64\cl.exe
c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\vc\Tools\MSVC\14.16.27023\bin\Hostx86\x64\cl.exe
c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\vc\Tools\MSVC\14.16.27023\bin\Hostx86\x86\cl.exe
```

This property sets the 'Hostx64' part of the path. Since there is no 'HostARM64' folder--that is, Visual Studio 2017 does not have a native ARM64 compiler--setting PreferredToolArchitecture to ARM64 is incorrect. When cross-compiling from x86 or x64 to ARM64, either the x86 or x64 tools should be used.

Note that if the host is ARM64, only the x86 tools will run (emulation). So [if `PROCESSOR_ARCHITEW6432` is set to ARM64](https://docs.microsoft.com/en-us/windows/desktop/WinProg64/wow64-implementation-details) then PreferredToolArchitecture should be x86.

gyp/pylib/gyp/msvs_emulation.py
```diff
@@ -519,7 +524,7 @@ def GetLibFlags(self, config, gyp_to_build_path):
     libflags.extend(self._GetAdditionalLibraryDirectories(
         'VCLibrarianTool', config, gyp_to_build_path))
     lib('LinkTimeCodeGeneration', map={'true': '/LTCG'})
-    lib('TargetMachine', map={'1': 'X86', '17': 'X64', '3': 'ARM'},
+    lib('TargetMachine', map={'1': 'X86', '17': 'X64', '3': 'ARM', '4': 'ARM64'},
```
jkunkee: It is worth noting that the enum from which these numbers are drawn, [machineTypeOption](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.vcprojectengine.machinetypeoption), does not have an official value for ARM64 yet. As I understand it this will be fixed in VS2017.

Please also note that the `link.exe` argument `/MACHINE:ARM64` *does* exist and works as expected.

gyp/pylib/gyp/MSVSNew.py
```diff
@@ -191,7 +191,8 @@ def __init__(self, path, version, entries=None, variants=None,
       self.variants = variants[:]
     else:
       # Use default
```
jkunkee: This is a little beyond my current GYP depth, so please forgive me if this is a dumb question: wouldn't it be better to have GYP pass in the correct Win32+ARM64 variants instead of emitting both even when not cross-compiling?

jkunkee:
It would also be nice if, once accepted, these changes somehow made it into the other two notable copies of GYP:

 - checked into Node.js https://github.com/nodejs/node/tree/master/tools/gyp
 - forked by @refack https://github.com/refack/GYP

refack:
Hello @kaadam could you submit the `GYP` only parts to https://github.com/refack/GYP. We have a much more extensive CI running there. It'll also be nice if you added some basic unit tests that cover this new feature.

gyp/pylib/gyp/generator/msvs.py
kaadam: Sorry for the delayed response. Thanks for clarifying this. I misunderstood that 'PreferredToolArchitecture' is. I will fix this based on your suggestion.

gyp/pylib/gyp/msvs_emulation.py
kaadam: Regarding to the number I suppose that it will be increased by one, considering the current biggest 'value' in the enum, so it might be '18' unofficially. That's interesting I found this '18' value for /MACHINE:ARM64 in [cmake repoitory](https://gitlab.kitware.com/cmake/cmake/blob/master/Source/cmLocalVisualStudio7Generator.cxx#L534).
Shall I leave a FIXME comment on that as a reminder? It could be fixed, when ARM64 is added to machineTypeOption officially.

gyp/pylib/gyp/MSVSNew.py
kaadam: Yes, that would be better, I will look at it.

gyp/pylib/gyp/generator/msvs.py
jkunkee: Great to hear. You're taking on cross compiling in Windows, and I really appreciate it.

gyp/pylib/gyp/msvs_emulation.py
jkunkee: I saw an email with a comment from you that I don't see here. `18` is a pretty good guess for now, and a simple TODO comment about ensuring it matches the machineTypeOption enum in the future would be great.

Also: I've been told the enum will be updated in VS2019, not VS2017.

gyp/pylib/gyp/msvs_emulation.py
```diff
@@ -237,6 +237,11 @@ def GetExtension(self):
   def GetVSMacroEnv(self, base_to_build=None, config=None):
     """Get a dict of variables mapping internal VS macro names to their gyp
     equivalents."""
+    target_arch = self.GetArch(config)
+    if target_arch == 'x86':
+      target_platform = 'Win32'
+    else:
+      target_platform = target_arch
     target_platform = 'Win32' if self.GetArch(config) == 'x86' else 'x64'
```
jkunkee: I believe this addition replaces the line following it, so line 245 here should be deleted.

jkunkee:
While you're working on this, @kaadam, a really thorough test is to check out https://github.com/jkunkee/node/tree/arm64-merged-for-build, apply your GYP changes, change to vcbuild.bat to invoke gyp properly and invoke msbuild twice, once with Win32 and once with ARM64, run `vcbuild.bat arm64`, and see what is still broken.

Of course, refack suggested on https://github.com/refack/GYP/issues/27 to try and get Ninja-based dual-toolchain cross-compilation working first before trying it with MSVC. That's another approach.

kaadam:
@jkunkee I updated the change.
Thanks for your comments, I have looked at building node itself targeting ARM64. I ran into that problem, what you mention https://github.com/refack/GYP/issues/27. As part of V8 build's task the bytecode_builtins_list_generator builds an ARM64 build target EXE and then run on host machine, of course it fails. The icupkg build has same issue. Somehow GYP ignores if there is a project which has to be build only on host machine, despite the GYP knows host machine, and target_as well.
I took a look at ninja, but it needs more work to support cross-compilation. As I played with it there was some problem parsing GYP and generating proper ninja files for multiple toolset. That's a good idea to make it work with ninja first, after that it could be used on cross-platform by default.

All in all this patch is intent to support native module cross-compilation, doesn't solve it general. My question is: could it be landed if it is just focus on native-module? A general approach for that could be a follow up change? What do you think?

@refack I will upload only the GYP part to your repository, and include unit tests as well.

jkunkee:
Here's a little of the process I went through, along with my recommendation here.

GYP itself did not require much work to get it to generate correct ARM64 build files for Node.js proper: https://github.com/refack/GYP/pull/23. In order to produce correct x64+ARM64 files for Node.js for x64-hosted ARM64 Node.js builds, GYP still needs the dual-toolchain feature (https://github.com/refack/GYP/issues/27).

When I tried to build the Node.js tests, I discovered that many of them include native modules and that node-gyp required some changes to get them to build and pass: https://github.com/nodejs/node-gyp/pull/1655

That is as far as I went.

From what I understand, @kaadam produced this patch as something to help make Node.js v12 embeddable in ARM64 Electron. That means that this change has been tested with far more node-gyp configurations than my change and, since this is intended to support native module and not Node.js compilation, it means that the dual-toolchain requirement is not present.

After reading through the new change, I think this makes sense to land as 'single-toolchain ARM64 build support'.

gyp/pylib/gyp/generator/msvs.py
kaadam: Fixed, based on your comments.

gyp/pylib/gyp/msvs_emulation.py
kaadam: Updated the value to 18. I also left TODO comments.

gyp/pylib/gyp/MSVSNew.py
kaadam: Looked at the variants and removed ARM64 from default. The platform variants is passed properly during configuration phase.

kaadam:
I've updated the title of the PR, and commit description as you suggested. Thanks for your help and the review. 
Yes, basically I'm focusing on Electron targeting Windows on ARM.
This change was manually tested on several native modules.

jkunkee:
@kaadam, instead of my repo please submit the GYP portion as a PR to @refack's GYP reboot: https://github.com/refack/GYP

(Note that, as of https://github.com/nodejs/node/pull/26620, that is the place GYP will come from for Node.js.)

@refack, are you prepping a PR to update `node-gyp` to GYP3 v6.0.0?

jkunkee:
@kaadam, thanks for lifting the GYP parts out to refack/GYP!

Would you mind taking the time to trim this PR down to just the node-gyp part? Maybe then we can get it merged.

refack:
> @refack, are you prepping a PR to update `node-gyp` to GYP3 v6.0.0?

Still cleaning up edge cases... Should be ready very soon.

kaadam:
@jkunkee Thanks for reminder, I've updated the PR.

jkunkee:
This is adding support for the experimental ARM64 Windows platform. It relies on changes to GYP as well that will come in another way; is there a way to merge this now?

joaocgreis:
The change here was included in https://github.com/nodejs/node-gyp/pull/1739, which just landed. I believe this can be closed now.

PR #1715, opened 2019-04-09
win: Support Visual Studio 2019

syohex:
##### Checklist

##### Description of change
- Support Visual Studio 2019.
Original code does not work with Visual Studio 2019 because path of VS2019's MSBuild.exe is different from VS2017's MSBuild.exe.
- And improve error check. If msbuild.exe is not found, then throws error.


lib/configure.js
```diff
@@ -142,16 +142,33 @@ function configure (gyp, argv, callback) {
...
       defaults['msbuild_toolset'] = 'v141'
       defaults['msvs_windows_target_platform_version'] = vsSetup.sdk
-      variables['msbuild_path'] = path.join(vsSetup.path, 'MSBuild', '15.0',
-                                            'Bin', 'MSBuild.exe')
+
+      var msbuildPath;
+      try {
+        // Visual Studio 2017
+        msbuildPath = path.join(vsSetup.path, 'MSBuild', '15.0', 'Bin', 'MSBuild.exe')
+        if (!fs.existsSync(msbuildPath)) {
```
Mr1008: How about inversing the logic here and prioritize rather VS2019 (Current) directory than 15.0. IMHO current approach would prevent users from compiling vs2019 projects when they have installed both 2017 and 2019 version of VS because msbuild 15.0 would be discovered in the first place. If it would be inverted, MSBuild 2019 would correctly detect older build tools as well. What do you think guys?

refack:
Problem is MSVS2019 comes with 14.2 toolset, so in that case it needs to set:
```
defaults['msbuild_toolset'] = 'v142'
```

But anyway, as soon as [GYP3](https://github.com/refack/GYP) passes all tests on 2017 & 2019, I'll PR to replace the old GYP, so we won't need these tricks (GYP3 has full support for MSVS 2017 & 2019)

lib/configure.js
syohex: I agree and fix by https://github.com/nodejs/node-gyp/pull/1715/commits/29b0c2e381f824fa4e641d8659274f5616703d87 . Thanks for good suggestion.

cleever:
Any estimate of when it will be available?

KENNYSOFT:
Please accept this ASAP...

fuyutsuki:
I want to use this solution ASAP…

joaocgreis:
@syohex support ended up landing from another PR, I believe we can close this one now. Thanks for working on this and opening this PR.

node-gyp v5.0.0 is out with Visual Studio 2019 support. It's not yet integrated into npm so it has to be installed manually. This should do it:

```
npm install --global node-gyp@latest
for /f "delims=" %P in ('npm prefix -g') do npm config set node_gyp "%P\node_modules\node-gyp\bin\node-gyp.js"

```

That is, install node-gyp globally and make the npm config variable `node_gyp` point to `node-gyp.js`.

<details>
<summary>How to undo this after node-gyp is integrated into npm</summary>

```console
npm uninstall --global node-gyp
npm config delete node_gyp

```

</details><br>

PR #1705, opened 2019-04-01
test: fix addon test for Node.js 12 and V8 7.4

richardlau:
##### Checklist

- [x] `npm install && npm test` passes
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

V8 7.4 removes some API functions. Replace those with their NAN
counterparts.

Example current failure: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1781/nodes=debian9-64/testReport/junit/(root)/citgm/node_gyp_v3_8_0/

```console
      undefinedadded 121 packages from 83 contributors and audited 228 packages in 5.603s
 found 0 vulnerabilities
 > node-gyp@3.8.0 test /home/iojs/build/workspace/citgm-smoker/nodes/debian9-64/citgm_tmp/98118c6b-7f61-4357-abfd-24950d89737d/node-gyp
 > tape test/test-*
 TAP version 13
 # build simple addon
 not ok 1 should be equal
   ---
     operator: equal
     expected: |-
       null
     actual: |-
       { [Error: Command failed: /home/iojs/build/workspace/citgm-smoker/nodes/debian9-64/smoker/bin/node /home/iojs/build/workspace/citgm-smoker/nodes/debian9-64/citgm_tmp/98118c6b-7f61-4357-abfd-24950d89737d/node-gyp/bin/node-gyp.js rebuild -C /home/iojs/build/workspace/citgm-smoker/nodes/debian9-64/citgm_tmp/98118c6b-7f61-4357-abfd-24950d89737d/node-gyp/test/node_modules/hello_world --loglevel=verbose
 gyp info it worked if it ends with ok
 gyp verb cli [ '/home/iojs/build/workspace/citgm-smoker/nodes/debian9-64/smoker/bin/node',
 gyp verb cli   '/home/iojs/build/workspace/citgm-smoker/nodes/debian9-64/citgm_tmp/98118c6b-7f61-4357-abfd-24950d89737d/node-gyp/bin/node-gyp.js',
 gyp verb cli   'rebuild',
 gyp verb cli   '-C',
 gyp verb cli   '/home/iojs/build/workspace/citgm-smoker/nodes/debian9-64/citgm_tmp/98118c6b-7f61-4357-abfd-24950d89737d/node-gyp/test/node_modules/hello_world',
 gyp verb cli   '--loglevel=verbose' ]
[...]

PR #1718, opened 2019-04-12
 v3.8.1 proposal

refack:
v3.8.1 2019-04-12
=================

### Due to security concerns this version drops support for Node.js versions < 4.0.0

 * [[`3578b2abf0`](https://github.com/nodejs/node-gyp/commit/3578b2abf0)] - **deps**: explicit limit on supported engines (Refael Ackermann)
* [[`ec8505e15f`](https://github.com/nodejs/node-gyp/commit/ec8505e15f)] - **deps**: updated tar package version to 4.4.8 (Pobegaylo Maksim) [#1713](https://github.com/nodejs/node-gyp/pull/1713)
* [[`6e1e425ffb`](https://github.com/nodejs/node-gyp/commit/6e1e425ffb)] - **(BREAKING for node < 4)** Upgrade to tar v3 (isaacs) [#1212](https://github.com/nodejs/node-gyp/pull/1212)
* [[`e6699d13cd`](https://github.com/nodejs/node-gyp/commit/e6699d13cd)] - **test**: fix addon test for Node.js 12 and V8 7.4 (Richard Lau) [#1705](https://github.com/nodejs/node-gyp/pull/1705)
* [[`0c6bf530a0`](https://github.com/nodejs/node-gyp/commit/0c6bf530a0)] - **lib**: use print() for python version detection (GreenAddress) [#1534](https://github.com/nodejs/node-gyp/pull/1534)

##### Checklist

- [X] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   "bin": "./bin/node-gyp.js",
   "main": "./lib/node-gyp.js",
   "dependencies": {
-    "fstream": "^1.0.0",
     "glob": "^7.0.3",
     "graceful-fs": "^4.1.2",
     "mkdirp": "^0.5.0",
@@ -32,11 +31,11 @@
     "request": "^2.87.0",
     "rimraf": "2",
     "semver": "~5.3.0",
-    "tar": "^2.0.0",
+    "tar": "^3.1.3",
     "which": "1"
   },
   "engines": {
-    "node": ">= 0.8.0"
+    "node": ">= 4.0.0"
   },
   "devDependencies": {
     "tape": "~4.2.0",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "request": "^2.87.0",
     "rimraf": "2",
     "semver": "~5.3.0",
-    "tar": "^4.4.8",
+    "tar": "4",
     "which": "1"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "request": "^2.87.0",
     "rimraf": "2",
     "semver": "~5.3.0",
-    "tar": "^3.1.3",
+    "tar": "^4.4.8",
     "which": "1"
   },
   "engines": {
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 gyp/test
 node_modules
 test/.node-gyp
+package-lock.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+v3.8.1 2019-04-12
+=================
+
+### Due to security concern this version drops support for Node.js versions < 4.0.0
+
+* [[`3578b2abf0`](https://github.com/nodejs/node-gyp/commit/3578b2abf0)] - **deps**: explicit limit on supported engines (Refael Ackermann)
+* [[`ec8505e15f`](https://github.com/nodejs/node-gyp/commit/ec8505e15f)] - **deps**: updated tar package version to 4.4.8 (Pobegaylo Maksim) [#1713](https://github.com/nodejs/node-gyp/pull/1713)
+* [[`6e1e425ffb`](https://github.com/nodejs/node-gyp/commit/6e1e425ffb)] - **(BREAKING for node < 4)** Upgrade to tar v3 (isaacs) [#1212](https://github.com/nodejs/node-gyp/pull/1212)
+* [[`e6699d13cd`](https://github.com/nodejs/node-gyp/commit/e6699d13cd)] - **test**: fix addon test for Node.js 12 and V8 7.4 (Richard Lau) [#1705](https://github.com/nodejs/node-gyp/pull/1705)
+* [[`0c6bf530a0`](https://github.com/nodejs/node-gyp/commit/0c6bf530a0)] - **lib**: use print() for python version detection (GreenAddress) [#1534](https://github.com/nodejs/node-gyp/pull/1534)
+
+
 v3.8.0 2018-08-09
 =================
 
--- a/lib/install.js
+++ b/lib/install.js
@@ -20,10 +20,8 @@ var fs = require('graceful-fs')
   , rm = require('rimraf')
   , path = require('path')
   , crypto = require('crypto')
-  , zlib = require('zlib')
   , log = require('npmlog')
   , semver = require('semver')
-  , fstream = require('fstream')
   , request = require('request')
   , mkdir = require('mkdirp')
   , processRelease = require('./process-release')
@@ -148,41 +146,33 @@ function install (fs, gyp, argv, callback) {
       var tarPath = gyp.opts.tarball
       var badDownload = false
         , extractCount = 0
-        , gunzip = zlib.createGunzip()
-        , extracter = tar.Extract({ path: devDir, strip: 1, filter: isValid })
 
       var contentShasums = {}
       var expectShasums = {}
 
       // checks if a file to be extracted from the tarball is valid.
       // only .h header files and the gyp files get extracted
-      function isValid () {
-        var name = this.path.substring(devDir.length + 1)
-        var isValid = valid(name)
-        if (name === '' && this.type === 'Directory') {
-          // the first directory entry is ok
-          return true
-        }
+      function isValid (path, entry) {
+        var isValid = valid(path)
         if (isValid) {
-          log.verbose('extracted file from tarball', name)
+          log.verbose('extracted file from tarball', path)
           extractCount++
         } else {
           // invalid
-          log.silly('ignoring from tarball', name)
+          log.silly('ignoring from tarball', path)
         }
         return isValid
       }
 
-      gunzip.on('error', cb)
-      extracter.on('error', cb)
-      extracter.on('end', afterTarball)
-
-      // download the tarball, gunzip and extract!
+      // download the tarball and extract!
 
       if (tarPath) {
-        var input = fs.createReadStream(tarPath)
-        input.pipe(gunzip).pipe(extracter)
-        return
+        return tar.extract({
+          file: tarPath,
+          strip: 1,
+          filter: isValid,
+          cwd: devDir
+        }).then(afterTarball, cb)
       }
 
       try {
@@ -222,7 +212,11 @@ function install (fs, gyp, argv, callback) {
         })
 
         // start unzipping and untaring
-        req.pipe(gunzip).pipe(extracter)
+        res.pipe(tar.extract({
+          strip: 1,
+          cwd: devDir,
+          filter: isValid
+        }).on('close', afterTarball).on('error', cb))
       })
 
       // invoked after the tarball has finished being extracted


refack:
CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/121/

richardlau:
https://github.com/nodejs/node-gyp/commit/fcb0117025 needs the rest of the changes to `lib/install.js` from https://github.com/nodejs/node-gyp/pull/1212/files#diff-f6618e1cc731d58106a806b7679a7616.

Could you also swap the ordering so that the update to tar 4.4.8 comes last? That way every commit would build in case we'd ever want to bisect.

refack:
Rebased over https://github.com/nodejs/node-gyp/pull/1718/commits/6e1e425ffbeca39381de37663431e05af7a00f56
https://ci.nodejs.org/job/nodegyp-test-pull-request/122/

stof:
Shouldn't CI jobs for 0.10 and 0.12 be removed from the matrix as support is dropped ?

refack:
> Shouldn't CI jobs for 0.10 and 0.12 be removed from the matrix as support is dropped ?

Will do, but only after this PR gets some LGTMs.

package.json
```diff
@@ -32,11 +31,11 @@
     "request": "^2.87.0",
     "rimraf": "2",
     "semver": "~5.3.0",
-    "tar": "^2.0.0",
+    "tar": "4",
```
emilis-tm: Should this be?:

```suggestion
    "tar": "^4.4.8",
```

Otherwise users without `npm audit` may still get the unsecure version.

pumano:
looks like version should be 4.0.0 according to semver due to breaking changes

jaredbeck:
> Shouldn't this be a semver-minor or semver-major though, since we're dropping support for Node.js 4?

When updating dependencies (in this case node) SemVer does not necessarily require a major release.

> What should I do if I update my own dependencies without changing the public API?
> That would be considered compatible since it does not affect the public API.
> https://semver.org/spec/v2.0.0.html#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

snappyfoo:
Can someone explain what the hold up is here please?

nvonpentz:
pls merge

BridgeAR:
Ping @refack 

If this includes a breaking change, it should also be a new major version. I believe it should be possible to release a 3.8.1 that just includes the non breaking commits and then to release v4.0.0 right after.

muxator:
Hi, would the major change be limited to requiring Node > 4?

If so, it would be acceptable to explain in the changelog (if it's visible enough) that projects that are already on a recent Node version can upgrade without problems. At least for Etherpad (https://github.com/ether/etherpad-lite/issues/3598) this would be ok.

A minor version that only solves the vulnerability would clearly even better (if technically possible).

dlipeles:
Hi, this issue is blocking CI for many projects, would it be possible to reach a resolution soon?

stof:
Well, it looks like dropping support for 0.10 and 0.12 is done to be able to upgrade tar. So dropping this change would also drop the vulnerability fix (making the 3.8.1 release useless). There is no patched version of tar supporting node 0.10 and 0.12.

stof:
@richardlau if you release the security fix only as 4.0, this would then require a coordinated effort in the community to upgrade all packages using node-gyp to version 4 ASAP though (as their existing semver constraint would not allow installing the patched version).

benwiley4000:
Is someone able to make an executive decision on this? I can understand the arguments both ways for minor or major, but it's also worth noting that there already is a coordinated effort by folks trying to figure out how to avoid using this vulnerable version of tar while node-gyp waits to release. A semver major release might be inconvenient, but lots of folks are waiting for that anyway, and `npm audit` will tell folks what they need to do at that point if they need to make a major version upgrade.

stoneark:
Could you guys prioritize it and make an immediate decision? You know a lot of projects have node-gyp as the dependency, and will absolutely get an error in CI/CD pipeline or cyber security check. I know the license is "provided as is", but I think it's better to make a new release as minor to fix it ASAP, just for the community. Thanks.

richardlau:
Since this appears stuck, I've tagged this on the agenda for this week's [TSC meeting](https://github.com/nodejs/TSC/issues/695#issuecomment-485412465).

ChALkeR:
/cc @nodejs/security-wg 

1. If updating node-tar to a non-vulnerable version breaks support for 0.10/0.12, and if this is the first release to break that support (I have not personally checked that), then this should be a semver-major and 4.0.0.

2. The tar vulnerability does not look dangerous here -- if one is in position to exploit the tar in node-gyp, they are most likely is position to just ship malicious code in package even without the tar vuln.
So this is not a security problem, strictly speaking.

3. Raising false positive `npm audit` warnings is not a good thing, because we don't want to diminish the signal from `npm audit` in situations when it's actually important.

4. Yes, releasing that as a semver major would mean that downstream would have to bump the version of this package to get the update. But there is nothing wrong with that. If someone could fix that, it's the [node-tar](https://github.com/npm/node-tar) package. I.e. backporting https://github.com/npm/node-tar/commit/b0c58433c22f5e7fe8b1c76373f27e3f81dcd4c8 to previous branches of node-tar would fix this potential vulnerability (and `npm audit` warning) for both node-gyp and all the other consumers of node-tar.

package.json
```diff
@@ -32,11 +31,11 @@
...
     "which": "1"
   },
   "engines": {
-    "node": ">= 0.8.0"
+    "node": ">= 4.0.0"
```
ChALkeR: This is a semver-major change.

YasharF:
If doing a semver-major bump of node-gyp due to the min engine requirement bump, it might be worth considering to pick a currently supported/security patched versions of NodeJS  https://github.com/nodejs/Release . With end-of-life of NodeJS 6.x next week, NodeJS 8.x might be the better choice:  "node": ">= 8.0.0"

rvagg:
OK, so I've done the last few releases of node-gyp iirc, and I've been getting emails about this one too. Here's the general problem that I bump up against and why I don't jump straight in to make it happen: releases of node-gyp are kind of critical, we can break very large chunks of the ecosystem if we don't get them right and I have a lot of personal experience with having it break. The test suite is almost non-existent and you just have to browse the issues here to see the diversity of environments that it's used in and the novel ways it can mess up. When I do a release I'm taking some responsibility for that. I'll also very likely end up being the person who has to go do emergency clean-up when the breakage happens. So releases are about far more than just typing `npm publish`.

It looks like @refack has done a good job of pulling in just the important pieces here so maybe this isn't such a big deal. I'll have a deeper look and maybe push out a 4.0.0 later today.

The divergence on `master` is getting large so that really needs to be dealt with too. I probably can't be relied on to solve that but I'm also not enthused by the idea of handing out publishing access liberally given what I've said about the state of this codebase and the impact of releases--they're really important to get right and it's really hard to know if they are right!

xzyfer:
To @rvagg's comment I can say that breaking 0.10 and 0.12 compatibility going to force some down streams to into major version bumps to quieten questionable relevant npm audit warning. Speaking for node-sass we still maintain BC for 0.10 and 0.12 so we'd need bump a major version to drop old Node support. This is something we have plans to do eventually but we're far from ready.

>If someone could fix that, it's the node-tar package. I.e. backporting npm/node-tar@b0c5843 to previous branches of node-tar would fix this potential vulnerability (and npm audit warning) for both node-gyp and all the other consumers of node-tar.

If someone could make this happen, it would be the least disruptive.

ChALkeR:
@xzyfer Perhaps an issue should be filed to https://github.com/npm/node-tar as the starting point, explaining that the usage of older branches is still high and that updating to tar@4.x comes with a major version bump because of dropping support Node.js < 4.0.

xzyfer:
@ChALkeR done in https://github.com/npm/node-tar/issues/212

ChALkeR:
@xzyfer Thanks! I left a comment there and also pinged npm about that.

guybedford:
If there are concerns with releasing this, there is nothing wrong with making it a major release.

package.json
rvagg: this is the only part left in 6bc9bf1 so I'm removing that commit entirely and it'll end up with `^4.4.8`

rvagg:
v4.0.0 is out, now we have to consider the `v3.x` branch closed, let's move on to `master` now which I'm rebasing and will put up a v5.0.0 proposal to see what that might look like

rvagg:
`master` and `v3.x` were a bit of a mess so I've done a rebase and force push to `master` so you may want to update your local copies. The old master is preserved at `pre-v4.0.0-master` and a v5.0.0 proposal is up at #1723 for discussion so we can clean up the backlog, it's getting a bit out of hand.

benwiley4000:
Thank you for releasing!!

xzyfer:
Is there any plan to cut a node-gyp release that doesn't break support for Node < 4 if a fix for node-tar@2 can be found?

rvagg:
@xzyfer I think it'd be reasonable to cut another v3.x if you can get an old branch of node-tar released that will work with it

BridgeAR:
What about forking tar, fixing the issue in v2 and to use the fork in node-gyp 3 (ideally we do not need a fork and just a PR and a new release for tar@2)? I know it's just a work around but that's at least something we could do for people who can't update yet.

xzyfer:
No need to fork. The node-tar folks have said they're open to accepting a PR.

rvagg:
we're having a hard enough time maintaining node-gyp, I don't think we really need another codebase to manage even if limited in scope

mikern12:
I mentioned this on this thread (https://github.com/nodejs/node-gyp/commit/1456ef27c72cc4e49f9b1d82cbb047cb673156c8) , but I am still having an issue when I update node-gyp.  I run npm install node-gyp and get the correct 4.0.0 package, but then I get an error that I something changed and I need to "Run npm rebuild node-sass to download the binding for your current environment." When I do this, it re-loads the 3.8.x version of node-gyp. Error below from the rebuild of node-sass:

gyp ERR! node -v v9.4.0
gyp ERR! node-gyp -v v3.8.0
gyp ERR! This is a bug in node-gyp.
gyp ERR! Try to update node-gyp and file an Issue if it does not help:
gyp ERR! https://github.com/nodejs/node-gyp/issues
Build failed with error code: 7
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! node-sass@4.11.0 postinstall: node scripts/build.js
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the node-sass@4.11.0 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

Fishrock123:
I have edited some info into the original post so as to make this less confusing.

xzyfer:
node-tar@2.2.2 has been released which patches the vulnerability. Would it be possible to cut release a node-gyp@2.x with this update so that upstreams aren't forced break old node BC?

xzyfer:
NVM looks like node-gyp depended on node-tar@^2.0.0 which should mean the issue is resolved.

erikeckhardt:
Github is still reporting the vulnerability as existing because NIST's vulnerability report didn't consider 2.2.2 of `tar` to be acceptable. So, I contacted NIST directly about 2.2.2 not having the vulnerability, and got the following response:

> After review of the CVEs, the information provided, and the configurations we have made the appropriate modifications. Please allow up to 24 hours for these changes to populate on the website and in the data feeds.

So stay tuned, because soon tar 2.2.2 will be recognized as an acceptable, non-vulnerable version, and appropriate versions of node-gyp using `tar@^2.0.0` will no longer be flagged as vulnerable.

Issue #1749, opened 2019-05-12
Support HTTPS_PROXY


50Wliu:
* **Node Version**: 10.15.3, 6.9.0
* **Platform**: Windows Enterprise 10.0.18362 N/D Compilación 18362
* **Compiler**: N/A
* **Module**: N/A

node-gyp supports basic proxy options: https://github.com/nodejs/node-gyp/blob/721eb691cf15556cc2700eda0558d5bad5f84232/lib/install.js#L432-L435
However, it doesn't look like it supports HTTPS_PROXY (and variants).
Would this be something that node-gyp is willing to support? I can create a PR if the answer is yes.

bzoz:
Sure, PRs are always welcomed!

rvagg:
closing in favour of PR @ #1751

PR #1751, opened 2019-05-14
Add support for HTTPS_PROXY

50Wliu:
##### Checklist

- [ ] tests are included 

##### Description of change

See #1749. If the requested URI is HTTPS, attempt to use an HTTPS proxy, if configured, and fallback to the HTTP proxy if one isn't found. This change should be backwards-compatible except in the case where someone is using a broken HTTPS proxy (in which case I would argue that that isn't node-gyp's problem anyway).

If desired, I can remove the `url` require and substring the uri to detect if it's HTTPS.

No tests were added because there aren't any existing proxy tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ Some additional resources for addons and writing `gyp` files:
 | `--devdir=$path`                  | SDK download directory (default is OS cache directory)
 | `--ensure`                        | Don't reinstall headers if already present
 | `--dist-url=$url`                 | Download header tarball from custom URL
-| `--proxy=$url`                    | Set HTTP proxy for downloading header tarball
+| `--proxy=$url`                    | Set HTTP(S) proxy for downloading header tarball
 | `--cafile=$cafile`                | Override default CA chain (to download tarball)
 | `--nodedir=$path`                 | Set the path to the node source code
 | `--python=$path`                  | Set path to the Python 2 binary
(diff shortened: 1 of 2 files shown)


rvagg:
I think I'm OK with this but I think it's going to have to be semver-major because it introduces potentially new behaviour in existing environments. Would you mind squashing the commits into one and changing the prefix to `lib:` please?

50Wliu:
Sounds good! I'll get to that in the next day or two.

50Wliu:
@rvagg squashed!

rvagg:
I'd like to get some tests in for this. Have a look in test/test-download.js, you'll see that you can interact directly with the `download` function and even pass it environment variables. So it should be straightfoward to come up with some tests that trigger each of these variables. The trick is going to be confirming that it's connecting via a proxy. There's a test/simple-proxy.js that's currently only used by docker.sh (I'm thinking that needs to be removed since it's not really used). That could be repurposed for this. There's also a self-signed certificate in test/fixtures/ that could be used for https proxy tests.
Let me know if you have trouble with this. I can help out as time allows if you're struggling.

50Wliu:
Popping in to say I haven't forgotten about this -- just a little short on time recently.

rvagg:
I believe we have this now via #1978, thanks for the code though

Issue #1774, opened 2019-06-13
SyntaxError: from __future__ imports must occur at the beginning of the file


paulmdavies:
The most recent release (released to npm today) doesn't work because of some python formatting issues:

`SyntaxError: from __future__ imports must occur at the beginning of the file`

This can be seen simply by checking out the repo and doing:

`python <repo location>/gyp/tools/pretty_vcproj.py`

or 

`python <repo location>/gyp/tools/pretty_sln.py`

It's a problem with both python 2 and python 3.

rvagg:
@cclauss, @refack, halp!

cclauss:
Why don't we see this in our Travis CI tests?  https://travis-ci.com/nodejs/node-gyp

PR #1775, opened 2019-06-13
Fix positions of 'from __future__' statements

paulmdavies:
See #1774


rvagg:
I think this needs to go into https://github.com/refack/GYP, right @refack?

rvagg:
whoops sorry @paulmdavies, I didn't piece it together and prompted this change that got merged already: https://github.com/nodejs/node-gyp/pull/1789
thanks for your submission anyway, my bad for not pulling it in

PR #1789, opened 2019-06-20
Move from __future__ import to the top of the file

### LINKED ISSUE #1774: SyntaxError: from __future__ imports must occur at the beginning of the file

cclauss:
Fixes #1774

##### Checklist

- [ ] tests are included 

### DIFF
--- a/gyp/tools/pretty_sln.py
+++ b/gyp/tools/pretty_sln.py
@@ -12,14 +12,14 @@
    Then it outputs a possible build order.
 """
 
-__author__ = 'nsylvain (Nicolas Sylvain)'
-
 from __future__ import print_function
 
 import os
 import re
 import sys
 import pretty_vcproj
+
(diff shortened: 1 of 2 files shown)


rvagg:
we're getting out of sync with refack/GYP, would you mind upstreaming to there too so when we do an upgrade we don't overwrite.

also tests to find this in the future would be nice, I guess we don't do any vcproj builds in our tests

rvagg:
2cf1f78

PR #1794, opened 2019-06-22
src: implement standard.js linting

rvagg:
IMO it's time for some proper linting, the code style is all over the place and the existing rules were super minimal. Current style is pretty close to standard.js, and it's a fairly universally accepted and understood style.

I've done a couple of additional things on top:

* moved `module.exports` to the bottom
* no single-line `if` statements
* no `if` statements without a `{`
* `const` for requires

This is going to make it harder to manage older pull requests of course, so this comes with a cost.

Thoughts?

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,6 +1,0 @@
----
-  env:
-    node: true
-    es6: true
-  rules:
-    no-unused-vars: error
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,7 +1,0 @@
-{
-  "asi": true,
-  "laxcomma": true,
-  "esversion": 6,
-  "node": true,
-  "strict": false
-}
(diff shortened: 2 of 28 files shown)


richardlau:
I've no real preferences re. `standard` vs other lint rules as long as the linting is part of the tests (which it is).

Since we have Travis enabled now and the linting is run as part of the `npm test` script we'll get visual feedback in newer and rebased PR's if the linting fails. I'm kind of assuming older PR's will need to be rebased anyway.

rvagg:
implemented 'use strict' and a couple of other minor tweaks including making consistent spacing with array declarations `[ foo, bar ]`, not `[foo, bar]`, we have both.

joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/135/

joaocgreis:
@rvagg it would be good to get this one landed to avoid it becoming outdated. AFAICT this is ready to land and I can do it tomorrow, please let me know if I shouldn't.

PR #1796, opened 2019-06-23
test: configure proper devDir for invoking configure()

rvagg:
test/test-configure-python.js downloads a fresh set of headers to the package directory each time. By setting to the default global cache dir we get to re-use cached headers and skip the download step. Speeds up tests and prevents the creation of a new directory with headers in it in the package root each test run.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -1,0 +1,3 @@
+const envPaths = require('env-paths')
+
+module.exports.devDir = () => envPaths('node-gyp', { suffix: '' }).cache
--- a/test/test-configure-python.js
+++ b/test/test-configure-python.js
@@ -2,6 +2,7 @@
 
 const test = require('tap').test
 const path = require('path')
+const devDir = require('./common').devDir()
 const gyp = require('../lib/node-gyp')
 const requireInject = require('require-inject')
 const configure = requireInject('../lib/configure', {
--- a/test/test-configure-python.js
+++ b/test/test-configure-python.js
@@ -30,6 +30,7 @@ test('configure PYTHONPATH with no existing env', function (t) {
     t.equal(process.env.PYTHONPATH, EXPECTED_PYPATH)
     return SPAWN_RESULT
   }
+  prog.devDir = devDir
   configure(prog, [], t.fail)
 })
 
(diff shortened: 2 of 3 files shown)


richardlau:
> 
> 
> test/test-configure-python.js downloads a fresh set of headers to the package directory each time. By setting to the default global cache dir we get to re-use cached headers and skip the download step. Speeds up tests and prevents the creation of a new directory with headers in it in the package root each test run.

Just to double check -- we still download the headers at least once elsewhere in the tests?

rvagg:
@richardlau maybe not, maybe I'll add one to test-download.js which mocks it all out I think.

richardlau:
> 
> 
> @richardlau maybe not, maybe I'll add one to test-download.js which mocks it all out I think.

Another way to check would be to see if coverage drops after this PR (now we're using `tap`).

rvagg:
I've rebased this and added a whole new test that downloads headers and verifies contents. @nodejs/node-gyp ptal.

test/common.js
```diff
@@ -0,0 +1,3 @@
+const envPaths = require('env-paths')
+
+module.exports.devDir = () => envPaths('node-gyp', { suffix: '' }).cache
```
richardlau: Based on https://github.com/nodejs/build/issues/1908#issuecomment-544298898 this will make tests write into the user's home directory and not clean up afterwards.

rvagg: Yes, it does, the alternative is having a separate one that needs cleaning up afterward.

So test/test-configure-python.js doesn't need to test downloading, but downloading gets in the way, which is why we end up with directories in cwd when running the tests, so it needs to point to somewhere. If we point it to the user's node-gyp cache then any user likely running these tests is also fairly likely to already have the current version of Node's headers downloaded in that directory so it gets to skip the download part for the purpose of test/test-configure-python.js. But then since we're no longer including the download part in the tests (if the user has the headers), I introduced the new test/test-download.js which uses the same cache path but cleans it before running it then puts a downloaded version there and leaves it.

Messing with the user's home directory isn't ideal, but I also don't want to have multiple _actual_ downloads in a single test run, making the tests longer and also leaving more of a mess to clean up. At least this way the only mess we're making is either going to be a duplicate of what the user already has or will be useful for them when they next need the headers for the version they're running.

Taking too many liberties outside of cwd?

richardlau: > 
> Taking too many liberties outside of cwd?

Let's live with it for now.

rvagg:
landed

PR #1802, opened 2019-06-28
Update README.md

lagorsse:
missing argument configure

##### Checklist

- [ ] tests are included 

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ If you have multiple Python versions installed, you can identify which Python
 version `node-gyp` uses by setting the `--python` variable:
 
 ``` bash
-$ node-gyp --python /path/to/python2.7
+$ node-gyp configure --python /path/to/python2.7
 ```
 
 If `node-gyp` is called by way of `npm`, *and* you have multiple versions of


README.md
```diff
@@ -63,7 +63,7 @@ If you have multiple Python versions installed, you can identify which Python
 version `node-gyp` uses by setting the `--python` variable:
 
 ``` bash
-$ node-gyp --python /path/to/python2.7
+$ node-gyp configure --python /path/to/python2.7
```
rvagg: The option applies to more than just `configure`, but what you could do here is this:

---

<pre>
```
$ node-gyp &lt;command&gt; --python /path/to/python2.7
```

Where `&lt;command&gt;` is one of the [commands](#commands) listed below.
</pre>

---

 (I'm using pre tags to format this, don't use the raw markdown, use the rendered markdown, also the 'bash' type should be dropped here, it gets in the way of proper formatting)

cclauss: Also, this line uses __/path/to/python2.7__ but...
the example just a few lines below uses __/path/to/executable/python2.7__

This would be more readable it we used consistent paths in these two examples.

rvagg:
modified it myself and landed in f82a007

PR #1808, opened 2019-07-05
test: downgrade to tap@^12 for continued Node 6 support

rvagg:
Fixes https://github.com/nodejs/node-gyp/pull/1795#issuecomment-508426193

PR #1809, opened 2019-07-05
test: cover supported node versions with travis

rvagg:
I think this might be the best way to get full coverage in travis. @cclauss @joaocgreis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ matrix:
   include:
     - name: "Python 2.7 on Linux"
       python: 2.7
-    - name: "Python 3.7 on Linux"
-      python: 3.7
     - name: "Python 2.7 on macOS"
       os: osx
       osx_image: xcode10.2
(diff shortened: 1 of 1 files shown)


cclauss:
Sure.  We can work this.  A lot of this was done on https://github.com/refack/GYP

.travis.yml
```diff
@@ -18,6 +16,18 @@ matrix:
...
+      python: 3.7
+      before_install: nvm install 8
+    - name: "Node.js 10 & Python 3.7 on Linux"
+      python: 3.7
+      before_install: nvm install 10
+    - name: "Node.js 12 & Python 3.7 on Linux"
+      python: 3.7
+      before_install: nvm install 12
   allow_failures:
     - os: windows
```
cclauss: Perhaps add __python: 3.7__ to __allow_failures__ ;-)

cclauss:
While you are in here, can you please change __F72__ on line 39 to __F7__ so that we add the [__flake8 F70x tests](http://flake8.pycqa.org/en/latest/user/error-codes.html) to our scans?  I know that we do not have any __F70x__ issues today but putting this in place would catch any new occurrences.

rvagg:
> please change F72 on line 39 to F7
gibberish! but done

cclauss:
Do you understand why Node 6 fails?  Is it worth fixing?

rvagg:
yes, it will turn magically green when #1808 is merged

Issue #1337, opened 2017-11-14
Support for Python 3


cclauss:
See ~#1150~, ~#1335~, ~#1336~, and ~https://travis-ci.com/nodejs/node/builds/79706150~ (~nodejs/node#21942~)

* **Node Version**: 
* **Platform**: 
* **Compiler**: 
* **Module**: 

<details><summary>Verbose output (from npm or node-gyp):</summary>

```

```

</details>

richardlau:
Historically the blocker to this has been the gyp dependency and normally changes should be upstreamed there unless we're planning on forking gyp in which case we would want our own gyp project so that we can share the changes with the main Node.js project.

cc @nodejs/gyp

mthacker801:
I cannot successfully install Angular on my mac.  I keep getting gyp error and when I look at gyp-mac-tool it says:
gyp verb command configure []
gyp verb check python checking for Python executable "python2" in the PATH
gyp verb `which` failed Error: not found: python2
gyp verb `which` failed     at getNotFoundError (/Users/thackerfam/AngularProj/angular-cli/node_modules/which/which.js:13:12)

I have had python 3 installed forever and do not want python 2 as it can cause havoc to have both installed as they are not compatible with one another.  If this is the reason for my inability to use Angular, what should I do?

cclauss:
What happens when you do:
* __python2 --version__
* __/usr/bin/python --version__
* __/usr/bin/python2 --version__
* __/usr/bin/python2.7 --version__

I would highly recommend [Homebrew](https://brew.sh) as essential if you are doing real development on a Mac.  You can easily install [both Python 2 and Python3](https://docs.brew.sh/Homebrew-and-Python.html) with: __brew install python2 python3__

See:

cclauss:
A few Python __2__ issues on [__gyp__](https://chromium.googlesource.com/external/gyp):

$ python2 -m flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
```
./gyptest.py:61:47: F821 undefined name 'opts'
    extra_path = [os.path.abspath(p) for p in opts.path]  # <-- 'opts' should be changed to 'args'
                                              ^
./pylib/gyp/generator/make.py:1637:50: F821 undefined name 'target'
      print "WARNING: no output for", self.type, target  # <-- 'target' should be changes to 'self.target'
                                                 ^
```

SatishMHiremath:
[...]

PR #1811, opened 2019-07-07
doc: Declare that node-gyp is Python 3 compatible

cclauss:
Edit: Blocked because Travis CI tests on Python 3 are failing...

NOTE: node-gyp is compatible with both Python 2.7 and 3.7 but Node.js itself is not yet compatible with Python 3.  See https://github.com/nodejs/node-gyp/issues/1337#issuecomment-508988254

Careful review please because I am not a Windows user.

##### Checklist

- [ ] tests are included 

##### Description of change

Declare that node-gyp is Python 3 compatible as discussed at https://github.com/nodejs/node-gyp/issues/1337#issuecomment-508988254

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,19 +25,23 @@ $ npm install -g node-gyp
 
 You will also need to install:
 
+NOTE: node-gyp is compatible with Python 2.7 and 3.7 but node itself is not yet compatible with Python 3.
+
 ### On Unix
 
-   * `python` (`v2.7` recommended, `v3.x.x` is __*not*__ supported)
+   * `Python v2.7 or v3.7` (v3.5 and 3.6 may work but are not currently tested)
    * `make`
    * A proper C/C++ compiler toolchain, like [GCC](https://gcc.gnu.org)
 
 ### On macOS
 
-   * `python` (`v2.7` recommended, `v3.x.x` is __*not*__ supported) (already installed on macOS)
+   * `Python v2.7 or v3.7` (v3.5 and 3.6 may work but are not currently tested)
    * [Xcode](https://developer.apple.com/xcode/download/)
      * You also need to install the `XCode Command Line Tools` by running `xcode-select --install`. Alternatively, if you already have the full Xcode installed, you can find them under the menu `Xcode -> Open Developer Tool -> More Developer Tools...`. This step will install `clang`, `clang++`, and `make`.
 
 ### On Windows
+
+Install the current version of Python from the [Microsoft Store package](https://docs.python.org/3.7/using/windows.html#the-microsoft-store-package).
 
 #### Option 1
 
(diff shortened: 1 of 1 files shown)


cclauss:
2.7 should be the lower limit.

joaocgreis:
> 2.7 should be the lower limit.

@cclauss is there any node-gyp reason for that? Last time I checked (some time ago) it worked. I understand Python 2.6 is not recommendable at all, but is there anything on the node-gyp side forcing us to break users who can't upgrade?

cclauss:
After [5 years with no security patches](https://devguide.python.org/devcycle/#end-of-life-branches), Python 2.6 must be a rather sizable attack vector.  Also the Python 3 compatibility changes that we have made have not been tested on Python 2.6 so we should not encourage our users to run untested configurations in production.   [Stop using Python 2.6](https://snarky.ca/stop-using-python-2-6), written by Python core team member, goes into more detail.

joaocgreis:
I don't think accepting it is the same as encouraging users to use it. Perhaps we should be clear in the docs that we only test on the latest version, others may or may not work. I still think that we don't have a reason to break 2.6, but I don't really have a strong opinion so I'll defer to other @nodejs/node-gyp members.

cclauss:
I have a strong opinion that we should not be encouraging our users to expose themselves to security risks.  Also, we are tied to really out-of-date technologies and this long overdue backward compatibility promise is stunting our modernization.

richardlau:
> 
> 
> I don't think accepting it is the same as encouraging users to use it. Perhaps we should be clear in the docs that we only test on the latest version, others may or may not work. I still think that we don't have a reason to break 2.6, but I don't really have a strong opinion so I'll defer to other @nodejs/node-gyp members.

I can get behind dropping support for Python 2.6 at the same time as dropping support for Node.js 6 in a a `node-gyp` semver major. Declaring Python 3 support (assuming it does actually work) would be a positive change for the semver major.

rvagg:
@cclauss @joaocgreis can you update us on the status of this? can we merge it?

cclauss:
I think we are ready for this PR to land.  Of the [PRs with a Python label](https://github.com/nodejs/node-gyp/pulls?q=is%3Aopen+is%3Apr+label%3APython) the only one that worries me is the macOS bit of #1844 but that is __semver-major__.

joaocgreis:
Can we "declare that node-gyp is Python 3 compatible" if it's broken on macOS?

I can't work on macOS support myself at the moment, so I won't block this if you think it makes sense. Node-gyp would be broken by default on macOS.

rvagg:
is this semver-major? would it be appropriate to put on 5.x?

rvagg:
landed in c763ca1 but if this shouldn't be in 5.x, please label it semver-major so it only goes into 6.

joaocgreis:
This should NOT go into 5.

This depends on https://github.com/nodejs/node-gyp/pull/1844, so this should only go into 6.

rvagg:
Maybe we should clarify in the readme that v5 is different to >v5 in this respect? A lot of people are going to be landing at this README regardless of their version.

Issue #1801, opened 2019-06-26
Travis CI tests are failing on Windows


cclauss:
Travis CI currently runs tests on Windows but those tests are currently run in __allow_failures__ mode.  https://github.com/nodejs/node-gyp/blob/master/.travis.yml#L15-L20

It would be really cool to get those tests passing.

richardlau:
> 
> 
> Travis CI currently runs tests on Windows but those tests are currently run in **allow_failures** mode. https://travis-ci.com/MobSF/Mobile-Security-Framework-MobSF/builds/116968308#L734
> 
> It would be really cool to get those tests passing.

Did you post the wrong link?

cclauss:
Yes.  Fixed inline above.

cclauss:
@krisavi @joaocgreis If you have any cycles/ideas on how to get Windows tests working on Travis CI...

PR #1812, opened 2019-07-07
Fix running Windows on Travis CI

joaocgreis:
##### Checklist

##### Description of change

`test-addon.js` includes compiling code, making the default 30 second timeout not suitable. This increases the timeout for all platforms, which is a potential problem everywhere, fixing the timeout that happens on Windows. Currently, the test takes 53 seconds on Windows and 17 on OSX.

This also adds Node 6 on Windows to the matrix. There is a lot of platform-specific code, so running the oldest supported version seems reasonable.

Fixes: https://github.com/nodejs/node-gyp/issues/1801

cc @cclauss

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,6 @@ matrix:
     - name: "Node.js 12 & Python 3.7 on Linux"
       python: 3.7
       before_install: nvm install 12
-  allow_failures:
-    - os: windows
 install:
   #- pip install -r requirements.txt
   - pip install flake8  # pytest  # add another testing frameworks later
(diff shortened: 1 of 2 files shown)


joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/140/

rvagg:
I was playing with this in the test itself https://github.com/nodejs/node-gyp/compare/rvagg/fix-windows-unicode-test?expand=1 but couldn't get it to work with the time i had available, my timeout wasn't being applied

anyway, this is reasonable I think, +1

cclauss:
[__travis_wait__](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) might be worth considering.

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/compare/656117c...bb92c76, thanks.

@richardlau the first two tests seem to use about half the time each, so it would be too tight. I've seen a run of 34 seconds just for the first test.

PR #1813, opened 2019-07-07
build: set Python semverRange: '>=2.7.0 <3.8.0'

cclauss:
##### Checklist

- [ ] tests are included 

##### Description of change

Based on the discussion #1811 let's try really running the tests on Python 3.  This currently breaks that macOS builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ matrix:
       python: 3.7
       before_install: nvm install 12
   allow_failures:
+    - os: osx
     - os: windows
 install:
   #- pip install -r requirements.txt
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -362,7 +362,7 @@ PythonFinder.prototype = {
   log: logWithPrefix(log, 'find Python'),
   argsExecutable: ['-c', 'import sys; print(sys.executable);'],
   argsVersion: ['-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);'],
-  semverRange: '>=2.6.0 <3.0.0',
+  semverRange: '>=2.7.0 <3.8.0',
 
   // These can be overridden for testing:
   execFile: cp.execFile,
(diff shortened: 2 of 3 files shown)


lib/configure.js
```diff
@@ -362,7 +362,7 @@ PythonFinder.prototype = {
   log: logWithPrefix(log, 'find Python'),
   argsExecutable: ['-c', 'import sys; print(sys.executable);'],
   argsVersion: ['-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);'],
-  semverRange: '>=2.6.0 <3.0.0',
+  semverRange: '>=2.7.0 <3.8.0',
```
richardlau: Do we need an upper bound on the semver check anymore? It was there before to exclude Python 3.

cclauss:
These tests would be more helpful if they printed __argsVersion__ on failure but I am not sure how to add that functionality.

cclauss:
Given the "purely stylistic" changes in #1794, should I also whack our whole codebase thru [python/black](https://github.com/python/black)?

lib/configure.js
cclauss: Now it is __semverRange: '>=2.7.0'__

.travis.yml
```diff
@@ -29,6 +29,7 @@ matrix:
       python: 3.7
       before_install: nvm install 12
   allow_failures:
+    - os: osx
```
sam-github: why?

lib/configure.js
```diff
@@ -383,9 +383,9 @@ function PythonFinder (configPython, callback) {
 
 PythonFinder.prototype = {
   log: logWithPrefix(log, 'find Python'),
-  argsExecutable: [ '-c', 'import sys; print(sys.executable);' ],
-  argsVersion: [ '-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);' ],
-  semverRange: '>=2.6.0 <3.0.0',
+  argsExecutable: [ '-c', 'import sys; print(sys.executable)' ],
```
sam-github: is dropping the `;` cosmetic, or necessary? If cosmetic, I'd prefer to leave as is to not mix cosmetic and functional changes.

sam-github:
other than small comments, LGTM

cclauss:
Closing in favor of #1815

PR #1819, opened 2019-07-12
Updated win_delay_load_hook.cc to work with /clr

ipetrovic11:
Added support for native managed modules with delay load

To support native modules that use managed code(/clr) with delay load 
we need to compile win_delay_load_hook.cc as unmanaged  by setting
```#pragma unmanaged```

### DIFF
--- a/src/win_delay_load_hook.cc
+++ b/src/win_delay_load_hook.cc
@@ -6,7 +6,7 @@
  *
  * This allows compiled addons to work when the host executable is renamed.
  */
-
+#pragma unmanaged
 #ifdef _MSC_VER
 
 #ifndef WIN32_LEAN_AND_MEAN

### CODE REVIEW

**Comment by joaocgreis** [comment]
@ipetrovic11 if I read https://docs.microsoft.com/en-us/cpp/preprocessor/managed-unmanaged?view=vs-2019 correctly, shouldn't the pragma directive be placed right before the function definition?

**Comment by ipetrovic11** [comment]
@joaocgreis it doesn't have to be just for one function.
```
#pragma unmanaged
// Unmanaged code here, could be multiple functions
#pragma managed
```

It should be related to one source file, so if you state at the top that `#pragma unmanaged`, the whole file will be considered as unmanaged. I believe that win_delay_load_hook.cc should be unmanaged.

joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/149/

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/commit/573607981e70dc5a28635676acc83afabe43317d. Thanks!

Issue #1834, opened 2019-07-19
npm install qruri get the error


kolvin2016:
[root@localhost client]# npm install qruri

> canvas@1.6.13 install /root/esplora/client/node_modules/canvas
> node-gyp rebuild

gyp ERR! UNCAUGHT EXCEPTION 
gyp ERR! stack SystemError [ERR_SYSTEM_ERROR]: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
gyp ERR! stack     at Object.userInfo (os.js:249:11)
gyp ERR! stack     at eaccesFallback (/usr/local/node/lib/node_modules/npm/node_modules/node-gyp/lib/install.js:374:92)
gyp ERR! stack     at /usr/local/node/lib/node_modules/npm/node_modules/node-gyp/lib/install.js:83:11
gyp ERR! stack     at callback (/usr/local/node/lib/node_modules/npm/node_modules/graceful-fs/polyfills.js:289:20)
gyp ERR! stack     at FSReqWrap.oncomplete (fs.js:154:21)
gyp ERR! System Linux 3.10.0-862.el7.x86_64
gyp ERR! command "/usr/local/node/bin/node" "/usr/local/node/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /root/esplora/client/node_modules/canvas
gyp ERR! node -v v10.15.0
gyp ERR! node-gyp -v v5.0.2
gyp ERR! This is a bug in `node-gyp`.
gyp ERR! Try to update node-gyp and file an Issue if it does not help:
gyp ERR!     <https://github.com/nodejs/node-gyp/issues>
npm WARN babelify@8.0.0 requires a peer of babel-core@6 || 7 || ^7.0.0-alpha || ^7.0.0-beta || ^7.0.0-rc but none is installed. You must install peer dependencies yourself.
npm WARN bootstrap@4.1.3 requires a peer of popper.js@^1.14.3 but none is installed. You must install peer dependencies yourself.
npm WARN esplora-client@0.1.0 No description
npm WARN esplora-client@0.1.0 No repository field.

npm ERR! code ELIFECYCLE
npm ERR! errno 7
npm ERR! canvas@1.6.13 install: `node-gyp rebuild`
npm ERR! Exit status 7
npm ERR! 
npm ERR! Failed at the canvas@1.6.13 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

[...]

PR #1835, opened 2019-07-20
lib: ignore non-critical os.userInfo() failures

rvagg:
Fixes: https://github.com/nodejs/node-gyp/issues/1834

There are certain environments in which `os.userInfo()` fails, I don't know what they are but I've seen this reported in various places. Since we don't _need_ the info we can safely ignore a failure.


rvagg:
rebased and landed in fa5b5aba97

PR #1843, opened 2019-07-23
Fix Travis runs with Python 3

joaocgreis:
##### Checklist

##### Description of change

This fixes the remaining issues with `test-find-python.js` and running on Windows when using Python 3. This makes node-gyp compatible with Python 3 for everything that we run in CI (which is not that much).

@cclauss the fixes here seemed logical to me, but I don't have much experience with Python, so a careful review would be very welcome.

This PR is semver-patch, using Python 3 still depends on `EXPERIMENTAL_NODE_GYP_PYTHON3`. I am preparing a different PR to address that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
       before_install: nvm install 12
     - name: "Python 3.7 on macOS"
       os: osx
-      osx_image: xcode11
+      #osx_image: xcode11
       language: shell  # 'language: python' is not yet supported on macOS
       env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
       before_install: HOMEBREW_NO_AUTO_UPDATE=1 brew install npm
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       python: 2.7
     - name: "Python 2.7 on macOS"
       os: osx
-      osx_image: xcode10.2
+      osx_image: xcode11
       language: shell  # 'language: python' is not yet supported on macOS
       env: NODE_GYP_FORCE_PYTHON=python2
       before_install: HOMEBREW_NO_AUTO_UPDATE=1 brew install npm
@@ -28,6 +28,7 @@ matrix:
         PATH=/c/Python27:/c/Python27/Scripts:$PATH
         NODE_GYP_FORCE_PYTHON=/c/Python27/python.exe
       before_install: choco install python2
+
     - name: "Node.js 6 & Python 3.7 on Linux"
       python: 3.7
       env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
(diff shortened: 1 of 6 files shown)


joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/151/

cclauss:
It would be of interest to know if the Travis tests would pass on Python 3 on macOS.  Please consider adding the changes from #1846 into this PR.

test/test-find-python.js
```diff
@@ -17,8 +16,8 @@ test('find python', function (t) {
     t.strictEqual(err, null)
     var proc = execFile(found, ['-V'], function (err, stdout, stderr) {
       t.strictEqual(err, null)
-      t.strictEqual(stdout, '')
-      t.ok(/Python 2/.test(stderr))
+      t.ok(stdout === '' || stderr === '')
+      t.ok(/Python 2/.test(stderr) || /Python 3/.test(stdout))
```
richardlau: Non-blocking but I think these two tests are going to be hard to debug if they fail (in the previous code if `stdout` wasn't equal to `''` the contents would be included in the test failure).

MattIPv4: Would it be possible to refactor this so that if we're on Python 3, we run checks for that, which are strict and so output contents if they fail, otherwise we run stricts for Python 2?

joaocgreis: Updated, PTAL.

joaocgreis:
Updated to include the macOS Travis run as discussed in https://github.com/nodejs/node-gyp/pull/1846.

@rvagg @MattIPv4 @richardlau let me know if you have any objection.

.travis.yml
```diff
@@ -44,6 +45,12 @@ matrix:
       python: 3.7
       env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
       before_install: nvm install 12
+    - name: "Python 3.7 on macOS"
+      os: osx
+      osx_image: xcode11
```
cclauss: I know that I am trying your patience but... can you please do me a favor comment out line 50.  The error messages (and investigations elsewhere) suggest to me that using the current xcode11 might be our problem.

https://docs.travis-ci.com/user/reference/osx/#xcode-version
```
+No receipt for 'com.apple.pkg.DeveloperToolsCLILeo' found at '/'.
    +No receipt for 'com.apple.pkg.DeveloperToolsCLI' found at '/'.
    +gyp: No Xcode or CLT version detected!
```

joaocgreis: No problem. Pushed a commit just to get a Travis run, I will discard it if xcode11 is not the problem.

The Python 2 version also uses xcode11, so I don't think this will fix it. If it does, we have to understand why it's not a problem with Python 2.

cclauss:
It moved us from 6 failing to 3 failing.  Thanks much!

joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/153/ ✔️ 

Landed in https://github.com/nodejs/node-gyp/compare/5459eca...2592036.

Thanks for the reviews!

@cclauss left the xcode line commented. Notice the number of passing tests is the same, so probably when a test fails we can't really consider the remaining tests in that file. `xcode11` works with Python 2, so it may make sense to duplicate the Python 3 entry, with and without xcode.

For reference:
- Run with `xcode11`: https://travis-ci.com/nodejs/node-gyp/jobs/219308516
- Run without xcode: https://travis-ci.com/nodejs/node-gyp/jobs/219321367

cclauss:
Prefect.  Thanks for this work.  Legacy Python continues to pass no matter what.

* On Python 3 on macOS xcode [9.4.1](https://docs.travis-ci.com/user/reference/osx/#xcode-version): [3 failing](https://travis-ci.com/nodejs/node-gyp/jobs/219321367#L3513)
* On Python 3 on macOS xcode 11: [6 failing]{https://travis-ci.com/nodejs/node-gyp/jobs/219308516#L3331}

PR #1844, opened 2019-07-23
Accept Python 3 by default

joaocgreis:
##### Checklist

##### Description of change

~This depends on https://github.com/nodejs/node-gyp/pull/1843, I will rebase when that PR lands. Only https://github.com/nodejs/node-gyp/commit/a9719470c1d1bf4ab0bd29ce135c2b8027c87a7a needs to be reviewed here.~

This makes node-gyp accept Python 3 and look for it before looking for Python 2, making Python 3 the default.

@nodejs/node-gyp please not that we cover little more than a basic hello world in our tests, there are probably parts of Gyp that are not covered and are not compatible with Python 3. We will have to be extra careful shipping the node-gyp version that includes this.

### DIFF
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -18,8 +18,7 @@ PythonFinder.prototype = {
   log: logWithPrefix(log, 'find Python'),
   argsExecutable: [ '-c', 'import sys; print(sys.executable);' ],
   argsVersion: [ '-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);' ],
-  semverRange: process.env.EXPERIMENTAL_NODE_GYP_PYTHON3 ? '2.7.x || >=3.5.0'
-    : '>=2.7.0 <3.0.0',
+  semverRange: '2.7.x || >=3.5.0',
 
   // These can be overridden for testing:
   execFile: cp.execFile,
@@ -92,6 +91,11 @@ PythonFinder.prototype = {
           before: () => { this.addLog('checking if "python" can be used') },
           check: this.checkCommand,
           arg: 'python'
+        },
+        {
+          before: () => { this.addLog('checking if "python3" can be used') },
+          check: this.checkCommand,
+          arg: 'python3'
         },
         {
           before: () => { this.addLog('checking if "python2" can be used') },
@@ -286,7 +290,7 @@ PythonFinder.prototype = {
     //                                                           X
     const info = [
       '**********************************************************',
-      'You need to install the latest version of Python 2.7.',
+      'You need to install the latest version of Python.',
       'Node-gyp should be able to find and use Python. If not,',
       'you can try one of the following options:',
       `- Use the switch --python="${pathExample}"`,
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,25 +31,25 @@ matrix:
 
     - name: "Node.js 6 & Python 3.7 on Linux"
       python: 3.7
-      env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
+      env: NODE_GYP_FORCE_PYTHON=python3
       before_install: nvm install 6
     - name: "Node.js 8 & Python 3.7 on Linux"
       python: 3.7
-      env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
+      env: NODE_GYP_FORCE_PYTHON=python3
       before_install: nvm install 8
     - name: "Node.js 10 & Python 3.7 on Linux"
       python: 3.7
-      env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
+      env: NODE_GYP_FORCE_PYTHON=python3
       before_install: nvm install 10
     - name: "Node.js 12 & Python 3.7 on Linux"
       python: 3.7
-      env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
+      env: NODE_GYP_FORCE_PYTHON=python3
       before_install: nvm install 12
     - name: "Python 3.7 on macOS"
       os: osx
       #osx_image: xcode11
       language: shell  # 'language: python' is not yet supported on macOS
-      env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
+      env: NODE_GYP_FORCE_PYTHON=python3
       before_install: HOMEBREW_NO_AUTO_UPDATE=1 brew install npm
     - name: "Node.js 12 & Python 3.7 on Windows"
       os: windows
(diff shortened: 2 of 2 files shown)


joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/152/

lib/find-python.js
```diff
@@ -18,8 +18,7 @@ PythonFinder.prototype = {
   log: logWithPrefix(log, 'find Python'),
   argsExecutable: [ '-c', 'import sys; print(sys.executable);' ],
   argsVersion: [ '-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);' ],
-  semverRange: process.env.EXPERIMENTAL_NODE_GYP_PYTHON3 ? '2.7.x || >=3.5.0'
-    : '>=2.7.0 <3.0.0',
+  semverRange: '2.7.x || >=3.5.0',
```
richardlau: Do we need to enforce an upper bound?

cclauss: Not yet.

joaocgreis: I don't think we should. If a new version of Python breaks something here, we should fix it. If we can't fix it for some reason, then would be a good time to exclude that version.

joaocgreis:
@cclauss I think https://github.com/nodejs/node-gyp/pull/1846 is the proper place to do it, this PR might take a while to land. I'll rebase on top of that PR and adjust accordingly when it lands!

joaocgreis:
This is blocked by https://github.com/nodejs/node-gyp/pull/1846, we can't land this with OSX failing.

lib/find-python.js
richardlau: Ah sorry. I misread this line as 3.5.0 being an upper bound but it's the lower bound for Python 3.

rvagg:
seems good to me but I'm confused on whether the macos problems block this and #1843 or if they can land independently of that?

cclauss:
We can land this with the macOS tests still running in __allow_failures__ mode.  Our README clearly states that Python 3 is not supported which aligns with Travis CI's intent for [__allow_failures__](https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail)
> [allow_failures] lets you add in experimental and preparatory builds to test against versions or configurations that you are not ready to officially support.

joaocgreis:
https://github.com/nodejs/node-gyp/pull/1843 has landed. In there, node-gyp only uses Python 3 if `EXPERIMENTAL_NODE_GYP_PYTHON3` is defined, so it should not break anything we currently have.

This PR removes that flag, making node-gyp accept Python 3. So, if a release is shipped with this PR, node-gyp will start using Python 3 in the users computers, and any issue we have will manifest there.

Currently, we know Python 3 is failing on osx, so shipping a release with this would break node-gyp on every osx machine that has Python 3 installed.

We know that our tests pass with Python 3 on Windows and Linux, but our tests don't have a good coverage of gyp. It's likely that there are modules out there that use parts of gyp that we don't test here, and those would be broken as well. GYP3 could help a lot with this.

So, we can't land this PR until all failures that we know of are fixed.

sam-github:
https://github.com/nodejs/node/issues/29246 bumped into this.

I'm not sure what the plan is. If I'm missing the conversation elsewhere, pls point me there! I don't want to fragment an existing conversation.

node's tests can't pass on a py3-only machine until npm has a copy of node-gyp that supports py3.

Does npm even get node-gyp from this repo? :-( Sorry, you can see how late to this "party" I'm arriving.

bnoordhuis:
> Does npm even get node-gyp from this repo?

Yes. :-)

joaocgreis:
@sam-github my comment above is still a valid summary of the current situation. Going forward, I believe the options on the table are either to use GYP3 (https://github.com/nodejs/node-gyp/pull/1845) or fix the osx issues (@cclauss might be working on that).

This PR is ready to be rebased and land landed when we believe Python 3 support to be ready. However, this PR does not change any Python code, it only removes the block to run on Python 3.

joaocgreis:
Rebased to make Travis run again, but it's still red. Current status: https://github.com/nodejs/node-gyp/pull/1844#issuecomment-515492930 and https://github.com/nodejs/node-gyp/pull/1844#issuecomment-530117175

cclauss:
The [macOS error](https://travis-ci.com/nodejs/node-gyp/jobs/239028189#L2513-L2522) looks related to nodejs/node-gyp#1854 so I will attack it.

cclauss:
Please rebase again to ensure macOS now passes.

cclauss:
Once this PR lands, does __NODE_GYP_FORCE_PYTHON=python3__ still have to be set to use Py3?
* On a machine that has both Python 2 and Python 3 installed?
* On a machine that only has Python 3 installed?

joaocgreis:
> Once this PR lands, does __NODE_GYP_FORCE_PYTHON=python3__ still _have_ to be set to use Py3?

Doesn't "have" to, but I'd rather keep it in most (if not all) of the entries in Travis, since not having it was the reason we didn't know what Python was actually being run before. This is a good way to avoid surprises.

> On a machine that has both Python 2 and Python 3 installed?
> On a machine that only has Python 3 installed?

Without `NODE_GYP_FORCE_PYTHON` the code looks for `python`, then `python3`, then `python2`. Whatever exists first is what's used.

joaocgreis:
Hit an issue because of https://github.com/isaacs/minizlib/issues/9 - Travis kept 1.3.1 in cache. Should be fixed now, forced update in one run and Travis updated the cache.

CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/155/ ✔️ 

Landed in https://github.com/nodejs/node-gyp/commit/3d1c60ab8183956cec40cf41ebc3b96a41bdcdcf

sam-github:
Does this change which version of python is chosen? I can't tell from the diff.

I.e
- with `python2`, `python3`, and `python` (2.7) in `/usr/bin`, which will get chosen?
- with `python2`, `python3`, and `python` (3.6) in `/usr/bin`, which will get chosen?

I don't have any concerns either way, but I think we should call out in the docs/changelog if the order of choice changed, right now it just says Python 3 is accepted, not under which conditions is preferred, if ever.

cclauss:
> Without NODE_GYP_FORCE_PYTHON the code looks for python, then python3, then python2.
> Whatever exists first is what's used.

I take that to mean __python__ will be used.

sam-github:
> This makes node-gyp accept Python 3 and look for it before looking for Python 2, making Python 3 the default.

@joaocgreis Can I edit this description? I think that isn't quite accurate, it should be

> This makes node-gyp accept Python 3 as compatible without needing EXPERIMENTAL_NODE_GYP_PYTHON3 in the environment. It also looks for `python`, `python3`, then `python2` (in that order). Note that `python` is usually Python 2.

Personally, I suggest that this PR can and should land on 5.x, with the minor tweek of moving `python3` to *after* `python2`, so that users will Python 2 if they have it.

cf https://github.com/nodejs/node-gyp/pull/1892#issuecomment-538009731 and previous discussion

PR #1845, opened 2019-07-24
 gyp: vendor in GYP3 6.0.3

refack:
https://github.com/refack/GYP/commit/8806c633810730d0a78138a550f0ce3e5047abfb

##### Checklist

- [X] `npm install && npm test` passes
- [X] tests are included 
- [X] documentation is changed or added
- [X] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

PR #1846, opened 2019-07-24
build: add test run Node.js 12 & Python 3.7 on macOS

cclauss:
##### Checklist

- [ ] tests are included 

##### Description of change

Travis CI: Add test run Node.js 12 & Python 3.7 on macOS

Edit: This test is currently run in __allow_failures__ mode because Linux and macOS share the same __env:__ values.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       python: 2.7
     - name: "Python 2.7 on macOS"
       os: osx
-      osx_image: xcode10.2
+      osx_image: xcode11
       language: shell  # 'language: python' is not yet supported on macOS
       env: NODE_GYP_FORCE_PYTHON=python2
       before_install: HOMEBREW_NO_AUTO_UPDATE=1 brew install npm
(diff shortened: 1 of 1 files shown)


joaocgreis:
@cclauss good call, there are indeed failures there. If you plan to fix these, you can add the fixes to this PR. If not, this new run should be added to allowed failures.

Just in case it's useful, to quickly see the error locally I use
```
EXPERIMENTAL_NODE_GYP_PYTHON3=1 NODE_GYP_FORCE_PYTHON=python3 node bin/node-gyp.js rebuild -C test/node_modules/hello_world/ --loglevel=verbose
```

cclauss:
The Python 3 on macOS job is already run in allow_failures mode because Linux and macOS share the same __env:__ values.  This new test in this PR is really only interesting in combination with #1843, #1844, or #1845.

joaocgreis:
The failures in `test-find-python` are fixed in https://github.com/nodejs/node-gyp/pull/1843, I think they're safe to ignore here. There are other failures there that show real issues.

@cclauss if you'd prefer, I can include this in https://github.com/nodejs/node-gyp/pull/1843 leaving osx python 3 in allowed failures. Would that be better?

cclauss:
Yes.  Please include in #1843 and I am OK withleaving osx python 3 in allow_failures mode.

cclauss:
Closing in favor of #1843 which now includes these changes.

PR #1852, opened 2019-08-08
Improved "unmanaged" handling for delay loading

peter-sabath:
##### Checklist

##### Description of change
Used scoped disabling of `managed` code handling to ensure no other files get affected.

### DIFF
--- a/src/win_delay_load_hook.cc
+++ b/src/win_delay_load_hook.cc
@@ -7,9 +7,9 @@
  * This allows compiled addons to work when the host executable is renamed.
  */
 
-#pragma unmanaged
-
 #ifdef _MSC_VER
+
+#pragma managed(push, off)
 
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


joaocgreis:
The change looks reasonable to me, from what I understand of the docs it should have the same effect as what is there now.

@peter-sabath I expected the unmanaged pragma to be effective only within this compilation unit. How does this affect other files? Is it a stateful thing in the compiler or does it actually become a runtime setting?

@ipetrovic11 please let me know if this causes any issue to you.

peter-sabath:
@joaocgreis You most likely are right, that only the current compilation unit is affected, but as good practice I use those `#pragma`s in a scoped way similar to `#if` `#endif` markers. This would even work fine if MS decides to compine compilation units like they already do for C# for faster compile times.

joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/154/

ipetrovic11:
@joaocgreis just tested, works fine for me with this improvement as well.
I am not proficient enough to comment if this is better practice or not.

@peter-sabath if you believe they will do something like that, I am pretty sure they will be handling solutions like first one, since they would break a lot of projects :)

peter-sabath:
@ipetrovic11 This was just my way of solving it. We ran into the `abort` problem from the loader. So we should not argue about good style / bad style, but be happy that this nasty problem is solved. So keep it or leave it, your solution also does what is required.

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/commit/af876e10f01ea8e3fdfeee20dbee3f7138ccffd5.

Thanks!

PR #1854, opened 2019-08-13
allow double-digit versions in XcodeVersion

### LINKED ISSUE #1849: Could not get CLTVersion
if CLT version is over 10.*, it could not get the version.
`version = re.match(r'(\d+\.\d+\.?\d*)', version).groups()`

and now latest CLT version is 10.14.

https://developer.apple.com/download/more/?=command%20line%20tools

https://github.com/nodejs/node-gyp/blob/872ee89a8541264ee2d96e162d410f478bba7815/gyp/pylib/gyp/xcode_emulation.py

roryqueue:
Fixes #1849

##### Checklist

- [ ] tests are included 

### DIFF
--- a/gyp/pylib/gyp/xcode_emulation.py
+++ b/gyp/pylib/gyp/xcode_emulation.py
@@ -1277,7 +1277,7 @@ def XcodeVersion():
   except:
     version = CLTVersion()
     if version:
-      version = re.match(r'(\d+\.\d+\.?\d*)', version).groups()[0]
+      version = re.match(r'(\d?\d\.\d?\d\.?\d?\d*)', version).groups()[0]
     else:
       raise GypError("No Xcode or CLT version detected!")
     # The CLT has no build information, so we return an empty string.


cclauss:
Related to #1849 ?

roryqueue:
yes, relates-- sorry at work and not able to dig further atm, but that change seems to fix

roryqueue:
(or rather, obviously doesn't work given it's failing, but the regex change matches both 9.X and 10.X, where as the other one fails on 10.X, so could be a hint to a real fix)

rvagg:
Sorry, I'm not a python person but this looks invalid to me, `\d+` will match double digits just as well as `\d?\d` in any normal regex engine. Can I close this or is someone willing to dispute that Python does something funky with regexes that makes this necessary?

cclauss:
Can we just replace this line with __".".join(version.split(".")[:3]__ and move on?  Not a fan of regex for trivial use cases.

gyp/pylib/gyp/xcode_emulation.py
```diff
@@ -1277,7 +1277,7 @@ def XcodeVersion():
   except:
     version = CLTVersion()
     if version:
-      version = re.match(r'(\d+\.\d+\.?\d*)', version).groups()[0]
+      version = re.match(r'(\d?\d\.\d?\d\.?\d?\d*)', version).groups()[0]
     else:
       raise GypError("No Xcode or CLT version detected!")
```
cclauss: On exception, let's add the version that was found...
` raise GypError("No Xcode or CLT version detected!  {}".format(version)`

cclauss:
I believe that #1890 fixed this. Please rebase and reverify.

cclauss:
Conflicts to resolve.

rvagg:
I think this is dealt with already and the solution posed here isn't correct as far as I can tell

PR #1860, opened 2019-08-19
Use "is" when comparing to None

MrMebelMan:
##### Checklist

- [ ] tests are included 

##### Description of change

Do not use the comparison operator `==` when checking if an object is `None`.
https://www.python.org/dev/peps/pep-0008/#programming-recommendations


rvagg:
landed in a301abc

Issue #1084, opened 2017-01-04
"node-gyp configure" fails with ECONNRESET error.


YaminiManthena1:
npm install -g node-gyp was successful but node-gyp configure throws the error below. Trying to upgrade node version in our application.

Mine is a Windows 7 64 bit machine. 

D:\App>node-gyp configure
gyp info it worked if it ends with ok
gyp info using node-gyp@3.4.0
gyp info using node@6.9.2 | win32 | x64
gyp http GET https://nodejs.org/download/release/v6.9.2/node-v6.9.2-headers.tar.gz
gyp WARN install got an error, rolling back install
gyp ERR! configure error
gyp ERR! stack Error: read ECONNRESET
gyp ERR! stack     at exports._errnoException (util.js:1022:11)
gyp ERR! stack     at TLSWrap.onread (net.js:569:26)
gyp ERR! System Windows_NT 6.1.7601
gyp ERR! command "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\XXXX\\AppData\\Roaming\\npm\\node_modules\\node-gyp\\bin\\node-gyp.j
s" "configure"
gyp ERR! cwd D:\IMVNode new version\IMV
gyp ERR! node -v v6.9.2
gyp ERR! node-gyp -v v3.4.0
gyp ERR! not ok

spectravp:
+1 Same issue here on Ubuntu 14.04

gyp info it worked if it ends with ok
gyp verb cli [ '/usr/bin/nodejs', '/usr/bin/node-gyp', 'rebuild', '--verbose' ]
gyp info using node-gyp@3.5.0
gyp info using node@6.9.4 | linux | x64
gyp verb command rebuild []
gyp verb command clean []
gyp verb clean removing "build" directory
gyp verb command configure []
gyp verb check python checking for Python executable "python2" in the PATH
gyp verb `which` succeeded python2 /usr/bin/python2
gyp verb check python version `/usr/bin/python2 -c "import platform; print(platform.python_version());"` returned: "2.7.6\n"
gyp verb get node dir no --target version specified, falling back to host node version: 6.9.4
gyp verb command install [ '6.9.4' ]
gyp verb install input version string "6.9.4"
gyp verb install installing version: 6.9.4
gyp verb install --ensure was passed, so won't reinstall if already installed
gyp verb install version not already installed, continuing with install 6.9.4
gyp verb ensuring nodedir is created /home/vagrant/.node-gyp/6.9.4
[...]

PR #1863, opened 2019-08-23
lib: adding keep-alive header to download requests

miladfarca:
##### Checklist

##### Description of change

We have been seeing random failures on AIX in the CI when building `node-addon-api`:
https://ci.nodejs.org/job/node-test-node-addon-api-new/nodes=aix61-ppc64/481/console

Taking a closer look shows the build fails when gyp is trying to download and extract `node source tarball`. This is done using a stream, data gets decompressed as its being downloaded:
```
...
gyp verb extracted file from tarball node-v13.0.0-nightly20190822775048d54c/include/node/openssl/opensslconf_no-asm.h
gyp verb extracted file from tarball node-v13.0.0-nightly20190822775048d54c/include/node/zconf.h
Error: read ECONNRESET
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:183:27) {
  errno: -73,
  code: 'ECONNRESET',
  syscall: 'read'
}
gyp WARN install got an error, rolling back install
...
```

I came up with a simpler test case to investigate the cause:
```js
const https = require('https');
var tar = require('tar');

const request = https.get("https://nodejs.org/download/nightly/v13.0.0-nightly20190822775048d54c/node-v13.0.0-nightly20190822775048d54c-headers.tar.gz", function(response) {
    response.pipe(tar.extract({
        strip: 1
    }));
});
```
Running the above on AIX using `node test.js` causes the same error to appear. It seemed like
the connection was being interrupted while stream was being decompressed.

I noticed requests are being handled by `cloudflare`. I tried moving the `headers.tar.gz` file to my personal server with https enabled and then ran the script. It worked with no errors.

I figured `cloudflare` is most likely dropping the TLS connection once it doesn't hear a tick which happens on AIX as the live tar extraction is interfering with it and is taking longer than other platforms.

I looked at the response headers from cloudflare and the default `connection` header is set  to `close`. Adding a `keep-alive` header in my request managed to fix the problem and force a live connection until the stream is done. I also tested this with `node-addon-api` by modifying `gyp` and the headers were downloaded successfully on AIX.

I have found a similar ticket from 2017 whit the same issue which might have been caused by cloudflare dropping connections:
https://github.com/nodejs/node-gyp/issues/1084

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -386,7 +386,8 @@ function download (gyp, env, url) {
   var requestOpts = {
     uri: url,
     headers: {
-      'User-Agent': 'node-gyp v' + gyp.version + ' (node ' + process.version + ')'
+      'User-Agent': 'node-gyp v' + gyp.version + ' (node ' + process.version + ')',
+      'Connection': 'keep-alive'
     }
   }
 


rvagg:
I don't really get this, it's just performing one request so keep-alive isn't technically necessary, do you think that turning it on is changing the cloudflare timeouts?

miladfarca:
I'm suspecting there is a timeout which this flag is disabling and server continues with sending the stream.

rvagg:
Well, it's odd, but I guess it can't _hurt_, so maybe we should give it a go. Any objections @nodejs/node-gyp?

rvagg:
landed in b2835405, let's cross fingers and see how this goes

PR #1875, opened 2019-09-13
src,win: add support for fetching arm64 node.lib

richard-townsend-arm:
##### Checklist

- [x] tests are included 

##### Description of change

Windows on Arm support is available in some versions of Node.js v12 and Electron v6. This update allows node-gyp to fetch the appropriate node.lib to build native modules. If an arm64 binary is not available for the target node version, it's logged but ignored.

arm64 is not expected to work in very old node.lib distribution formats, the test URLs in these cases are added to be consistent with x64.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -298,7 +298,7 @@ function install (fs, gyp, argv, callback) {
 
       function downloadNodeLib (done) {
         log.verbose('on Windows; need to download `' + release.name + '.lib`...')
-        var archs = [ 'ia32', 'x64' ]
+        var archs = [ 'ia32', 'x64', 'arm64' ]
         var async = archs.length
         archs.forEach(function (arch) {
           var dir = path.resolve(devDir, arch)
@@ -324,6 +324,11 @@ function install (fs, gyp, argv, callback) {
             req.on('error', done)
             req.on('response', function (res) {
               if (res.statusCode !== 200) {
+                if (res.statusCode === 404 && arch === 'arm64') {
+                  // Arm64 is a newer platform on Windows and not all node distributions provide it.
+                  log.verbose(res.statusCode + ' status code downloading ' + name)
+                  return
+                }
                 done(new Error(res.statusCode + ' status code downloading ' + name))
                 return
               }
--- a/lib/install.js
+++ b/lib/install.js
@@ -323,12 +323,15 @@ function install (fs, gyp, argv, callback) {
 
             req.on('error', done)
             req.on('response', function (res) {
-              if (res.statusCode !== 200) {
-                if (res.statusCode === 404 && arch === 'arm64') {
+              if (res.statusCode === 404) {
+                if (arch === 'arm64') {
                   // Arm64 is a newer platform on Windows and not all node distributions provide it.
-                  log.verbose(res.statusCode + ' status code downloading ' + name)
-                  return
+                  log.verbose(`${name} was not found in ${libUrl}`)
+                } else {
+                  log.warn(`${name} was not found in ${libUrl}`)
                 }
+                return
+              } else if (res.statusCode !== 200) {
                 done(new Error(res.statusCode + ' status code downloading ' + name))
                 return
               }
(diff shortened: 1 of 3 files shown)


joaocgreis:
@richard-townsend-arm thanks for opening this PR! This has been on my backlog for a while, because some more changes are needed. Now it's almost done:
- The Unofficial Builds server needs to have `node.lib` for ARM64. I've made the changes and I'm using https://unofficial-builds.nodejs.org/download/release/v12.8.0/ to test.
- I'd rather avoid having x86 and x64 binaries in that server to avoid confusion, so node-gyp needs to ignore failures downloading those as well. I've pushed a commit here to address this, but I can take it back if you prefer that I don't change this PR. A better solution would be to download only the file needed for the target architecture, but that's a much less simple change. I believe this is good for now.
- The Unofficial Builds server needs to have a copy of the headers and a `RELEASE_URLBASE` pointing to it: https://github.com/nodejs/unofficial-builds/pull/9. When that lands, I'll test this and make a release of the latest Node version. It'll be a while before a node-gyp with this change will be included in a Node release, but we can always update manually.

richard-townsend-arm:
Cool, thanks for that @joaocgreis: I'm happy for your fixups to go into into this PR.

joaocgreis:
Tested with v12.8.0 from unofficial builds.

CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/156/

Landed in https://github.com/nodejs/node-gyp/commit/2441932dc490b2c0f1b7f962a52810d35a7e3d30

joaocgreis:
Published ARM64 release of v12.10.0: https://unofficial-builds.nodejs.org/download/release/v12.10.0/ (v12.11.0 has an issue with unofficial builds, fix in https://github.com/nodejs/unofficial-builds/pull/10).

To use node-gyp directly from the master branch, including this PR:
```
npm i -g https://github.com/nodejs/node-gyp
for /f "delims=" %P in ('npm prefix -g') do npm config set node_gyp "%P\node_modules\node-gyp\bin\node-gyp.js"
```

MehediH:
Hi, I am trying to build my Electron app for Windows 10 on ARM, but this issue is preventing me from doing so. 

Firstly, I don't have a `node-gyp` folder in my `%APPDATA/Local%` directory, and the `node.lib` file download link meant for ARM64 leads to a 404[ on this page.](https://www.electronjs.org/docs/tutorial/windows-arm#linking-against-the-correct-nodelib)

Could you please point me towards the right file for ARM64 and where it should be placed for the ARM64 compilation to work as expected?

tommyvct:
> Hi, I am trying to build my Electron app for Windows 10 on ARM, but this issue is preventing me from doing so.
> 
> Firstly, I don't have a `node-gyp` folder in my `%APPDATA/Local%` directory, and the `node.lib` file download link meant for ARM64 leads to a 404[ on this page.](https://www.electronjs.org/docs/tutorial/windows-arm#linking-against-the-correct-nodelib)
> 
> Could you please point me towards the right file for ARM64 and where it should be placed for the ARM64 compilation to work as expected?

They are way too lazy to document things down. 
here is the lib.
https://unofficial-builds.nodejs.org/download/release/v15.5.1/win-arm64/

EwoutH:
I keep getting an 404 using [node-gyp](https://github.com/nodejs/node-gyp):

`gyp http 404 https://nodejs.org/download/release/v12.15.0/win-arm64/node.lib`

How can I instruct gyp to use the unofficial-builds?
(`https://unofficial-builds.nodejs.org/download/release/v12.15.0/win-arm64/node.lib` in this case)

richardlau:
Try setting `--dist-url`?

richardlau:
(As an aside https://github.com/EwoutH/Signal-Desktop/runs/1749166864?check_suite_focus=true#step:9:44 shows that the x64 version of Node.js is being run which is why it's looking in `https://nodejs.org/download/release` as the default place to look is relative to `process.release.headersUrl`.)

wbsdty331:
NodeJS now officially support Windows ARM64 platform, can we reconsider this issue?
https://github.com/nodejs/build/issues/2540

PR #1888, opened 2019-09-24
fix: the usage instructions

halitogunc:
##### Checklist

- [ ] tests are included 

##### Description of change
The descriptions of build and list commands are undefined on the usage instruction. This is happening because an error was made while exporting `usage` variable. 

image


rvagg:
awesome, thanks for picking this up and thanks for the fix!
landed in 0f5e258

PR #1828, opened 2019-07-16
Revert "build,test: add duplicate symbol test"

bnoordhuis:
This reverts commit 2761afbf730f7001675bbdbfa4a9e4821243adcb.

Building with `-fvisibility=hidden` breaks some of Node's add-on tests
and therefore likely also affects third-party add-ons. This change was
landed in a patch release so I'm opting to revert it until the next
major release.

Refs: https://github.com/nodejs/node/pull/28647#issuecomment-511715968

cc @gabrielschulhof

CI: https://ci.nodejs.org/job/nodegyp-test-pull-request/144/

PR #1891, opened 2019-09-26
Re-introduce duplicate symbol check

rvagg:
Originally in #1689 and was reverted in #1828, we decided in https://github.com/nodejs/node/pull/28647#issuecomment-511715968 it was "breaking" as far as Node was concerned.

Given the dependency chain to get to Node I'm not sure what to do about this. Is it enough to put a semver-major label on it? It's still likely to bubble up to Node 12 (and maybe 10) through their regular npm updates though. Is there a rule in place that would prevent certain versions of npm getting into Node 12? And how do we couple that with our moves to get to Python 3 support and the semver-major nature of them? Do we then prevent Node 12 from getting Python 3 support if we block a semver-major release of node-gyp getting back in to older versions of Node?

/cc @gabrielschulhof @bnoordhuis @nodejs/npm

### DIFF
--- a/addon.gypi
+++ b/addon.gypi
@@ -90,10 +90,14 @@
 
     'conditions': [
       [ 'OS=="mac"', {
+        'cflags': [
+          '-fvisibility=hidden'
+        ],
         'defines': [
           '_DARWIN_USE_64_BIT_INODE=1'
         ],
         'xcode_settings': {
+          'GCC_SYMBOLS_PRIVATE_EXTERN': 'YES', # -fvisibility=hidden
           'DYLIB_INSTALL_NAME_BASE': '@rpath'
         },
       }],
--- a/test/node_modules/duplicate_symbols/binding.cc
+++ b/test/node_modules/duplicate_symbols/binding.cc
@@ -2,9 +2,9 @@
 #include "common.h"
 
 void Init(v8::Local<v8::Object> exports) {
-  exports->Set(Nan::New("pointerCheck").ToLocalChecked(),
-               Nan::New<v8::FunctionTemplate>(Something::PointerCheck)
-      ->GetFunction());
+  Nan::Set(exports, Nan::New("pointerCheck").ToLocalChecked(),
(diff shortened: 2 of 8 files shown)


rvagg:
Unless someone thinks this is urgent I think we should punt this till 7.0.0 so we can focus on python 3 compat in 6.x.

gabrielschulhof:
@rvagg I have no special reason to push for urgency. It would be good if it landed though, because, as it is, there might be all kinds of weird errors with equally weird solutions.

cclauss:
Reading the output was superimpossible because of all the backslashes but at its root the test failure is caused by #1904

rvagg:
green after rebase to master

vweevers:
@rvagg I'm wondering why node doesn't use the `RTLD_LOCAL` flag for `dlopen()` by default? In my testing (of `leveldown` on mac) that fixed an overlap of symbols, to the same effect as `-fvisibility=hidden`.

rvagg:
I don't know, historical I suppose, and changing these things is tricky since it has so much breaking potential. But maybe that could be done in a semver-major in Node. @bnoordhuis I'm sure you understand this way better than I, thoughts?

vweevers:
Some data: https://github.com/vweevers/about-native-modules

rvagg:
To get some discussion going, maybe try opening a PR against nodejs/node to actually make the change? You'll get a lot of folks coming out of the woodwork with opinions if you did that I reckon.

vweevers:
I've surveyed a bunch of native addons and only found real issues in `leveldown` and `rocksdb` so far, apart from unrelated issues with shared libraries. So I don't have a strong case for changing node behavior. I'll continue to survey addons, improving my understanding of the issue along the way.

rvagg:
sooooo I'm not seeing a path to reintroducing this without coordination between node-gyp, npm and node.js to make sure it lands semver-major in each of them. I just don't see that happening.

PR #1893, opened 2019-09-26
Release proposal: v5.0.4

rvagg:
This is the same as the #1892 but without the lone apparent semver-major: #1818. See thread in https://github.com/nodejs/node-gyp/issues/1791#issuecomment-535169433 for more discussion on this approach.

Will take advice on what else this needs, I don't have any strong opinions other than "let's get this stuff out".

-------------------------

* [[`f753c167c5`](https://github.com/nodejs/node-gyp/commit/f753c167c5)] - **gyp**: decode stdout on Python 3 (cclauss) [#1890](https://github.com/nodejs/node-gyp/pull/1890)
* [[`60a4083523`](https://github.com/nodejs/node-gyp/commit/60a4083523)] - **doc**: update xcode install instructions to match Node's BUILDING (Nhan Khong) [#1884](https://github.com/nodejs/node-gyp/pull/1884)
* [[`19dbc9ac32`](https://github.com/nodejs/node-gyp/commit/19dbc9ac32)] - **deps**: update tar to 4.4.12 (Matheus Marchini) [#1889](https://github.com/nodejs/node-gyp/pull/1889)
* [[`5f3ed92181`](https://github.com/nodejs/node-gyp/commit/5f3ed92181)] - **bin**: fix the usage instructions (Halit Ogunc) [#1888](https://github.com/nodejs/node-gyp/pull/1888)
* [[`aab118edf1`](https://github.com/nodejs/node-gyp/commit/aab118edf1)] - **lib**: adding keep-alive header to download requests (Milad Farazmand) [#1863](https://github.com/nodejs/node-gyp/pull/1863)
* [[`1186e89326`](https://github.com/nodejs/node-gyp/commit/1186e89326)] - **lib**: ignore non-critical os.userInfo() failures (Rod Vagg) [#1835](https://github.com/nodejs/node-gyp/pull/1835)
* [[`785e527c3d`](https://github.com/nodejs/node-gyp/commit/785e527c3d)] - **doc**: fix missing argument for setting python path (lagorsse) [#1802](https://github.com/nodejs/node-gyp/pull/1802)
* [[`a97615196c`](https://github.com/nodejs/node-gyp/commit/a97615196c)] - **gyp**: rm semicolons (Python != JavaScript) (MattIPv4) [#1858](https://github.com/nodejs/node-gyp/pull/1858)
[...]

PR #1890, opened 2019-09-25
Decode stdout on Python 3

cclauss:
##### Checklist

- [ ] `npm install && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [ ] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

Fixes #1849, #1854, #1861  https://github.com/nodejs/node-gyp/search?q=CLT&type=Issues

Python 2 does not make a distinction between bytes and str but Python 3 does.  When we make calls to __subprocess.Popen()__ on Python 3 we should __stdout.decode("utf8")__ before performing string manipulations on it.

This modification also allows __Accept Python 3 by default__ #1844 to pass all Travis CI tests.

PR #1894, opened 2019-09-26
more decode stdout on Python 3

cclauss:
##### Checklist

- [ ] tests are included 

##### Description of change

Python 2 does not make a distinction between bytes and str but Python 3 does. When we make calls to subprocess.Popen() on Python 3 we should stdout.decode("utf8") before performing string manipulations on it.

Like #1890 

Addresses the code review comments at https://github.com/nodejs/node-gyp/pull/1893#pullrequestreview-293495919

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
       before_install: nvm install 12
     - name: "Python 3.7 on macOS"
       os: osx
-      osx_image: xcode11
+      #osx_image: xcode11
       language: shell  # 'language: python' is not yet supported on macOS
       env: NODE_GYP_FORCE_PYTHON=python3 EXPERIMENTAL_NODE_GYP_PYTHON3=1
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -47,7 +47,7 @@ matrix:
       before_install: nvm install 12
     - name: "Python 3.7 on macOS"
       os: osx
-      #osx_image: xcode11
+      osx_image: xcode11
```
bnoordhuis: This should be a separate commit, I think?

sam-github: Did it perhaps come from an incomplete rebase onto or merge from master? It does seem unrelated to the stdout encoding.

cclauss: It was one of the items mentioned at https://github.com/nodejs/node-gyp/pull/1893#pullrequestreview-293495919

rvagg:
landed in 67dec149

PR #1899, opened 2019-09-30
src: update to standard@14

rvagg:
This mainly undoes all of my nice, consistent array spacing. They fell down on the (erroneous IMO) side of no spaces either side of square brackets, regardless of their use. Unnecessary quoting of array/object properties is also now checked.

### DIFF
--- a/lib/process-release.js
+++ b/lib/process-release.js
@@ -1,3 +1,5 @@
+/* eslint-disable node/no-deprecated-api */
+
 'use strict'
 
 const semver = require('semver')
--- a/bin/node-gyp.js
+++ b/bin/node-gyp.js
@@ -131,7 +131,7 @@ function errorMessage () {
 
 function issueMessage () {
   errorMessage()
-  log.error('', [ 'This is a bug in `node-gyp`.',
+  log.error('', ['This is a bug in `node-gyp`.',
     'Try to update node-gyp and file an Issue if it does not help:',
     '    <https://github.com/nodejs/node-gyp/issues>'
   ].join('\n'))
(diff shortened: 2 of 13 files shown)


lib/process-release.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable node/no-deprecated-api */
```
rvagg: `url.resolve()` and `url.parse()` are used in here. I'm not entirely comfortable switching over fully to `URL` yet, it took a little while to stabilise, so maybe not until we want to drop support for Node 8, or until Node starts printing deprecation warnings to stdout.

cclauss:
> undoes all of my nice, consistent array spacing

Are there formatter directives like [Python Black's __# fmt: off__ and __# fmt: on__](https://github.com/psf/black#the-black-code-style) so we can use sparingly to preserve your array artistry?

cclauss:
Conflicts to resolve.

rvagg:
resolved.

> Are there formatter directives like Python Black's # fmt: off and # fmt: on so we can use sparingly to preserve your array artistry?

No, intentionally so. The purpose of `standard` is to provide a uniform standard that everyone can get on board with. We make a pact with each other to let go of most of our syntax preferences and accept `standard` as close-enough to what we like and just roll with it, removing most of the syntax arguing that's taken up so much GitHub comment space over the years. IMO the `[]` decision was short-sighted but I'm already putting up with it in many other projects and just getting on with the important work.

cclauss:
Agreed.  Same philosophy behind the Python Software Foundation’s __Black__...  You can have any color you want as long as it is Black.  ;-)

rvagg:
landed in 3555ff4f

PR #1900, opened 2019-09-30
Blacken our code

cclauss:
##### Checklist

- [x] tests are included 

##### Description of change

Like #1899 but for Python code.  Uses the Python Software Foundation's __Black__  https://github.com/psf/black to format all Python code in the canonical way and adds a test to Travis CI to avoid any backsliding.


rvagg:
By doing this, we're basically making it impossible for changes from refack/GYP3 to be pulled over, so we'd need to be sure that we're committing to maintaining GYP here and having the canonical version. I'm fine with applying consistent style linting, it makes contributions much more straightforward, but this is a big step, so let's let this linger and see what others have to say.

cclauss:
If we blacken __both__ codebases, we will get really clean diffs.

rvagg:
> If we blacken both codebases, we will get really clean diffs.

but lose the ability to cherry-pick isolated changes

PR #1892, opened 2019-09-26
Release proposal: v6.0.0

rvagg:
~~Just a proposal at this stage, I expect there's more to go into this, but this is what it might look like:~~

Due out on Oct 4th, UTC.

* [[`dd0e97ef0b`](https://github.com/nodejs/node-gyp/commit/dd0e97ef0b)] - **(SEMVER-MAJOR)** **lib**: try to find `python` after `python3` (Sam Roberts) [#1907](https://github.com/nodejs/node-gyp/pull/1907)
* [[`f60ed47d14`](https://github.com/nodejs/node-gyp/commit/f60ed47d14)] - **travis**: add Python 3.5 and 3.6 tests on Linux (Christian Clauss) [#1903](https://github.com/nodejs/node-gyp/pull/1903)
* [[`c763ca1838`](https://github.com/nodejs/node-gyp/commit/c763ca1838)] - **(SEMVER-MAJOR)** **doc**: Declare that node-gyp is Python 3 compatible (cclauss) [#1811](https://github.com/nodejs/node-gyp/pull/1811)
* [[`3d1c60ab81`](https://github.com/nodejs/node-gyp/commit/3d1c60ab81)] - **(SEMVER-MAJOR)** **lib**: accept Python 3 by default (João Reis) [#1844](https://github.com/nodejs/node-gyp/pull/1844)
* [[`c6e3b65a23`](https://github.com/nodejs/node-gyp/commit/c6e3b65a23)] - **(SEMVER-MAJOR)** **lib**: raise the minimum Python version from 2.6 to 2.7 (cclauss) [#1818](https://github.com/nodejs/node-gyp/pull/1818)

~~_(EDIT: pulled in v5.0.4 as a foundation and removed the commits that are common on both)_~~
_(EDIT: pulled in the v5.0.5 proposal from #1905 as a foundation and removed the commits that are common on both, `branch-diff rvagg/v5.0.5-proposal rvagg/v6.0.0-proposal`)_

PR #1901, opened 2019-10-01
Undefined name: cflags --> ldflags

cclauss:
@rvagg This would be good to land before #1892

The current code would raise NameError at runtime.

##### Checklist

- [ ] tests are included 


rvagg:
landed in a84b885, label semver-major if it shouldn't be backported

cclauss:
This can be backported...  My suspicion is that this part of the code is never run because if it was, the NameError would be raised at runtime.

cclauss:
This landed but is still open?

PR #1903, opened 2019-10-02
Travis CI: Add Python 3.5 and 3.6 tests on Linux

cclauss:
##### Checklist

- [ ] tests are included 


sam-github:
https://github.com/nodejs/node-gyp#on-unix  should be updated

I think this should be done, older ubuntu's only have py 3.5, and the docs say it "may" work, might as well add some tests to ensure that it does work, and continues to work.

rvagg:
landed in f60ed47d

PR #1907, opened 2019-10-03
lib: try to find `python` after `python3`

sam-github:
Unadorned `python` can be either Python 2 or Python 3, but it is likely
to be Python 2 for quite a while.

To find Python3, it is recommended to use the explicit name `python3`.

See:
- https://www.python.org/dev/peps/pep-0394/#for-python-runtime-distributors
- https://github.com/nodejs/node-gyp/pull/1892#issuecomment-537637433

##### Checklist

- [ ] tests are included 

### DIFF
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -26,8 +26,8 @@ PythonFinder.prototype = {
   win: win,
   pyLauncher: 'py.exe',
   winDefaultLocations: [
-    path.join(process.env.SystemDrive || 'C:', 'Python27', 'python.exe'),
-    path.join(process.env.SystemDrive || 'C:', 'Python37', 'python.exe')
+    path.join(process.env.SystemDrive || 'C:', 'Python37', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Python27', 'python.exe')
   ],
 
   // Logs a message at verbose level, but also saves it to be displayed later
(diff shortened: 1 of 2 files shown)


sam-github:
I'm not sure what to do about https://github.com/nodejs/node-gyp/blob/f36bd228a40d1d8b6fedab8ed650a4fb196ba228/lib/find-python.js#L186-L210

I don't have any Windows systems available ATM, so if I duplicate/refactor the Py Launcher stuff to look for 3 first, then 2, I've no way to test, unless its tested in CI, but I'm thinking it isn't.

I left it alone for now, though I did switch the order of the Windows default locations: https://github.com/nodejs/node-gyp/blob/f36bd228a40d1d8b6fedab8ed650a4fb196ba228/lib/find-python.js#L29-L30

cclauss:
I agree with this order for __v6.0.0__:
1. `python3`
2. `python`
3. `python2`

However, the only acceptable Pythons are 2.7, 3.5, 3.6, and 3.7.  So if `python3` is Python 3.4 then we do not use it and we continue down the list.

sam-github:
I note with relief that there are some Windows tests in Travis.

joaocgreis:
@sam-github The change of order of default locations LGTM. You can simply move the PyLauncher to after the default location check, swap the blocks starting on lines 108 and 115. I plan to review the PyLauncher code for Python 3 sometime, and add registry detection like we have in core.

sam-github:
@joaocgreis reordered as you suggested, which required some of the test monkey-patching to be adjusted for the new order of `execFile()`s.

rvagg:
landed in dd0e97e, let's pull the trigger then eh?

PR #1908, opened 2019-10-03
doc: clarify Python configuration, etc

sam-github:
Clarify that:
- node-gyp is not used to build Node.js
- PYTHON can be used to add a Python search path
- NODE_GYP_FORCE_PYTHON can be used to override all Python search paths
- That a compatible Python is searched for

##### Checklist

- [ ] tests are included 

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,12 @@
 # `node-gyp` - Node.js native addon build tool
 
-`node-gyp` is a cross-platform command-line tool written in Node.js for compiling
-native addon modules for Node.js. It bundles the [gyp](https://gyp.gsrc.io)
-project used by the Chromium team and takes away the pain of dealing with the
-various differences in build platforms.
+`node-gyp` is a cross-platform command-line tool written in Node.js for
+compiling native addon modules for Node.js. It contains a fork of the
+[gyp](https://gyp.gsrc.io) project that was previously used by the Chromium
+team and takes away the pain of dealing with the various differences in build
+platforms.
+
+Note that `node-gyp` is _not_ used to build Node.js itself.
 
 Multiple target versions of Node.js are supported (i.e. `0.8`, ..., `4`, `5`, `6`,
 etc.), regardless of what version of Node.js is actually installed on your system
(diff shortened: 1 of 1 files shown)


sam-github:
Note that this is not semver-major, what it documents is equally true for 5.x and 6.x, since they are both compatible with the same Python versions, and the docs don't say in which order it searches for Pythons.

README.md
```diff
@@ -1,9 +1,13 @@
...
-project used by the Chromium team and takes away the pain of dealing with the
-various differences in build platforms.
+`node-gyp` is a cross-platform command-line tool written in Node.js for
+compiling native addon modules for Node.js. It contains a fork of the
+[gyp](https://gyp.gsrc.io) project that was previously used by the Chromium
+team and takes away the pain of dealing with the various differences in build
+platforms.
+
+Note that `node-gyp` is _not_ used to build Node.js itself, for information
+on that, see https://github.com/nodejs/node.
```
richardlau: It feels that we should be pointing to more specific information if we say "for information on that" but I'm not aware that we have much (any?) documents in https://github.com/nodejs/node that talk about `gyp`. Maybe [BUILDING.md](https://github.com/nodejs/node/blob/master/BUILDING.md) is the most suitable link?

sam-github: I was weasely wording on purpose, because I don't think the node-gyp project should encode nodejs/node documentation locations into its README. I think we should say _something_ about it, because "node gyp" makes it sound like it's node's gyp.... and it isn't. How about I just terminate the sentence at the `,`, removing the entire ", for information ..."?

richardlau: > How about I just terminate the sentence at the `,`, removing the entire ", for information ..."?

👍 Works for me.

README.md
```diff
@@ -76,6 +81,14 @@ value:
 $ npm config set python /path/to/executable/python
 ```
 
+If the `PYTHON` environment variable is set to the path of a Python executable,
+it will be used if it is a compatible Python.
+
+If the `NODE_GYP_FORCE_PYTHON` environment variable is set to the path of a
+Python executable, it will be used instead of any of the other configured or
+builtin Python search paths. If its not a compatible Python, no further
+searching will be found.
```
cclauss: ~found~ done

sam-github: fixed, thanks

rvagg:
landed in b1bf32e, and will backport

PR #1911, opened 2019-10-04
v5.x Python compat fixes for README

rvagg:
When backporting #1908 I noticed that it leaves README in a weird 1/2 way state, still stating that it needs Python 2.7.

This PR adds #1908 plus additional changes required to make README match master. I don't believe there's anything different that needs to be said between the two at this stage.

The only minor difference is in the initial list of requirements, `master` has "`Python v2.7, v3.5, v3.6, or v3.7`", this one just has "Python v2.7, v3.5, v3.6, or v3.7" (no backticks). I'll PR that to master.

@cclauss @joaocgreis @sam-github any of you around to +1 this so I can get a 5.0.5 out with it?


rvagg:
added fixup commit to bring back 2.6 support in docs ... will need to do another PR to bring it back in find-python.js

rvagg:
removed mention of 2.6 in here, it's only "supported" in the sense the our `semverRange` has it included for 5.x

rvagg:
landed in v5.x as 3891391

PR #1913, opened 2019-10-04
Travis CI: Add the long dead Python 2.6

cclauss:
##### Checklist

- [x] tests are included 

##### Description of change

As discussed in #1912, we need to test software that we claim to support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ language: python
 cache: pip
 matrix:
   include:
+    - name: "Python 2.6 on Linux"
+      dist: trusty
+      env: NODE_GYP_FORCE_PYTHON=python2
+      python: 2.6
+      before_install: nvm install 10
     - name: "Python 2.7 on Linux"
       env: NODE_GYP_FORCE_PYTHON=python2
       python: 2.7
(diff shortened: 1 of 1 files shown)


rvagg:
remade the branch and cherry-picked the HEAD commit onto it so it's properly on v5.x now, will watch Travis to see what it says, regardless I don't think this is (a) needed or (b) requires us to document it, it's just not something we can change on v5.x because it's been baked in already.

cclauss:
I rebased on v5.x at same time you did so who knows what we got.  Let’s see what Travis says.  If it fails then I believe our release notes should reflect that.

cclauss:
Flake8 no longer supports Python 2.6 but even when flake8 is fully disabled, the Py26 run fails.

rvagg:
it failed because it's running Node 0.10 on that image, I switched to 10 (`nvm install 10`), but then it failed because it didn't have `semverRange`, so I rebased onto rvagg/restore-python-2.6, but then it's failing because the compilers is too old on precise: https://travis-ci.com/nodejs/node-gyp/jobs/242023105, but that's a pretty good sign that it actually works, or at least it's not failing because of Python 2.6, it's making the build resources so getting cleanly through `gyp`. Would be interesting to figure out how to make this run. Can we do 2.6 on a newer image than precise?

cclauss:
Try upgrading to __trusty__

rvagg:
green with trusty 👍 will merge this into v5.x

rvagg:
No, I won't add this, because it disables flake8 for everything, we'll need to differentiate before it can land. I've cleaned it up into a single commit on top of the current v5.x and force pushed it up.

PR #1919, opened 2019-10-14
Some small adjustments to the README.md for new users

dannypike:
##### Checklist

##### Description of change
Documentation that expresses opinions rather than stating facts can be inappropriate.

A user may well only look at the documentation for a product when she finds that it does not work as she expects. If, within the first few lines of reading the introduction, she comes across a line saying "our product is easy to use", what is she to think? That the author believes that she must be stupid if she has not already got it to work? Or, worse still, the author thinks that the product is perfect and doesn't really need to be documented at all?

In my opinion. it is much better to keep to factual descriptions and helpful examples of typical usage in the product documentation. This is especially true of the first three paragraphs of the README. It shuuld avoid using words like "clearly", "obviously", "basically" and phrases like "all you have to do is...", "it is easy to use".

Such a README gives a much better impression to a new user, encouraging her to keep reading in the hope that the author may understand her predicament.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,7 @@
 `node-gyp` is a cross-platform command-line tool written in Node.js for
 compiling native addon modules for Node.js. It contains a fork of the
 [gyp](https://gyp.gsrc.io) project that was previously used by the Chromium
-team and takes away the pain of dealing with the various differences in build
-platforms.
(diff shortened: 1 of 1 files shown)


rvagg:
landed in 3538a31, sorry for the delay

Issue #1918, opened 2019-10-14
The 'tar' node package needs to be updated


jkim7:
Hi

Older versions of tar has a file called cb-never-called-1.0.1.tgz, and it triggers a warning on virus scans, could you guys upgrade to a newer version of tar to remove it?

thanks

* **Node Version**: 
* **Platform**: 
* **Compiler**: 
* **Module**: 

<details><summary>Verbose output (from npm or node-gyp):</summary>

```
Paste your log here, between the backticks. It can be:
  - npm --verbose output,
  - or contents of npm-debug.log,
  - or output of node-gyp rebuild --verbose.
```

</details>

rvagg:
Please use a newer version of node-gyp, or npm if you are using it from there. This should already have been resolved.

rvagg:
OK, I take that back. I don't know about the specific issue you're pointing out but it looks like a v5 of node-tar has been released. We should look at that.

rvagg:
It looks like this was resolved in node-tar v3, https://github.com/npm/node-tar/pull/72#issuecomment-300347778, we are on to v4 and I can't find the file in the package. So, indeed, upgrade your node-gyp or npm and it should go away. I'll deal with a node-tar upgrade separately.

PR #1920, opened 2019-10-15
update deps to roughly match current npm@6

rvagg:
Some overdue dependency updates. I've had a look through the details of most of these updates and I'm not finding anything that would likely impact us. I'm also removing the broad dependency options (`||`) which used to be useful when npm was shipping very different versions. I believe they're just shipping version 6 now. Most of the changes here are matching the dependencies that the current npm master is using. I think with these changes, `env-paths` might be the only thing we're pulling in to npm when bundled.

There's newer versions of some of these, like tar (this update was prompted by #1918), but npm isn't shipping that yet. I suspect npm is holding back on a lot of updates to maintain older Node compatibility (the README says v6 but .travis.yml says v8).

I believe these changes can be safely backported to v5.x.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,26 +22,26 @@
   "bin": "./bin/node-gyp.js",
   "main": "./lib/node-gyp.js",
   "dependencies": {
-    "env-paths": "^1.0.0",
-    "glob": "^7.0.3",
-    "graceful-fs": "^4.1.2",
-    "mkdirp": "^0.5.0",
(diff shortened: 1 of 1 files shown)


joaocgreis:
Rebased and force-pushed to clear the Windows Travis failure.

CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/167/

rvagg:
c60c22d

PR #1952, opened 2019-11-02
Fix toolkit selection for VS 2017/2019

haxtibal:
##### Checklist

##### Description of change
`npm install --build-from-source sqlite3` failed for me in a fresh Visual Studio 2019 Build Tools environment, because node-gyp fell back to platform tools v140 which are not installed in VS 2019 by default, but v142 is. The error message was something like
```
The builds tools for v140 (Platform Toolset = 'v140') cannot be found.
```
This patch basically repeats some lines that @refack has already in GYP3 to default to toolset `v141` for VS2017 and `v142` for VS2019, see
- refack/GYP3@c40235ddad9f588d50ab79f6a8d1fca17717d8d0
- refack/GYP3@703706c4995b4c9ed0d5f319d684ba52647ab255
- refack/GYP3@eb296f67da078ec01f5e3a9ea9cdc6d26d680161

In case you're already about to replace GYP with GYP3, sorry for the noise, just ignore this PR.

Disclaimer:

- Registry based detection in `MSVSVersion._DetectVisualStudioVersions` is still incomplete. Maybe it's even in GYP3, because [registry handling has changed](https://developercommunity.visualstudio.com/comments/539856/view.html) in VS2017/19 and keys are stored in a private location now. I didn't bother much, it works in node-gyp context anyway because node-gyps configure.js does it's own detection based on `VCINSTALLDIR` and sets `GYP_MSVS_OVERRIDE_PATH` and `GYP_MSVS_VERSION` accordingly. Btw., why is detection done in both the JS part and GYP, shouldn't it go to GYP solely for separation of concerns?

- I guessed `project_version='16.0'` for 2019, but it seems the `Version` element is not meaningful anymore at all in MSBuild project files?

### DIFF
--- a/gyp/pylib/gyp/MSVSVersion.py
+++ b/gyp/pylib/gyp/MSVSVersion.py
@@ -236,6 +236,24 @@ def _CreateVersion(name, path, sdk_based=False):
   if path:
     path = os.path.normpath(path)
   versions = {
+      '2019': VisualStudioVersion('2019',
+                                  'Visual Studio 2019',
+                                  solution_version='12.00',
+                                  project_version='16.0',
+                                  flat_sln=False,
+                                  uses_vcxproj=True,
+                                  path=path,
+                                  sdk_based=sdk_based,
+                                  default_toolset='v142'),
+      '2017': VisualStudioVersion('2017',
+                                  'Visual Studio 2017',
+                                  solution_version='12.00',
+                                  project_version='15.0',
+                                  flat_sln=False,
+                                  uses_vcxproj=True,
+                                  path=path,
+                                  sdk_based=sdk_based,
+                                  default_toolset='v141'),
       '2015': VisualStudioVersion('2015',
                                   'Visual Studio 2015',
                                   solution_version='12.00',
@@ -358,6 +376,8 @@ def _DetectVisualStudioVersions(versions_to_check, force_express):
       2012(e) - Visual Studio 2012 (11)
       2013(e) - Visual Studio 2013 (12)
       2015    - Visual Studio 2015 (14)
+      2017    - Visual Studio 2017 (15)
+      2019    - Visual Studio 2019 (16)
     Where (e) is e for express editions of MSVS and blank otherwise.
   """
   version_to_year = {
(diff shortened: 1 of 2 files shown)


joaocgreis:
Visual Studio detection is completely [done in node-gyp](https://github.com/nodejs/node-gyp/blob/68319a2c344c822d48bd6d5dd32f82dd41384e19/lib/find-visualstudio.js#L292-L296). Gyp is [forced to use the detected version](https://github.com/nodejs/node-gyp/blob/68319a2c344c822d48bd6d5dd32f82dd41384e19/lib/configure.js#L147), completely bypassing the detection mechanism there. We do this to be able to provide good logging output, filter Visual Studio installations based on ability to compile C++ and support by the Node.js version being used.

I was surprised when I was able to reproduce your issue. `sqlite3` indeed does not compile, but other modules (`bcrypt`, `leveldown`, ...) compile without issues. I investigated a bit and found that this is an issue in the `sqlite3` module, it overwrites the variable node-gyp uses to specify the toolset.

PR in `sqlite3` with a fix: https://github.com/mapbox/node-sqlite3/pull/1242

This PR should not land here, and I don't know how well this behaves when multiple versions of VS are installed. I don't have strong feelings about the Gyp changes, but managing Gyp has been a problem since Google stopped supporting it so it's probably better that we change it as little as possible. Modules affected by this problem should be fixed instead. @haxtibal thanks for opening a PR and actually working toward solving the issue!

haxtibal:
> PR in sqlite3 with a fix: mapbox/node-sqlite3#1242

@joaocgreis That will solve my issue, thank's a lot for investigating.

> Modules affected by this problem should be fixed instead.

Addon writers are [officially instructed](https://nodejs.org/api/addons.html#addons_building) to create a `binding.gyp`, i.e. use the GYP language. They may well assume they have full GYP functionality available, including detection. Would it make sense to fix the documentation in first place (telling what works and what not), and only then the packages?

joaocgreis:
I believe GYP works pretty much as expected. If you search for `defaults` and `variables` in https://github.com/nodejs/node-gyp/blob/master/lib/configure.js, there are a few more that node-gyp passes to GYP. So, node-gyp controlling some parts of the process is not unexpected.

@haxtibal I think it would be good to add a note in the documentation if you want to do that. Perhaps in https://github.com/nodejs/node-gyp/tree/68319a2c344c822d48bd6d5dd32f82dd41384e19#the-bindinggyp-file as it may be too specific for the core docs, but others may disagree so feel free to try a PR there.

joaocgreis:
PR in node-sqlite3 landed, we can close this here but let me know if there are still issues. Thanks!

Issue #1791, opened 2019-06-21
The GYP issue: where, how and who to maintain


rvagg:
So we have https://github.com/refack/GYP, I don't know what its current state is but I know things have been merged here straight into gyp/ that probably conflict with it so I assume we're in a messy out-of-sync state.

https://github.com/nodejs/admin/issues/247 was closed, I guess @refack doesn't want to transfer it to the org. IMO that's in no way a prerequisite to using it as our GYP upstream so I don't consider that a problem if @refack wants to continue maintaining it under his own name.

I'd just like clarity on where GYP changes are going, who approves them, how we keep things in sync here, with an upstream and with nodejs/node.

@refack do you still have the stomach to continue with maintaining it?

@nodejs/python some input from you folks too would be great. This is well outside my comfort zone and I'd rather node-gyp be concerned with wrapping GYP, not being the primary place it's maintained. That separation of concerns would also make issues and PRs here easier to manage and we have an appropriate place to point people who want to make GYP changes instead of everything getting tangled.

391861737:
I can find nodejs/node#26620 is on progress... So I think we should wait for it?

rvagg:
I think @refack may have lost interest in helping maintain this tooling and we might be needing an alternative solution. We also have gyp.js and ninja.js in the picture, I'm about to put up a pull request for including gyp.js (the 3rd pull request to date) and see if we can work toward an opt-in for it and gradually work to full compatibility. But we need a strategy for maintaining it too, I don't believe @indutny is interested in pursuing these two projects and it'd be good to have a collective space for a group of people to maintain them.

For now I was thinking that setting up a new lightweight github org to maintain a collection of gyp and gyp-related tools might be a good approach, at least for the short term. Pull in GYP / GYP3 (or some synthesis of what's been done here, in nodejs/node and refack/GYP), gyp.js and ninja.js if @indutny wants to give them up, and use it as an upstream for node-gyp and anything else using gyp (nodejs/node, libuv?). I'd love to hear others' thoughts on this.

391861737:
LGTM

indutny:
I'd be happy to transfer both `gyp.js` and `ninja.js` to `nodejs` org upon request.

bnoordhuis:
My preference would be to put our efforts into gyp.js and get rid of the python-based gyp, at least for add-ons.

[...]

PR #1964, opened 2019-11-11
add msvs 2017 support

dothebart:


richardlau:
Based on https://github.com/nodejs/node-gyp/pull/1952#issuecomment-549631126 this isn't necessary? cc @joaocgreis

dothebart:
after opening the PR I saw that these changes already there in GYP3. However, as long as you use the node-gyp this is needed.

joaocgreis:
This shouldn't be needed, VS detection is not done by GYP but in https://github.com/nodejs/node-gyp/blob/master/lib/find-visualstudio.js.

@dothebart what issue did you see that needs this? Is the detection in node-gyp not working for you? Can you paste the output of a failing `node-gyp rebuild --verbose` here?

dothebart:
Ah, I don't compile V8 for nodejs - its https://github.com/ArangoDB/ArangoDB - so I don't have this file.
I now continued with gyp3, which has this. So if you don't want it, close it.

joaocgreis:
I guess we'll have to land something like this or remove autodetection from the GYP side, but I'd rather wait until we figure out how GYP will be managed here. @dothebart thanks for opening this PR and trying to help.

dothebart:
hm, asking a heretic question - how do you run `.js` before having a ready compiled V8?

joaocgreis:
I'm not sure I understand your question... Here in node-gyp, Node.js is required to run, so V8 is always already compiled.

For compiling Node.js itself in https://github.com/nodejs/node/ a different mechanism is used. It uses `vswhere` from the command line: https://github.com/nodejs/node/blob/ed401236f6988c6ad62a21dbbf808da83782bee0/vcbuild.bat#L247

dothebart:
ah, nice. a shellscript to tell the environment autodetection infrastructure about the infrastructure to work around the autodetection. 
Is Gyp3 the way ahead?

joaocgreis:
There's discussion about the way forward in https://github.com/nodejs/node-gyp/issues/1791. There was a PR to change to GYP3 https://github.com/nodejs/node-gyp/pull/1845 but it was closed. From what I see, it looks like we'll be maintaining GYP here specifically to be used in node-gyp at least for a while.

PR #1975, opened 2019-11-20
gyp: sync code base with nodejs repo

targos:
This syncs the gyp code base so that it is identical to the one in nodejs/node.

This was done by removing the code in nodejs/node-gyp, copying from nodejs/node and then manually reviewing and adapting the diff.


gyp/pylib/gyp/generator/make.py
```diff
@@ -1951,13 +1951,11 @@ def _InstallableTargetInstallPath(self):
     """Returns the location of the final output for an installable target."""
     # Xcode puts shared_library results into PRODUCT_DIR, and some gyp files
     # rely on this. Emulate this behavior for mac.
-
-    # XXX(TooTallNate): disabling this code since we don't want this behavior...
```
rvagg: note the TODO here specifically for node-gyp, I wonder how much this matters still? is it just a case of disallowing end-user from specifying `shared_library` in their binding.gyp? 🤷‍♂

rvagg:
Very brave of you @targos.
I feel far from qualified to review this properly. Hopefully @cclauss @joaocgreis @bnoordhuis and others could pitch in?
It seems to me that there are a lot of formatting changes that can be ignored. There's a ton of Windows changes which I believe were applied to nodejs/node but are not necessary here because we make up for it with JS, but wouldn't hurt to be here. The rest is the grey stuff that it would be good to have some knowledgeable eyes over.
We could bump to v7 to release this since there's a lot of changes and I suspect we wouldn't be able to guarantee this _doesn't_ break existing use.

gyp/pylib/gyp/generator/make.py
targos: Change was done in https://github.com/nodejs/node-gyp/commit/a2ed0df84e294a197e3eb36b8213db6d8116028c

If we need to keep it here, can we also apply it to node? I don't know the implications...

targos:
Most of the changes here come from upstream gyp updates that were applied to nodejs/node two to three years ago.

gyp/pylib/gyp/common.py
```diff
@@ -436,6 +436,8 @@ def GetFlavor(params):
     return 'solaris'
   if sys.platform.startswith('freebsd'):
     return 'freebsd'
+  if sys.platform.startswith('dragonfly'):
+    return 'freebsd'
```
cclauss:
```python
  if sys.platform.startswith(('freebsd', 'dragonfly')):
    return 'freebsd'
```

gyp/pylib/gyp/generator/make.py
targos: I don't know what to do with this. The change is in node-gyp since 2012, the commit message doesn't explain why it was done and I can't find the pull request that introduced it.

targos: I'll try to remove this code in nodejs/node and see if CI is happy. Maybe the gyp files that rely on this are not ours.

targos: Well, removing these lines breaks the Node.js build: https://ci.nodejs.org/job/node-test-commit-linux-containered/16453/nodes=ubuntu1804_sharedlibs_shared_x64/console
So we're back to the original question: how does it matter to node-gyp?
@TooTallNate sorry to bother you, but since you committed the change, maybe you remember why it was needed?

joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/169/

targos:
I'll need help, because it doesn't seem like the actual error message from MSBuild.exe was printed to the console.

rvagg:
how about we bump semver-major with this and go with a node-gyp v7. node-gyp 5 is still being used by npm so 6 is currently for people using it manually, a bump to 7 gives us some breakage room and lets people opt-in to an upgrade, then npm could choose to jump to v7 at some point if it wants instead of v6.

gyp/pylib/gyp/generator/msvs.py
```diff
@@ -294,20 +302,29 @@ def _ConfigFullName(config_name, config_data):
...
-              r'HKLM\Software\Wow6432Node\Microsoft\Microsoft SDKs\Windows\%s']:
-    sdkdir = MSVSVersion._RegistryGetValue(key % ver, 'InstallationFolder')
-    if not sdkdir:
-      continue
-    version = MSVSVersion._RegistryGetValue(key % ver, 'ProductVersion') or ''
-    # find a matching entry in sdkdir\include
-    names = sorted([x for x in os.listdir(r'%s\include' % sdkdir) \
-                    if x.startswith(version)], reverse = True)
-    return names[0]
+def _ConfigWindowsTargetPlatformVersion(config_data, version):
```
joaocgreis:
```suggestion
def _ConfigWindowsTargetPlatformVersion(config_data, version):
  target_ver = config_data.get('msvs_windows_target_platform_version')
  if target_ver and re.match(r'^\d+', target_ver):
    return target_ver
```

@targos sorry for the delay getting to this. This PR always fails when using VS2017 (not only Windows 2016). I see `msvs_windows_target_platform_version` was renamed to `msvs_windows_sdk_version`. However, the semantics here changed a bit so simply renaming in `lib/configure.js` does not work. I suggest to bring back the old variable name with old semantics, if this passes in CI. I'm not sure how well this works outside node-gyp/node, but I don't know if that's relevant.

targos:
Re-synced. Thanks @joaocgreis for the fix!

GitHub Actions passed.
Here's a CI run on Jenkins: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/175/

targos:
All green!

@nodejs/node-gyp PTAL and merge if you like :)

From my point of view, this can be semver-major. I only care about unifying the code between this repo and node core, so it doesn't matter in which release this goes first.

gyp/pylib/gyp/generator/make.py
gengjiawen: Will this be a blocker ? or we can merge this for now and left as it is ?

targos: I think this is not a blocker anymore.

gengjiawen:
@rvagg let's merge this to master ?

gengjiawen:
@rvagg let's merge this to master ?

joaocgreis:
Sorry for not dismissing my review earlier, I missed it when I investigated the failure.

CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/176/ ✔️ (this does not rebase, would be good to run CI on the final rebased version before landing)

gengjiawen:
> Sorry for not dismissing my review earlier, I missed it when I investigated the failure.
> 
> CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/176/ ✔️ (this does not rebase, would be good to run CI on the final rebased version before landing)

Looks like the CI passed. If we have no future questions, I want to land this.

gengjiawen:
@rvagg I have launched the PR. If you have any problems, feel free to revert this.

@targos I think we continue to work on nodejs/node#30563.

PR #1982, opened 2019-12-02
GitHub Action: Run Python tests on Ubuntu & macOS

cclauss:
Running Python standalone tests on multiple OSes would free up Travis CI for tests of various combinations of Node.js and Python as well as tests on other [CPU architectures](https://docs.travis-ci.com/user/multi-cpu-architectures).
__arch: amd64, arm64, ppc64le, s390x__

##### Checklist

- [x] tests are included 

### DIFF
--- a/.github/workflows/Python_tests.yml
+++ b/.github/workflows/Python_tests.yml
@@ -24,7 +24,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install flake8 pytest  # -r requirements.txt
       - name: Lint with flake8
-        if: matrix.os == "ubuntu-latest"
+        if: matrix.os == 'ubuntu-latest'
         run: |
           # stop the build if there are Python syntax errors or undefined names
           flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
--- a/.github/workflows/Python_tests.yml
+++ b/.github/workflows/Python_tests.yml
@@ -24,7 +24,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install flake8 pytest  # -r requirements.txt
       - name: Lint with flake8
-        if: ${{ matrix.os }} == "ubuntu-latest"
+        if: matrix.os == "ubuntu-latest"
         run: |
           # stop the build if there are Python syntax errors or undefined names
           flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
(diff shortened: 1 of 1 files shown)


rvagg:
I'd love if we could push as much as possible to GitHub Actions with a view to one day retiring Travis. What exactly can't we do on Actions yet? Is it just the non-x64 architectures? When they launch self-hosted runners we could start using Node infra to do these ourselves. Not sure what the timeline is on that though.

cclauss:
[using-self-hosted-runners-in-a-workflow](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-self-hosted-runners-in-a-workflow)

rvagg:
nice!

sadly they're still being restrictive on OS/arch support: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-self-hosted-runners#supported-operating-systems-for-self-hosted-runners

but they did talk about making this supported on more permutations and I bet IBM are already talking to them about adding their platforms.

cclauss:
Can I please get you review on this in its current state?  I would be interested to land is so we can start exploring the three todos.

.github/workflows/Python_tests.yml
```diff
@@ -0,0 +1,36 @@
...
+name: Python_tests
+on: [push]
+jobs:
+  Python_tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      max-parallel: 15
+      matrix:
+        os: [macos-latest, ubuntu-latest]  # , windows-latest]
```
gengjiawen: Why Windows not get supported ? A block from github action ?

.github/workflows/Python_tests.yml
```diff
@@ -0,0 +1,36 @@
+# TODO: Line 14, enable os: windows-latest
+# TODO: Line 15, enable python-version: 3.5
+# TODO: Line 36, enable pytest --doctest-modules
+
+name: Python_tests
+on: [push]
```
gengjiawen: I think this can be enabled `pull_request` too.

.github/workflows/Python_tests.yml
cclauss: There are three todos at the top of the file that we can work on after it lands.  This is one of them because the Windows doctests fail because they try to exersize some dead code that tries to the version of VSCode or similar which we now do elsewhere.

rvagg:
5a64e9bd3

PR #1985, opened 2019-12-03
Fix typo in README.md

suraneti:
##### Checklist

- [ ] tests are included 


gengjiawen:
Thanks for the contribution :)

rvagg:
@gengjiawen prefix needs to be added to commit messages and messages should not start with upper-case, same rules as nodejs/node I think.
I'll fix this one up with a `doc: fix typo ...` and force push, so just for future reference. Thanks for staying on top though.

PR #1992, opened 2019-12-15
doc: update macOS_Catalina.md

jameshome:
The fixes suggested in macOS_Catalina.md are not working reliably in the latest versions of Catalina. @rrrix found that installing the Command Line Tools from dmg rather than xcode-select --install does work. This was confirmed by two other people, including myself.

##### Checklist

### DIFF
--- a/macOS_Catalina.md
+++ b/macOS_Catalina.md
@@ -1,39 +1,37 @@
 # Installation notes for macOS Catalina (v10.15)
 
-_This document specifically refers to upgrades from previous versions of macOS to Catalina (10.15). It should be removed from the source repository when Catalina ceases to be the latest macOS version or updated to deal with challenges involved in upgrades to the next version of macOS._
+_This document specifically refers to upgrades from previous versions of macOS to Catalina (10.15). It should be removed from the source repository when Catalina ceases to be the latest macOS version or when Catalina has been fixed to solve these problems._
 
-Lessons learned from:
-* https://github.com/nodejs/node-gyp/issues/1779
-* https://github.com/nodejs/node-gyp/issues/1861
-* https://github.com/nodejs/node-gyp/issues/1927 and elsewhere
-
-Installing `node-gyp` on macOS can be found at https://github.com/nodejs/node-gyp#on-macos
-
-However, upgrading to macOS Catalina changes some settings that may cause normal `node-gyp` installations to fail.
+**Upgrading to macOS Catalina may cause normal `node-gyp` installations to fail.**
 
 ### Is my Mac running macOS Catalina?
-Let's make first make sure that your Mac is currently running Catalina:
-% `sw_vers`
+Let's first make sure that your Mac is running Catalina:
+```
+% sw_vers
     ProductName:	Mac OS X
     ProductVersion:	10.15
     BuildVersion:	19A602
+```
+If `ProductVersion` is less then `10.15` then this document is not for you. Normal install docs for `node-gyp` on macOS can be found at https://github.com/nodejs/node-gyp#on-macos
 
-If `ProductVersion` is less then `10.15` then this document is not really for you.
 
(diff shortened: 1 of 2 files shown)


macOS_Catalina.md
```diff
@@ -52,18 +50,23 @@ There are two main ways to install `node-gyp` on macOS:
...
-### Installing `node-gyp` using the Xcode Command Line Tools
+### Installing `node-gyp` using the Xcode Command Line Tools via `xcode-select --install`
 1. If the _acid test_ has not succeeded, then try `xcode-select --install`
 2. Wait until the install process is _complete_.
 3. `softwareupdate -l`  # No listing is a good sign.
     * If Xcode or Tools upgrades are listed, use "Software Update" to install them.
 4. `xcode-select -version`  # Should return `xcode-select version 2370` or later.
 5. `xcode-select -print-path`  # Should return `/Library/Developer/CommandLineTools`
-6. Try the [_acid test_ steps above](#The-acid-test) to see if your Mac is ready.
+6. Try the [_acid test steps above](#The-acid-test) to see if your Mac is ready.
```
cclauss: This starts _italics_ but never finishes it.

jameshome: Thanks for catching that.

macOS_Catalina.md
```diff
@@ -1,39 +1,37 @@
...
 
-If `ProductVersion` is less then `10.15` then this document is not really for you.
 
 ### The acid test
-Next, lets see if `Xcode Command Line Tools` are installed:
+To see if `Xcode Command Line Tools` is installed in a way that will work with `node-gyp`, run:
 1. `/usr/sbin/pkgutil --packages | grep CL`
-    * If nothing is listed, then [skip to the next section](#Two-roads).
-    * If `com.apple.pkg.CLTools_Executables` is listed then try:
+    * This should return `com.apple.pkg.CLTools_Executables`. If it doesn't, this test failed.
```
rvagg: it's likely to return more than just that, "is listed" is more accurate than "should return"

macOS_Catalina.md
```diff
@@ -1,39 +1,37 @@
 # Installation notes for macOS Catalina (v10.15)
 
-_This document specifically refers to upgrades from previous versions of macOS to Catalina (10.15). It should be removed from the source repository when Catalina ceases to be the latest macOS version or updated to deal with challenges involved in upgrades to the next version of macOS._
+_This document specifically refers to upgrades from previous versions of macOS to Catalina (10.15). It should be removed from the source repository when Catalina ceases to be the latest macOS version or when Catalina has been fixed to solve these problems._
```
rvagg: I'm a little uncomfortable with "has been fixed", can we say for sure this is a macOS _bug_ that needs to be fixed and not just some side effect of the way they want people to interact with their systems? I still haven't upgraded the mac I have to Catalina so haven't gone through any of this pain so my perspective is more limited. If you (and @cclauss) are comfortable asserting that this is something deserving of a "fix" and not just something they could address to help deal with the challenges of dealing with a diverse ecosystem then that's fine. Otherwise, this should be softened a little I think.

rvagg:
will defer to @cclauss on final approval of this

macOS_Catalina.md
jameshome: Fair point. Substantially softened.

macOS_Catalina.md
jameshome: I changed both lines back to "is listed". Thanks for the clarification.

jameshome:
@cclauss Anything more you need from me on this? There have been another half dozen posts to [1927](https://github.com/nodejs/node-gyp/issues/1927), so this is impacting a lot of folks right now.

gengjiawen:
Thanks for the contribution ❤️

PR #2003, opened 2019-12-22
Update README.md

crydotsnake:
It is always better, to download it from the official Website :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Depending on your operating system, you will need to install:
 
 ### On Windows
 
-Install the current version of Python from the [Microsoft Store package](https://docs.python.org/3/using/windows.html#the-microsoft-store-package).
+Install the current version of Python from the [Python Website](https://www.python.org/downloads/windows/).
 
 #### Option 1
 


cclauss:
My sense is that installing from Microsoft's store is a better way to install the current version of Python is a way where it will be easily updated as Python progresses.

richardlau:
Maybe instead of replacing we could list both options?

README.md
```diff
@@ -43,7 +43,7 @@ Depending on your operating system, you will need to install:
 
 ### On Windows
 
-Install the current version of Python from the [Microsoft Store package](https://docs.python.org/3/using/windows.html#the-microsoft-store-package).
+Install the current version of Python from the [Python Website](https://www.python.org/downloads/windows/).
```
richardlau: Maybe instead of replacing we could list both options?

cclauss: I do not see the advantage of selecting the python.org option but I do see the downside of selecting it.

crydotsnake: I agree with @richardlau

PR #2006, opened 2019-12-29
Update to semver ≥ 7 (Fixes: #2005)

### LINKED ISSUE #2005: Update to semver ≥ 7
* **Node Version**: 10.17.0
* **Platform**: Linux/Debian
* **Compiler**: gcc 9
* **Module**: 

node-gyp is not compatible with semver ≥ 7. I found this little fix that seems to fix the problem:

```diff
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -226,7 +226,7 @@
       }
       this.addLog(`- version is "${version}"`)

-      const range = semver.Range(this.semverRange)
+      const range = new semver.Range(this.semverRange)
       var valid = false
       try {
         valid = range.test(version)

```

guimard:
##### Description of change

This PR updates semver to 7.x. and fixes #2005

### DIFF
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -226,7 +226,7 @@ PythonFinder.prototype = {
       }
       this.addLog(`- version is "${version}"`)
 
-      const range = semver.Range(this.semverRange)
+      const range = new semver.Range(this.semverRange)
       var valid = false
       try {
         valid = range.test(version)


rvagg:
I'm not keen on this, our package.json has been optimised to fit nicely with npm's so that there's a lot of deduplication. Both node-gyp and npm currently have `^5.7.1` (https://github.com/npm/cli/blob/797a59797a2e226631f068c912f351ba340581fe/package.json#L130). I don't understand the need to push it to 7 or why this lack of support for 7 is a problem for anyone, as I commented in #2005.

guimard:
Hi @rvagg,

I'm [Debian](https://www.debian.org) Developer. For security reason, our policy does not accept embedded code. That's why there is only one semver in our distribution. I posted here my patch published in our package.
Note that node-gyp accepted it.

Cheers,
Xavier

rvagg:
(Following on from discussion in #2006)

The `new` should work with both 5 and 7 and you're going to force semver@7 when you package this upstream anyway, whether we change package.json or not.

So back out the package.json change, leaving us with semver@5, and I'll lift my objection. Without anyone else objecting we should be able to merge this (with a bit of manual testing). We'll bump to semver@7 when npm does--which may be sooner than later with the retirement of Node.js 8 (i.e. a lot of npm's dependencies have moved to support newer JS features that aren't supported by Node.js 8).

I'll +1 this change since it's trivial, but it's not an endorsement (from me at least) of dancing around the incompatible bundling policies of Debian et. al.. That's something you either have to continue to manually patch your way around or push for some policy leniency for runtime ecosystems such as Node.js, whose dependency systems are incompatible with the existing distro policies.

guimard:
> (Following on from discussion in #2006)
> 
> The `new` should work with both 5 and 7 and you're going to force semver@7 when you package this upstream anyway, whether we change package.json or not.
> 
> So back out the package.json change, leaving us with semver@5, and I'll lift my objection. Without anyone else objecting we should be able to merge this (with a bit of manual testing). We'll bump to semver@7 when npm does--which may be sooner than later with the retirement of Node.js 8 (i.e. a lot of npm's dependencies have moved to support newer JS features that aren't supported by Node.js 8).
> 
> I'll +1 this change since it's trivial, but it's not an endorsement (from me at least) of dancing around the incompatible bundling policies of Debian et. al.. That's something you either have to continue to manually patch your way around or push for some policy leniency for runtime ecosystems such as Node.js, whose dependency systems are incompatible with the existing distro policies.

Done, thanks.

Of course, Debian Developer will continue to manually patch the node.js module they need :wink:

rvagg:
landed f242ce4

PR #2011, opened 2020-01-03
doc: remove backticks from Python version list

rvagg:
This has been bothering me for a while now, it was fixed on `v5.x` and is the main difference in the README between the two. The backticks have a complicated history in this list and never got properly cleaned up.


rvagg:
landed 9a7dd16

PR #2018, opened 2020-01-08
lib: ignore VS instances that cause COMExceptions

amcasey:
##### Checklist

  Some tests are failing, but they also fail without my change.  I assume the CI system will catch actual failures.
- [ ] tests are included 
  I see that find-visualstudio has pretty good test coverage, but I didn't see an obvious way to simulate this COM failure
  Are there user-facing docs for find-visualstudio?  I'd be happy to add a blurb about ignoring certain instances.

##### Description of change
I have quite a few instances of VS installed and it looks like Find-VisualStudio.cs enumerates all of them, even when find-visualstudio.js already knows which one it wants (from the environment variable in the developer command prompt).  One of them (from 15.7.2, if that's interesting) causes a COMException on the ISetupInstance2.GetPackages call.  Ignoring such packages seems harmless and unblocks the rest of the run.

Note: an alternative approach would be to print a JSON blob with an empty packages list and let the caller reject it with an explicit error message.

### DIFF
--- a/lib/Find-VisualStudio.cs
+++ b/lib/Find-VisualStudio.cs
@@ -209,7 +209,7 @@ namespace VisualStudioConfiguration
                 {
                     instances.Add(InstanceJson(rgelt[0]));
                 }
-                catch
+                catch (COMException)
                 {
                     // Ignore instances that can't be queried.
                 }
--- a/lib/Find-VisualStudio.cs
+++ b/lib/Find-VisualStudio.cs
@@ -205,7 +205,14 @@ namespace VisualStudioConfiguration
                     return;
                 }
 
-                instances.Add(InstanceJson(rgelt[0]));
+                try
+                {
+                    instances.Add(InstanceJson(rgelt[0]));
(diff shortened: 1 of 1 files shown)


joaocgreis:
@amcasey thanks for opening this PR! Ideally we would print an error message mentioning the COM issue, but I'll happily take this as is.

One thing though, I fear this might make any other problem in InstanceJson extremely confusing to debug. Can you catch the COMException explicitly?

amcasey:
@joaocgreis Works for me. :smile:

joaocgreis:
CI: https://ci.nodejs.org/view/All/job/nodegyp-test-pull-request/174/ ✔️ 

The Travis failures are unrelated.

I'll land this early next week (we need to wait at least 48h).

joaocgreis:
Landed in https://github.com/nodejs/node-gyp/commit/1f7e1e93b5116e727d0719dcca87ee38c18042ae

Thanks @amcasey

PR #2061, opened 2020-03-06
doc: add Chocolatey instructions for Windows

kripod:
##### Checklist

##### Description of change

Added instructions for installing Windows build tools through Chocolatey.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Depending on your operating system, you will need to install:
 
 ### On Windows
 
-Install the current version of Python from the [Microsoft Store package](https://docs.python.org/3/using/windows.html#the-microsoft-store-package).
+Install the current version of Python from the [Microsoft Store package](https://docs.python.org/3/using/windows.html#the-microsoft-store-package) or through [Chocolatey](https://www.chocolatey.org/packages/python/).
 
 #### Option 1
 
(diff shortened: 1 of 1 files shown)


cclauss:
I specifically removed this alternative.  Why do we want to put it back it?

Why do we want to have to support multiple ways of configuring (or misconfiguring) Windows now that Microsoft has invested heavily to put the current Python into their app store?

cclauss:
I specifically removed this alternative installation of Python.  What is the advantage of us taking on the additional support burden two alternative ways for our users to configure (or mis-configure) Windows now that Microsoft has invested to put the current Python into the Windows AppStore?

kripod:
For development purposes, Chocolatey is a more versatile tool, offering packages like the Visual Studio Build Tools, unlike the Microsoft Store.

bzoz:
I'm generally +1 on this, choco is already supported by Node. But I would prefer to have this as a separate section, not interleaved with other methods. A separate list of commends, like:
```
choco install python -y
choco install visualstudio2017buildtools -y
choco install visualstudio2017-workload-vctools -y
```

Also, we should provide a link to the choco installer.

cclauss:
I still do not see the advantage of us supporting a third party install when the operating system vendor has created a supported install.

carlo-quinonez:
Advantage to choco vs store installs is scriptability and automation.

cclauss:
Python is a scripting language ;-).  Like what kind of stuff?

bzoz:
@cclauss I guess @carlo-quinonez meant that you can use choco to install python via automation scripts

cclauss:
> install python via automation scripts

Same w/ the Microsoft version https://docs.python.org/3/using/windows.html#installing-without-ui plus the OS will know how to do upgrades.  For example, Python 3.8.3 is due this weekend.

bzoz:
I think those commands are for the MSI installer, they don't seem to work with the Windows Store one.

bzoz:
I'm still -1 on this in this form: https://github.com/nodejs/node-gyp/pull/2061#issuecomment-598262012

rvagg:
I'm going with @bzoz' advice on this. This PR should be adjusted to have a separate choco section, it can't land as it is now.

cclauss:
Perhaps choco should be added to https://docs.python.org/3/using/windows.html before it is added to this repo.

Issue #2074, opened 2020-03-17
Upgrade mkdirp deps to 1.0.3 to fix Critical CVE


mleneveut:
The package mkdir 0.5.1 contains a dependency to minimist 0.0.8, which has the CVE-2020-7598, scored 9.8

* **Node Version**: 
* **Platform**: 
* **Compiler**: 
* **Module**: 

https://nvd.nist.gov/vuln/detail/CVE-2020-7598

https://github.com/npm/cli/issues/1027

rvagg:
Ugh, this is not "critical". So typical of these reports to be overhyped by reporters who have a business interest in the hype.

node-gyp doesn't use minimist _at all_ even though it might be in the tree, and the vulnerability is _super_ hypothetical since the whole point of minimist is to parse CLI arguments so the attack vector involves getting something to execute via a shell. And it's a ReDOS, so not "critical" inasmuch as it's useful for anything beyond a DOS if you could even pull that off.

We try and keep our dependencies roughly in line with npm's so we don't end up inserting a ton of dupes into npm's package list, so we might be best to wait till npm reacts: https://github.com/npm/cli/blob/d9a7b30f6afc0419ad4939bfb478ca8a589e769f/package.json#L94 although I suspect it might be safe to assume a `^1` bump is coming.

richardlau:
If we're willing to drop support for Node.js < 10 we could probably drop mkdirp completely and use recursive mkdir on fs.

rvagg:
@richardlau on master I'm +1 on that approach, we'll still have to deal with v5 though where we should probably be more careful

rvagg:
As I mentioned @ https://github.com/nodejs/node/issues/32296#issuecomment-600396915, it's been fixed in mkdirp 0.5.3 and we use a `^` range specifier without shipping a package-lock.json, so I don't think we need to do anything about this.

I'll put in a patch to the 5.x branch to formalise it, but I think we can dismiss this as _taken care of_.

Issue #2047, opened 2020-02-15
request has gone into maintenance mode. Maybe replace it.


Tchiller:
Dear team,

thanks for your work!

I would like to report that the package `request` which you are loading has went into maintenance mode https://github.com/request/request/issues/3142

Therefore, yarn is currently giving me: 

> 
> warning node-gyp > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142

Maybe you find time (and a good replacement) to replace it. 
Thanks!

rvagg:
Yeah, thanks @Tchiller, it got deprecated last week after a long warning period (mind you, I didn't pick up on the fact that it would be properly deprecated in npm so this was a bit of a surprise).

We don't have a plan to deal with this, maybe we'll follow what npm does (or is doing, I don't know status). This will take someone doing the work and making sure it's maximally compatible back to Node 10.x at least (yes 10 is going away soon but node-gyp can't be as aggessive as a lot of software since it's a piece of core infra).

rvagg:
Yeah, maybe not, npm uses its own library that's a wrapper around a fork of node-fetch, not an ideal tangle to get involved in https://github.com/npm/make-fetch-happen

We could consider `bent` but we have yet to embrace async/await in node-gyp. We're either going to have to do an awkward piece-wise migration or go whole-hog at some point.

Too hard to think about right now! Proposals are welcome!

imatlopez:
@rvagg are you saying that for node-gyp v7.x Node 6-9 are getting dropped? You could cage any calls to `request` in v6.x, then swap the underlying api for v7.x

rvagg:
What version of Node we support is partly informed by npm. I don't have much insight into their current strategy but recent years suggest that they're being more aggressive with dropping older versions of Node from support as those versions get dropped upstream.

I don't think there's any reason to consider supporting Node 6, it's roughly 1 year out of support.
Node 8 was dropped at the beginning of this year. There's an argument to be made that we should consider continuing to make node-gyp compatible with Node 8. I wouldn't want npm to decide to ship an older node-gyp because they wanted to be more compatible than us. But, at this point, the kinds of features we want to start adopting are in Node 8 anyway - async/await being the main one, so it probably wouldn't be hard.

[...]

PR #2075, opened 2020-03-18
lib: update to mkdirp@1 w/ Promises interface

rvagg:
Ref #2074

I did want to update to 0.5.3 but Isaac has deprecated all 0.x versions. To avoid the deprecation warning we need to go to @1 which has a Promises interface. We don't have any async/await in 5.x so I don't want to introduce it. We do have one use of Promises already so this isn't a major change. I'd just like some checking of my code!

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -73,18 +73,17 @@ function configure (gyp, argv, callback) {
 
   function createBuildDir () {
     log.verbose('build dir', 'attempting to create "build" dir: %s', buildDir)
-    mkdirp(buildDir, function (err, isNew) {
-      if (err) {
-        return callback(err)
-      }
-      log.verbose('build dir', '"build" dir needed to be created?', isNew)
-      if (win) {
-        findVisualStudio(release.semver, gyp.opts.msvs_version,
-          createConfigFile)
-      } else {
-        createConfigFile()
-      }
-    })
+    mkdirp(buildDir)
+      .catch(callback)
+      .then((isNew) => {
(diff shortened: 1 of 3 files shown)


rvagg:
Node 6 tests fail, mkdirp now uses varargs: `opts = { mode: 0o777 & (~process.umask()), fs, ...opts }`

So ... how do we feel about dropping Node 6 support in this old release line? Is 6 ancient enough to stop caring or should we preserve 6 support for those people stuck on it? Alternatives I see if we want to maintain 6 support are (a) implement our own `mkdirp` for Node versions that don't have `recursive`, our needs are simpler than the use-cases mkdirp covers or (b) stick with 0.5.3 which will fix the security warning but include a deprecation warning.

rvagg:
Was reminded by #2047 that we have a `request` deprecation in place anyway, so an additional deprecation probably isn't a big deal.

My inclination here is to use mkdirp 0.5.3 to avoid the security warning but swallow the deprecation warning and not impact our Node version compatibility. We can move on with node-gyp 7 (6 is probably finished now) where we can start using fs.mkdir recursive where available and switch away from request (undecided how yet).

It really depends on what npm wants to do, if they jump to mkdirp 1 then we should probably follow suit to dedupe.

@darcyclarke I hear you're the one handling this next npm release? Do you want to coordinate on this or will you just be bumping to 0.5.3 (which our semver range allows).

darcyclarke:
@rvagg I'm going to push out a release to `npm` in the next little bit here which should pull in the latest `mkdirp0.5.3` & resolve this.

PR #2078, opened 2020-03-27
Add docs for macOS Catalina Software Updates

karlhorky:
##### Checklist

##### Description of change

Every time I do a Software Update in macOS Catalina, the command-line tools are borked with node-gyp and I need to follow this "I did all that..." section. I have also seen this affect others.

I've added a few simple changes to the steps to guide users in case the `xcode-select --install` step fails.

### DIFF
--- a/macOS_Catalina.md
+++ b/macOS_Catalina.md
@@ -1,8 +1,8 @@
-# Installation notes for macOS Catalina (v10.15)
+# Installation notes for macOS Catalina (v10.15) Updates
 
 _This document specifically refers to upgrades from previous versions of macOS to Catalina (10.15). It should be removed from the source repository when Catalina ceases to be the latest macOS version or when future Catalina versions no longer raise these issues._
 
-**Upgrading to macOS Catalina may cause normal `node-gyp` installations to fail.**
+**Both upgrading to macOS Catalina and running a Software Update in Catalina may cause normal `node-gyp` installations to fail.**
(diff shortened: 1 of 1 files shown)


karlhorky:
@cclauss Definitely, would be really nice to have a better workflow here!

It would be nice if the existing command-line tools installation would just keep working after software updates. Or at least having some command to fix the installation easily.

rvagg:
@gibfahn are you still with Apple? I don't suppose you could help us find someone there that can help us out here. We're still dealing with a steady stream of folks on Catalina that need to run through hoops just to get to a compile state. See dedicated doc this PR references (https://github.com/nodejs/node-gyp/blob/master/macOS_Catalina.md).

macOS_Catalina.md
```diff
@@ -1,8 +1,8 @@
-# Installation notes for macOS Catalina (v10.15)
+# Installation notes for macOS Catalina (v10.15) Updates
```
rvagg: I don't think "Updates" is needed here, it's not _just_ about updates, it's about upgrades as well, so if we remove this word it still applies to both cases

karlhorky: Sure, done. Happy to make changes given any guidance about wording, etc.

rvagg:
e15fad7c4

karlhorky:
Thanks @rvagg!

PR #2082, opened 2020-03-29
doc: add docs for Install the Visual Studio 2015 Build Tools

B4Lee:
##### Checklist

##### Description of change

adding aditional flags for install the Visual Studio 2015 Build tools instead of the Visual Studio 2017
ones. both install manualy or via CMD/Powershell

By default, `npm install --global --production windows-build-tools` will install 2017 build tools. To change that run that script with the `--vs2015` parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Install the current version of Python from the [Microsoft Store package](https:/
 
 #### Option 1
 
-Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
+Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools --vs2015` or `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
 
 #### Option 2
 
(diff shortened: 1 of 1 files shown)


richardlau:
> 
> 
> I'm strongly -1 on this. Node does not support building with VS2015 anymore, we should not encourage users to use it.

We don't support building Node.js with VS2015 but apparently do still support building addons with it: https://github.com/nodejs/build/blob/master/doc/windows-visualstudio-supported-versions.md#for-building-nodejs-addons

cc @joaocgreis

bzoz:
While we support VS2015, we should not suggest using it.

cclauss:
So, would it be a good idea keep this PR but change the `--vs2015` parameter to `--vs2019` (or `--vs2017`)?

bzoz:
TBH, I don't know if the windows-build-tools include vs2019. 

That said, the change right now is worded in a way that suggests that VS2015 should be installed besides VS2017 or build tools. That would have to be changed too.

B4Lee:
I've tried `npm install --global --production windows-build-tools` and it didn't work on my windows, then i uninstall. afterwards I added the --vs2015` parameter it worked

**link to related issue**
https://github.com/nodejs/node-gyp/issues/2025#issuecomment-599110253

> 
> 
> TBH, I don't know if the windows-build-tools include vs2019.
> 
> That said, the change right now is worded in a way that suggests that VS2015 should be installed besides VS2017 or build tools. That would have to be changed too.

B4Lee:
> 
> 
> We should not recommend outdated versions of dependencies. While VS2015 is still supported, it will only be while it is both supported upstream and easy for us to keep supporting it here. In fact, this README should be updated to recommend VS2019 - we did not do that before because we were waiting for a recent version of `node-gyp` to land on all supported Node.js versions through npm.
> 
> @B4Lee thanks for the effort in opening this PR. I've left a comment in the linked issue: [#2025 (comment)](https://github.com/nodejs/node-gyp/issues/2025#issuecomment-606321133) . Unfortunately, I don't see anything we can do about `node-sass` here without making the problem worse.

thanks for the explaining @joaocgreis 
I made this changes because there might be someone like me who experienced the same thing as me, which is an incompatible visual studio build tools version on windows

should I close this PR's? or make of some changes?

joaocgreis:
> should I close this PR's? or make of some changes?

I'll close, but feel free to propose other changes if you think something else should be done. Thanks!

PR #2084, opened 2020-03-30
dep: remove mkdirp dependency

rharriso:
#1071 ### Checklist

- [x] tests are included 

##### Description of change
`mkdirp` hasn't been updated in 4 years, and node's fs library
has had a recursive option for `fs` since v10.12 (oldest active LTS
version)

I was notified of an issue with `minmist` (a dependency of `mkdirp`) by `npm audit`. https://npmjs.com/advisories/1179. This library shouldn't be necessary as the old LTS supports recursive directory creation

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -1,10 +1,10 @@
 'use strict'
 
+const nodeFs = require('fs')
 const fs = require('graceful-fs')
 const path = require('path')
 const log = require('npmlog')
 const os = require('os')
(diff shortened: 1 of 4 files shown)


lib/configure.js
```diff
@@ -1,10 +1,10 @@
 'use strict'
 
+const nodeFs = require('fs')
```
rvagg: remove this and just use `fs`, it should pass through properly

PR #2105, opened 2020-04-24
Update acid test

vladimyr:
##### Checklist

##### Description of change

Joined two-step acid test into a single c/p-able shell snippet.


macOS_Catalina.md
```diff
@@ -17,14 +17,28 @@ If `ProductVersion` is less then `10.15` then this document is not for you. Norm
...
 To see if `Xcode Command Line Tools` is installed in a way that will work with `node-gyp`, run:
-1. `/usr/sbin/pkgutil --packages | grep CL`
-    * `com.apple.pkg.CLTools_Executables` should be listed. If it isn't, this test failed.
-2. `/usr/sbin/pkgutil --pkg-info com.apple.pkg.CLTools_Executables`
-    * `version: 11.0.0` (or later) should be listed. If it isn't, this test failed.
-    
-If both tests succeeded, _you are done_! You should be ready to install `node-gyp`.
+```
+/usr/sbin/pkgutil --pkgs |
+  grep -i com.apple.pkg.cltools |
```
cclauss, line 22: https://github.com/nodejs/node-gyp/blob/972780bde72b9a61153f3ea7ea9700cec001ffd7/gyp/pylib/gyp/xcode_emulation.py#L1436-L1438

vladimyr: Which would indicate that current docs are wrong too? 😕

vladimyr: @cclauss How about using this snippet:
```sh
#!/bin/bash

pkgs=(
  "com.apple.pkg.DeveloperToolsCLILeo" # standalone
  "com.apple.pkg.DeveloperToolsCLI"    # from XCode
  "com.apple.pkg.CLTools_Executables"  # Mavericks
)

for pkg in "${pkgs[@]}"; do
  output=$(/usr/sbin/pkgutil --pkg-info "$pkg" 2>/dev/null)
  if [ "$output" ]; then
    version=$(echo "$output" | grep 'version' | cut -d' ' -f2)
    break
  fi
done

if [ "$version" ]; then
  echo "Command Line Tools version: $version"
else
  echo >&2 'Command Line Tools not found'
fi
```

vladimyr:
I've converted the acid test to a remote shell script and introduced custom `git.io` URL mapping:
```
curl https://git.io/ -i -F "url=https://raw.githubusercontent.com/nodejs/node-gyp/master/acid-test.sh" -F "code=node-gyp:acid-test"
```
As a result, end-users can simply invoke it using:
```
curl -L https://git.io/node-gyp:acid-test
```

macOS_Catalina.md
```diff
@@ -17,14 +17,13 @@ If `ProductVersion` is less then `10.15` then this document is not for you. Norm
...
 ### The acid test
 To see if `Xcode Command Line Tools` is installed in a way that will work with `node-gyp`, run:
-1. `/usr/sbin/pkgutil --packages | grep CL`
-    * `com.apple.pkg.CLTools_Executables` should be listed. If it isn't, this test failed.
-2. `/usr/sbin/pkgutil --pkg-info com.apple.pkg.CLTools_Executables`
-    * `version: 11.0.0` (or later) should be listed. If it isn't, this test failed.
-    
-If both tests succeeded, _you are done_! You should be ready to install `node-gyp`.
+```
+curl -L https://git.io/node-gyp:acid-test | bash
```
rvagg, line 21: I'm not entirely comfortable promoting a shortened link, piping anything to bash involves a lot of trust so I'd rather this be explicit _and_ also present a more manual way of doing it. You could update your text to point directly to `acid-test.sh` if they want to download it themselves, and then use the full URL as an option if they're comfortable: `https://github.com/nodejs/node-gyp/blob/master/acid-test.sh`

_But_ also this test needs to be renamed, I know this makes it tedious, but this is only an issue with macOS Catalina and we need to separate this out as a concern. I suggest: `macOS_Catalina_acid_test.sh` so it shows up right next to this doc.

I'm not vouching for the accuracy here btw. That's another matter which needs signoff.

cclauss:
Homebrew now kicks out the following message:
```
Warning: A newer Command Line Tools release is available.
Update them from Software Update in System Preferences or run:
  softwareupdate --all --install --force

If that doesn't show you an update run:
  sudo rm -rf /Library/Developer/CommandLineTools
  sudo xcode-select --install

Alternatively, manually download them from:
  https://developer.apple.com/download/more/.
```
This worked like a champ for me.  I did the first command and about 30 seconds later a macOS __System Preferences/Software Update__ dialog box popped up asking me to perform and update.
```
com.apple.pkg.CLTools_Executables
com.apple.pkg.CLTools_SDK_macOS1015
com.apple.pkg.CLTools_SDK_macOS1014
com.apple.pkg.CLTools_macOS_SDK
```
Should we adopt this approach?

macOS_Catalina.md
vladimyr: > I'm not entirely comfortable promoting a shortened link, piping anything to bash involves a lot of trust so I'd rather this be explicit _and_ also present a more manual way of doing it. You could update your text to point directly to `acid-test.sh` if they want to download it themselves, and then use the full URL as an option if they're comfortable: `https://github.com/nodejs/node-gyp/blob/master/acid-test.sh`

Understood that and will expand it back to the full URL although it needs to be /raw instead of /blob.

> _But_ also this test needs to be renamed, I know this makes it tedious, but this is only an issue with macOS Catalina and we need to separate this out as a concern. I suggest: `macOS_Catalina_acid_test.sh` so it shows up right next to this doc.

Sure, it perfectly makes sense 👍

vladimyr:
> [Homebrew](https://brew.sh/) now kicks out the following message:
> ```
> Warning: A newer Command Line Tools release is available.
> Update them from Software Update in System Preferences or run:
>   softwareupdate --all --install --force
> 
> If that doesn't show you an update run:
>   sudo rm -rf /Library/Developer/CommandLineTools
>   sudo xcode-select --install
> 
> Alternatively, manually download them from:
>   https://developer.apple.com/download/more/.
> ```
> This worked like a champ for me.  I did the first command and about 30 seconds later a macOS __System Preferences/Software Update__ dialog box popped up asking me to perform an update.

@cclauss It happened to me too and I did the same except it totally corrupted my clt installation and the only way out was to remove it completely and manually install again by downloading from Apple's web (last resort described here). Simply put it isn't safe!

rvagg:
f7bfce96edd

vladimyr:
@rvagg Be warned that this isn't reliable at all. All of a sudden it gives me false negative on my own machine!

I essentially rewrote Python code that @cclauss pointed to in bash but that is yesterday news. Today's HB CLT detection looks much more sophisticated https://github.com/Homebrew/brew/blob/5fb158f/Library/Homebrew/os/mac/xcode.rb#L198-L326

Issue #1945, opened 2019-10-27
Remove Node.js 6 support 


gengjiawen:
Since Node.js is not support anymore. It's maybe good to remove related code and test.

rvagg:
PR would be good.

This would be good to do in 6, or 7 (not planned yet but might be something we do in time for Node 14), but we should probably be more careful with 5, for people who are stuck.

richardlau:
If it's in a semver-major I'd be okay dropping Node.js 8 support too.

codler:
Node js 8 is only supported untill 2019-12-31, so might be good to drop node 8 also

PR #2120, opened 2020-05-11
chore: remove support for EOL versions of Node.js

codebytere:
##### Description of change

Refs https://github.com/nodejs/node-gyp/issues/1945.

Should we also consider moving to GitHub Actions and adding specs for Node.js v14? Feel free to tell me if there's any config code specific to these EOL versions that i missed and i'll ditch it :)

##### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,14 +13,6 @@ jobs:
       env: NODE_GYP_FORCE_PYTHON=python2
       python: 2.7
 
-    - name: "Node.js 6 & Python 3.8 on Linux"
-      python: 3.8
-      env: NODE_GYP_FORCE_PYTHON=python3
-      before_install: nvm install 6
-    - name: "Node.js 8 & Python 3.8 on Linux"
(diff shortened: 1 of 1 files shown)


cclauss:
Nice initiative.  +1 on switching to GitHub Actions.  Why the failing tests on Py36?

codebytere:
@cclauss i was seeing a bunch of unexplained macOS GitHub Action runner errors earlier today on other repos too - just gave them another kick but i don't think it's related 🤔 the error showed up like this:

Screen Shot 2020-05-11 at 9 27 10 PM

which frankly is useless enough that it makes me think Actions itself was just struggling

cclauss:
* Why not test `node_js: 14` ?
* Why not test `dist: focal` ?

rvagg:
741ab09

+1 to Actions, and to adding 14 in there, and switching to Focal, although `ubuntu-latest` will likely get that for us anyway.

PR #2021, opened 2020-01-14
doc: fix changelog 6.1.0 release year to be 2020

quentinvernot:
##### Checklist

- [x] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change
Changelog said that v6.1.0 was released in 2019, fixed.

PR #2122, opened 2020-05-13
node --version

rvagg:
This commit was on `master`, I think probably accidentally as I don't believe there's a PR for it and there wasn't any metadata. I've pulled it out and force pushed a new `master` impacting the two commits following it, which I think should be fine.

This one looks necessary, but maybe the answer here is to remove the `Node.js 6 & Python 3.8 on Linux` instead of having a manual skip. I imagine this was a testing commit that accidentally made it onto `master` @cclauss?

Not a big deal but let's figure out what the final form is and get it landed ASAP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,9 @@ before_script:
   - npm list
 script:
   - node -e 'require("npmlog").level="verbose"; require("./lib/find-python")(null,()=>{})'
-  - npm test
+  - node --version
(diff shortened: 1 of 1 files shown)


rvagg:
(title of the PR was original commit summary, commit now has a new summary, btw)

rvagg:
made redundant by #2021 thankfully, will need to deal with the 5.x branch though but I might trim that to Node 8

PR #2123, opened 2020-05-13
deps: replace mkdirp with {recursive} mkdir

rvagg:
only supported on Node.js 10+, marking as semver-major

Closes: #2084


gengjiawen:
@rvagg This need rebase.

rvagg:
ooops, I thought I'd dealt with this already

rebased and found another usage that wasn't caught previously

PR #2125, opened 2020-05-13
v5.1.1 Proposal

rvagg:
The main change is @codebytere's Windows binary size fix, otherwise it's just test & doc.


rvagg:
released and merged

PR #2131, opened 2020-05-15
Don't use the "-2" flag with the py.exe launcher on Windows

### LINKED ISSUE #2130: Consider using the "py.exe" launcher to find Python 3 as well, not just for Python 2
* **Node Version**:  Applies to any versions of Node/NPM, but tested on Node 10.20.1 with npm 6.14.4
* **Platform**:  This is about Windows, in general. Tested on Windows 10.
* **Compiler**:  N/A; this is just about finding a Python binary. No compiler needed on the system to find a Python binary.
* **Module**:  Nothing, just did `findPython = require(./lib/find-python)` in a little standalone `.js` script to test Python detection.

<details><summary>My script ("fpy.js") (to be saved to/called from the node-gyp repo's main folder, like so: "node fpy.js".)</summary>

```js
findPython = require('./lib/find-python')

findPython(null, function (err, found) {
  if (err) {
    console.error(err)
  } else {
    python = found
    console.log('python was found: ' + python)
  }
})
```

</details>

```console
node.exe ./fpy.js
ERR! find Python
ERR! find Python Python is not set from command line or npm configuration
ERR! find Python Python is not set from environment variable PYTHON
ERR! find Python checking if "python3" can be used
ERR! find Python - "python3" is not in PATH or produced an error
ERR! find Python checking if "python" can be used
ERR! find Python - "python" is not in PATH or produced an error
ERR! find Python checking if "python2" can be used
ERR! find Python - "python2" is not in PATH or produced an error
ERR! find Python checking if Python is C:\Python37\python.exe
ERR! find Python - "C:\Python37\python.exe" could not be run
ERR! find Python checking if Python is C:\Python27\python.exe
ERR! find Python - "C:\Python27\python.exe" could not be run
ERR! find Python checking if the py launcher can be used to find Python 2
ERR! find Python - "py.exe" is not in PATH or produced an error
ERR! find Python
ERR! find Python **********************************************************
ERR! find Python You need to install the latest version of Python.
ERR! find Python Node-gyp should be able to find and use Python. If not,
ERR! find Python you can try one of the following options:
ERR! find Python - Use the switch --python="C:\Path\To\python.exe"
ERR! find Python   (accepted by both node-gyp and npm)
ERR! find Python - Set the environment variable PYTHON
ERR! find Python - Set the npm configuration variable python:
ERR! find Python   npm config set python "C:\Path\To\python.exe"
ERR! find Python For more information consult the documentation at:
ERR! find Python https://github.com/nodejs/node-gyp#installation
ERR! find Python **********************************************************
ERR! find Python
Error: Could not find any Python installation to use
    at PythonFinder.fail (C:\Users\[User]\node-gyp\lib\find-python.js:307:47)
    at PythonFinder.runChecks (C:\Users\[User]\node-gyp\lib\find-python.js:136:21)
    at PythonFinder.<anonymous> (C:\Users\[User]\node-gyp\lib\find-python.js:205:18)
    at PythonFinder.execFileCallback (C:\Users\[User]\node-gyp\lib\find-python.js:271:16)
    at ChildProcess.exithandler (child_process.js:301:5)
    at ChildProcess.emit (events.js:198:13)
    at maybeClose (internal/child_process.js:982:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:259:5)
```

</details>

---

#### Intro, Background, and the Gist of it

Hi folks, I'm wondering if the `node-gyp` team would be interested in using the `py.exe` launcher to find not only/exclusively Python 2, like it does now, but potentially Python 3 as well.

_(I just installed Python 3.8 from python.org, in the default install location, and `node-gyp` can't find it. `py.exe` can find it, though.)_

Apparently, judging by the commit history, using the Python launcher strictly to find Python 2 dates back just over four years ago, to _BEFORE_ `node-gyp` had any support for Python 3. From here: https://github.com/nodejs/node-gyp/pull/894 (The latest published version of `node-gyp` would have been `v3.3.1` at the time, judging by the release history: https://github.com/nodejs/node-gyp/releases?after=v3.5.0)

Thus, I assume the justification for only finding Python 2 at that time would have been the following: "`node-gyp` can't work with Python 3, so finding Python 3 would be counter-productive! We must only find Python 2..." (That's no longer true, and I think it might be nice to find Python 3 with such a convenient helper/launcher like `py.exe`.)

---

#### About the `py.exe` Launcher, and the installation location of Python on Windows

The `py.exe` launcher is quite useful for finding Python. It claims to be able to find the version downloadable from python.org AND the Python downloadable from the Windows Store. Source: https://docs.python.org/3/using/windows.html#launcher) It comes with Python installers for Windows since Python 3.3.

The Python situation on Windows is a little odd now, because it seems it used to be saved under something like `C:\Python[Maj[Min]\python.exe` (for example: `C:\Python37`), but recent installers save to somewhere more like this: `'C:\Users\[User]\AppData\Local\Programs\Python\Python38-32'` and in fact, the location can be customized during the install wizard. So it could end up anywhere, 
I guess.

I think the launcher might be the most reliable way to find Python on Windows these days. And in any case, it makes an excellent fallback. I only find that it is a shame that it isn't being used to find Python 3.

Indeed, going to Python.org, downloading Python 3.8.x and installing it will result in a Python installation that `node-gyp` can't find. I would like to approximate `node-gyp`'s Python finding algorithm (or require its `find-python.js` file directly), but not being able to find the default install location from python.org seems somehow not right. I didn't want to re-implement what seems like a (small) oversight in the repo I'm working on.

I _do_ think the Python from the Windows Store might end up on the path as `python` and/or `python3`, but I haven't tested or confirmed this.

---

#### Suggested Solution / Notes on an Implementation

Sorry I'm not very good with JS, or I'd whip up an implementation for a PR. But I think you could really just drop the `-2` [here](https://github.com/nodejs/node-gyp/blob/741ab096d554f2d3e6b6a7232b06872a0c1375f8/lib/find-python.js#L199), and update the comment above that, and that should do it.

(If that solution would be sufficient, or you'd like to see it run it through the CI tests and such, I would be _GLAD_ to whip that up as a PR.)

From a quick read of the [documentation page](https://docs.python.org/3/using/windows.html#from-the-command-line), the `py.exe` launcher seems to prefer Python 3 over Python 2, so just dropping the `-2` flag might not be appropriate for for the `node-gyp` 5.x branch (which, from what I can see, generally prefers Python 2 over Python 3.)

DeeDeeG:
##### Checklist

- [ ] tests are included 

##### Description of change

Fixes #2130

The [`py.exe` launcher for Windows](https://docs.python.org/3/using/windows.html#launcher) helps to locate Python on the system.

The `-2` or `-3` flags can specify a major version to find. Now that `node-gyp` supports Python 3, it makes sense to not restrict the `py.exe` launcher to only finding Python 2. It would be advantageous to let it find Python 3 as well.

The `py.exe` launcher prefers newer (higher version number) Python when multiple versions are found. (i.e. it prefers Python 3 over Python 2 if both are available).

### DIFF
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -119,7 +119,7 @@ PythonFinder.prototype = {
         checks.push({
           before: () => {
             this.addLog(
-              'checking if the py launcher can be used to find Python 2')
+              'checking if the py launcher can be used to find Python')
           },
           check: this.checkPyLauncher
         })
@@ -188,15 +188,10 @@ PythonFinder.prototype = {
   // Distributions of Python on Windows by default install with the "py.exe"
   // Python launcher which is more likely to exist than the Python executable
   // being in the $PATH.
-  // Because the Python launcher supports all versions of Python, we have to
-  // explicitly request a Python 2 version. This is done by supplying "-2" as
-  // the first command line argument. Since "py.exe -2" would be an invalid
-  // executable for "execFile", we have to use the launcher to figure out
-  // where the actual "python.exe" executable is located.
   checkPyLauncher: function checkPyLauncher (errorCallback) {
     this.log.verbose(
-      `- executing "${this.pyLauncher}" to get Python 2 executable path`)
-    this.run(this.pyLauncher, ['-2', ...this.argsExecutable], false,
+      `- executing "${this.pyLauncher}" to get Python executable path`)
+    this.run(this.pyLauncher, this.argsExecutable, false,
       function (err, execPath) {
       // Possible outcomes: same as checkCommand
         if (err) {
--- a/test/test-find-python.js
+++ b/test/test-find-python.js
@@ -146,14 +146,13 @@ test('find python - no python2, no python, unix', function (t) {
 })
 
 test('find python - no python, use python launcher', function (t) {
-  t.plan(4)
+  t.plan(3)
 
   var f = new TestPythonFinder(null, done)
   f.win = true
 
   f.execFile = function (program, args, opts, cb) {
     if (program === 'py.exe') {
-      t.notEqual(args.indexOf('-2'), -1)
       t.notEqual(args.indexOf('-c'), -1)
       return cb(null, 'Z:\\snake.exe')
     }


DeeDeeG:
I will update the commit message(s) to follow the commit guidelines, and probably squash them into one commit, unless maintainers prefer it broken up like this.

DeeDeeG:
Some additional notes:

- If a similar fix is desired for the `node-gyp` 5.x branch, where Python 2 is somewhat preferentially found before Python3, I would suggest running with the `-2` flag first, then without (or with the `-3` flag).
- Long-term, if `node-gyp` ever takes the position that Python 2 is to be actively banished, the `-3` flag would be useful for finding strictly Python 3.
- This was originally coded as a special case because the `-2` flag wasn't possible to pass neatly to the `execFile` function. (According to the inline comments of this file.)
  - There's a chance there might be some code de-duplication possible here.
- I'm open to fixing nits/small code style issues if there are any. Just point them out. JS isn't really one of my strong areas.

lib/find-python.js
```diff
@@ -188,15 +188,10 @@ PythonFinder.prototype = {
...
-  // explicitly request a Python 2 version. This is done by supplying "-2" as
-  // the first command line argument. Since "py.exe -2" would be an invalid
-  // executable for "execFile", we have to use the launcher to figure out
-  // where the actual "python.exe" executable is located.
   checkPyLauncher: function checkPyLauncher (errorCallback) {
     this.log.verbose(
-      `- executing "${this.pyLauncher}" to get Python 2 executable path`)
-    this.run(this.pyLauncher, ['-2', ...this.argsExecutable], false,
+      `- executing "${this.pyLauncher}" to get Python executable path`)
+    this.run(this.pyLauncher, [...this.argsExecutable], false,
```
DeeDeeG, line 194: I don't think `[...this.argsExecutable]` needs to be an array anymore. I'm not sure if the ... is needed, either.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

https://stackoverflow.com/questions/31048953/what-do-these-three-dots-in-react-do

cclauss: Small moves please.  Leave this PR as is and allow maintainers time to review it as is.

DeeDeeG: Tests pass with the following: 
```suggestion
    this.run(this.pyLauncher, this.argsExecutable, false,
```

DeeDeeG: Will wait for updates from reviewers, thanks.

rvagg: yeah, go with your suggestion, it's correct 👍

rvagg:
> The py.exe launcher prefers newer (higher version number) Python when multiple versions are found. (i.e. it prefers Python 3 over Python 2 if both are available).

If that's true then this change seems good to me! I don't think I have a good way to vet that claim but I imagine @cclauss has more of a clue than me about such things so 👍 .

DeeDeeG:
Hi, @rvagg. Regarding this:

>> The py.exe launcher prefers newer (higher version number) Python [. . . ]

I read that in the [documentation of the `py.exe` launcher](https://docs.python.org/3/using/windows.html#launcher). Here are the relevant parts:

> Unlike the PATH variable, the launcher will correctly select the most appropriate version of Python. It will prefer per-user installations over system-wide ones, and **orders by language version** rather than using the most recently installed version.

> You should find that the latest version of Python you have installed is started [. . .]

> If you have multiple versions of Python installed (e.g., 2.7 and 3.8) you will have noticed that Python 3.8 was started [. . .]

> If you want the latest version of Python 2.x you have installed, try the command:
> ```
> py -2
> ```


---

To verify this, I started with a Windows install with no Python on the system, and installed Python 2 and 3 with the `python-[version_here].exe` installers from python.org. I can confirm `py.exe` finds all of Python 3.8/3.7 and Python 2.7.

DeeDeeG:
By the way: The thing about "per-user installs being preferred over system-wide ones" sounded a little odd (what if Python 3 is installed system-wide, and Python 2 is installed just for the current user? Would it launch Python 2 by default?) I tried to reproduce this behavior as described in the documentation, but I couldn't reproduce it.

I tested with the following Pythons installed:

- Python 3.8 (system-wide install)
- Python 3.7 (per-user install)
- Python 2.7 (per-user install)

`py -0` (list available Pythons) gives the following output: 

```console
C:\Users\[User]>py -0
Installed Pythons found by py Launcher for Windows
 -3.8-32 *
 -3.7-32
 -2.7-32
```

The asterisk indicates the default Python that `py.exe` will run if no flags are passed to it. Indeed it does run Python 3.8:

```console
C:\Users\[User]>py
Python 3.8.3 (tags/v3.8.3:6f8c832, May 13 2020, 22:20:19) [MSC v.1925 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.executable
'C:\\Program Files (x86)\\Python38-32\\python.exe'
```

I can't reproduce the "per-user install takes precedence over a system-wide install" at all. And I can't think of an official way to install the same version of Python both system-wide and per-users, because the install wizard won't let you do that. (You can install the 64-bit and 32-bit editions of the same Python version, but the `py.exe` launcher seems to treat these as totally separate Pythons, and the 64-bit edition is preferred regardless of which is system-wide installed or per-user installed.)

DeeDeeG:
Force pushed with the code change discussed above.

 Also added this to the bottom of the first commit:

```
Fixes: https://github.com/nodejs/node-gyp/issues/2130
Refs: https://docs.python.org/3/using/windows.html#launcher
```

Happy to adjust or remove that if requested, just let me now.

cclauss:
The failing tests are unrelated to these changes. Please do not make further modifications.  Keep it simple and give maintainers time to review your work.  Further ongoning code changes and more paragraphs of text will only slow down the review process

rvagg:
all sounds good, and py.exe is an official Python thing too, correct? so we'd expect to see this installed by default on Windows systems with _some_ version of Python installed? Has it been shipping for long enough to likely be everywhere that Python is on Windows?

cclauss:
Yes.  `py.exe` is an official Python thing that has been around since 2011
https://www.python.org/dev/peps/pep-0397

DeeDeeG:
I have the ability to test things and report back / explain the nitty-gritty technical details, so please feel free to direct mention me (@DeeDeeG) if there is anything you need.

rvagg:
excellent work, landed in c255ffb, will get it into v7.0.0 this week

PR #2134, opened 2020-05-18
doc: give more attention to Catalina issues doc

mmarchini:
It's easy to miss the Catalina issues doc when reading the readme. Since
it can be a common issue among developers, it makes sense to give more
attention to it in the readme.

##### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,11 @@ Depending on your operating system, you will need to install:
 
 ### On macOS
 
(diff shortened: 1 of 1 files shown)


mmarchini:
Agreed, but even finding via Google is not so ergonomic unfortunately. Depending on the query used, the first few results Google gives you are from other unrelated repositories, and not from `node-gyp` (that's the result I got when searching the error on Google at least).

I also suggested adding a link to the guide on the website (https://github.com/nodejs/nodejs.org/pull/3180), so at least there's another way to find it (and _maybe_ it improves SEO for the guide? IDK how SEO works).

rvagg:
9aed628

Issue #439, opened 2014-05-06
node-gyp fails when on drive/folder with space in name


davidrinnan:
**this was originally a bug that I posted at node-canvas** 
https://github.com/LearnBoost/node-canvas/issues/407

I have two harddrives which means that osx gave them the names HD 1 and HD 2.

When I npm install canvas I get:

clang: error: no such file or directory: 'HD'
clang: error: no such file or directory: '2/davidrinnan/.node-gyp/0.10.26/src'
clang: error: no such file or directory: 'HD'
clang: error: no such file or directory: '2/davidrinnan/.node-gyp/0.10.26/deps/uv/include'
clang: error: no such file or directory: 'HD'
clang: error: no such file or directory: '2/davidrinnan/.node-gyp/0.10.26/deps/v8/include'

The drive used is called 'Macintosh HD 2' and we can see how it breaks down the harddrive name into individual paths rather than putting the name between ' '

The strange thing is that I have many modules installed but node canvas is the first that has caused issues. It apparently tries to do a node-gyp rebuild. I guess very few others do. 

So I made sure that node-gyp was updated to the latest version. But same issue anyways...

I then found out that npm uses its own node-gyp, regardless of the version that I have installed but I found the command to update npms internal node-gyp as well. Still same issue.

And as one can see in the error message I am running 0.10.26 of node. First I was running 10.20 but I had the same error message there as well...

a0viedo:
hey @davidrinnan, it's been some time now. is this still an issue?

bnoordhuis:
Right, this is still an issue but it's unlikely to get fixed because it's not just node-gyp that's sensitive to blanks in paths.

I'll go ahead and close this out as a wontfix.  The workaround is to use a path without blanks, like @working-name describes.

PR #2139, opened 2020-05-25
gyp: quote a few paths potentially containing spaces in makefile gene…

benob:
This is a tentative fix for https://github.com/nodejs/node-gyp/issues/439 (at least when using the makefile generator). It protects a few paths that may contain spaces in the generator. There are probably more instances of this problem, in that generator and other generators.

Unfortunately I am unable to write tests for assessing the fix and make a proper PR. I was nevertheless able to compile better-sqlite3 from a directory containing spaces.
 
##### Checklist

- [ ] tests are included


rvagg:
Sorry that we probably haven't made it very clear but this code is now being primarily maintained @ https://github.com/nodejs/gyp-next so could you please open a PR there and discuss with the team maintaining it? Thanks.

PR #2141, opened 2020-05-26
gyp: fix XCode CLT version detection on Catalina

szekelyisz:
Since Catalina XCode Command Line Tools don't show up among installed packages,
which breaks version detection in gyp/pylib/gyp/xcode_emulation.py.

The workaround is to remove the CLT installation directory, and run
`xcode-select --install` which reinstalls an older version of CLT. This older
version will eventually be upgraded by Software Update, which breaks gyp again.

##### Checklist

##### Description of change
This patch adds a fallback mechanism to detect newer CLT versions based on
software update history.

### DIFF
--- a/gyp/pylib/gyp/xcode_emulation.py
+++ b/gyp/pylib/gyp/xcode_emulation.py
@@ -1541,6 +1541,13 @@ def CLTVersion():
         except GypError:
             continue
 
+    regex = re.compile('Command Line Tools for Xcode\s+(?P<version>\S+)')
+    try:
+        output = GetStdout(['/usr/sbin/softwareupdate', '--history'])
+        return re.search(regex, output).groupdict()['version']
+    except:
+        return None
+
 
 def GetStdoutQuiet(cmdlist):
     """Returns the content of standard output returned by invoking |cmdlist|.
--- a/gyp/pylib/gyp/xcode_emulation.py
+++ b/gyp/pylib/gyp/xcode_emulation.py
@@ -1545,7 +1545,7 @@ def CLTVersion():
     try:
         output = GetStdout(['/usr/sbin/softwareupdate', '--history'])
         return re.search(regex, output).groupdict()['version']
-    except:
+    except (GypError, OSError):
         return None
 
 


gyp/pylib/gyp/xcode_emulation.py
```diff
@@ -1541,6 +1541,13 @@ def CLTVersion():
         except GypError:
             continue
 
+    regex = re.compile('Command Line Tools for Xcode\s+(?P<version>\S+)')
+    try:
+        output = GetStdout(['/usr/sbin/softwareupdate', '--history'])
+        return re.search(regex, output).groupdict()['version']
+    except:
```
cclauss, line 1548: Please, no bare exceptions https://realpython.com/the-most-diabolical-python-antipattern/

gyp/pylib/gyp/xcode_emulation.py
```diff
@@ -1541,6 +1541,13 @@ def CLTVersion():
         except GypError:
             continue
 
+    regex = re.compile('Command Line Tools for Xcode\s+(?P<version>\S+)')
+    try:
+        output = GetStdout(['/usr/sbin/softwareupdate', '--history'])
```
cclauss, line 1546: When tested on Travis CI, this will pass on `osx_image: xcode11.4` and `osx_image: xcode11.5` but will fail on all previous osx images listed at https://docs.travis-ci.com/user/reference/osx/#macos-version

gyp/pylib/gyp/xcode_emulation.py
szekelyisz: Obviously I'm not a Python wizard. :) Fixed.

gyp/pylib/gyp/xcode_emulation.py
szekelyisz: Fixed. The `except` block will now catch `OSError` if the command was not found and return `None`.

jkroepke:
After every CLT upgrade on 10.15, I have to reinstall Command Line Tools.

Please merge and release this, soon.

richardlau:
Since we vendor in gyp from https://github.com/nodejs/gyp-next this should really land there.

cclauss:
Can we delete the code from this repository that we vendor in from some other repository?

rvagg:
we could use git submodules but they're such a developer-hostile solution that it's usually better to just copy the code

I think we just need to improve our documentation and get more pro-active with pointing people over there when they open a PR here. The code's the same so moving the PR shouldn't be too difficult since you can just copy your edited files.

jkroepke:
In times of Github Actions, a bot would help here. If changes are in specific paths, just post a comment about the situation.

richardlau:
Unfortunately GitHub Actions running on pull requests from forks do not have permissions to post comments.

jkroepke:
Unless you are using a own personal access token.

PR #2156, opened 2020-06-16
docs: node-gyp v7 should solve the gyp: No Xcode or CLT version detec…

rvagg:
stray commits found on `master`, I assume that was a mistake and this should have been a PR @cclauss? Putting this through a PR instead.


rvagg:
added some additional text and organised the top of the doc a bit more to make it clearer (I think), pls review!

rvagg:
ee6fa7d

PR #2165, opened 2020-07-09
build: support apple silicon (arm64 darwin) builds

MarshallOfSound:
This provides initial support for setting `npm_config_arch=arm64` to build arm64 versions of native node modules.  In the future we should add support for universal native modules using `lipo` to link together `x64` and `arm64` versions of the `.node` file.  For now this at least gets things building.

Relies on https://github.com/nodejs/gyp-next/pull/49


rvagg:
Thanks @MarshallOfSound! This is a little more complicated these days because you'll need to split the commits over two repos. We now vendor in https://github.com/nodejs/gyp-next as our upstream version of GYP and don't make edits locally to it. The changes look simple enough and I imagine they'd be covered by the tests once we have arm64 darwin machines in our CI. I do wonder if you should be looking at more than just make.py. I think you might need to make the same changes in cmake.py and ninja.py because we have consumers of both of those.

Remove the gyp/ change from this commit and leave the other one here. Open a new PR against https://github.com/nodejs/gyp-next and when that's in we can vendor a new version here and land your other commit.

MarshallOfSound:
@rvagg sounds good, opened a PR to `gyp-next` here --> https://github.com/nodejs/gyp-next/pull/49

> I do wonder if you should be looking at more than just make.py. I think you might need to make the same changes in cmake.py and ninja.py because we have consumers of both of those.

Potentially, I don't have any test cases for those build systems though, this just fixes the native modules that get tested as part of Electrons CI.  If you have test cases on hand for cmake / ninja based modules I'm happy to use my DTK to fix those methods up as well in a follow up PR 😄

MarshallOfSound:
@rvagg What's the process for pulling in a new `gyp-next`?

targos:
@MarshallOfSound 

1. Download and extract `gyp-next`: https://github.com/nodejs/gyp-next/archive/v0.4.0.tar.gz
1. Replace `node-gyp/gyp` with the extracted directory

MarshallOfSound:
Thanks @targos 😄

rvagg:
neato!

merged in f461d56c5 & 3baa4e417, will try and wrangle a release soon

MarshallOfSound:
Hey @rvagg any chance we can get this released, apple-silicon is beginning to heat up, would be great to stop depending on git commit SHAs for things 😄

rvagg:
https://github.com/nodejs/node-gyp/pull/2192

PR #2167, opened 2020-07-15
build: add update-gyp script

MarshallOfSound:
Mostly because I hate doing things by hand 😆 

Usage:
```bash
./update-gyp.py v0.4.0
```

Automates the steps in https://github.com/nodejs/node-gyp/pull/2165#issuecomment-658411514

### DIFF
--- a/update-gyp.py
+++ b/update-gyp.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+import tarfile
+import tempfile
+import urllib.request
+
+BASE_URL = "https://github.com/nodejs/gyp-next/archive/"
+CHECKOUT_PATH = os.path.dirname(os.path.realpath(__file__))
+CHECKOUT_GYP_PATH = os.path.join(CHECKOUT_PATH, 'gyp')
+
+parser = argparse.ArgumentParser()
+parser.add_argument("tag", help="gyp tag to update to")
+args = parser.parse_args()
+
+tar_url = BASE_URL + args.tag + ".tar.gz"
+
+changed_files = subprocess.check_output(["git", "diff", "--name-only"]).strip()
+if len(changed_files) is not 0:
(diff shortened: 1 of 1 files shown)


update-gyp.py
```diff
@@ -0,0 +1,44 @@
...
+CHECKOUT_GYP_PATH = os.path.join(CHECKOUT_PATH, 'gyp')
+
+parser = argparse.ArgumentParser()
+parser.add_argument("tag", help="gyp tag to update to")
+args = parser.parse_args()
+
+tar_url = BASE_URL + args.tag + ".tar.gz"
+
+changed_files = subprocess.check_output(["git", "diff", "--name-only"]).strip()
+if len(changed_files) is not 0:
```
cclauss, line 23:
```suggestion
if changed_files:
```
Identity is not the same thing as equality in Python so __use ==/!= to compare constant literals (str, bytes, int, float, tuple)__ to avoid [SyntaxWarnings on Python >= 3.8](https://docs.python.org/3/whatsnew/3.8.html#porting-to-python-3-8).

% `python3.8`
```
>>> 0 is 0
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```

update-gyp.py
```diff
@@ -0,0 +1,44 @@
...
+  raise Exception("Can't update gyp while you have uncommitted changes in node-gyp")
+
+with tempfile.TemporaryDirectory() as tmp_dir:
+  tar_file = os.path.join(tmp_dir, 'gyp.tar.gz')
+  unzip_target = os.path.join(tmp_dir, 'gyp')
+  with open(tar_file, 'wb') as f:
+    print("Downloading gyp-next@" + args.tag + " into temporary directory...")
+    print("From: " + tar_url)
+    f.write(urllib.request.urlopen(tar_url).read())
```
cclauss, line 32:
```suggestion
    with urllib.request.urlopen(tar_url) as in_file:
        f.write(in_file.read())
```
To avoid unclosed handles.

rvagg:
aaf33c3

PR #2189, opened 2020-08-08
gyp: add missing extensions to compile_commands_json

manuel-arguelles:
##### Checklist

##### Description of change

Extends compile_commands_json generator to consider more file extensions
than just c and cc. This commit adds the cpp and cxx extensions to the
list of input source files.

### DIFF
--- a/gyp/pylib/gyp/generator/compile_commands_json.py
+++ b/gyp/pylib/gyp/generator/compile_commands_json.py
@@ -61,8 +61,9 @@ def AddCommandsForTarget(cwd, target, params, per_config_commands):
         defines = ["-D" + s for s in defines]
 
         # TODO(bnoordhuis) Handle generated source files.
+        extensions = (".c", ".cc", ".cpp", ".cxx")
         sources = target.get("sources", [])
-        sources = [s for s in sources if s.endswith(".c") or s.endswith(".cc")]
+        sources = [s for s in sources if s.endswith(extensions)]
 
         def resolve(filename):
             return os.path.abspath(os.path.join(cwd, filename))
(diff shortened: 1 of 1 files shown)


gyp/pylib/gyp/generator/compile_commands_json.py
```diff
@@ -61,8 +61,9 @@ def AddCommandsForTarget(cwd, target, params, per_config_commands):
         defines = ["-D" + s for s in defines]
 
         # TODO(bnoordhuis) Handle generated source files.
+        extensions = (".c", ".cc", ".cpp", ".cxx")
         sources = target.get("sources", [])
-        sources = [s for s in sources if s.endswith(".c") or s.endswith(".cc")]
+        sources = [s for s in sources if s.endswith(extensions)]
```
cclauss, line 66:
```suggestion
        sources = [s for s in target.get("sources", []) if s.endswith(extensions)]
```
And remove line 65 to streamline the logic.

gyp/pylib/gyp/generator/compile_commands_json.py
```diff
@@ -63,7 +63,7 @@ def AddCommandsForTarget(cwd, target, params, per_config_commands):
         # TODO(bnoordhuis) Handle generated source files.
         extensions = (".c", ".cc", ".cpp", ".cxx")
         sources = target.get("sources", [])
```
cclauss, line 65:
```suggestion
```
This line is no longer needed.

rvagg:
sorry to do this but this needs to move over to https://github.com/nodejs/gyp-next which we vendor in here, we no longer touch the code in gyp/ except for pulling in new versions from over there.

PR #2206, opened 2020-08-29
doc: drop the --production flag for installing windows-build-tools

DeeDeeG:
##### Checklist

##### Description of change

Drop `--production` from the install instructions for `windows-build-tools`. The `--production` flag isn't needed here, as "not installing devDependencies" is already implied when `npm install`ing a module by name.

(This was probably copy-pasted a while ago from `windows-build-tools`' `README.md`, which has since been updated to drop the `--production` flag from its install instructions: https://github.com/felixrieseberg/windows-build-tools/commit/4420dc848b74cdc70784ce00a1253d10465b1cb0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Install the current version of Python from the [Microsoft Store package](https:/
 
 #### Option 1
 
-Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
+Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
 
 #### Option 2
 


cclauss:
Why should we still recommend Windows Build Tools?  It’s current release installs legacy Python reached its end of life On 1/1/2020 (242 days ago).  The Microsoft Store install of Python is a much better/safer/easier way to go and it will auto update like when Python 3.9 is released next month.

DeeDeeG:
Hmm. As I see it, `windows-build-tools` is a one-liner on the command-line to get Visual Studio Build Tools. Much easier than tracking down the installer and selecting the right components. (It's surprisingly easy to not select the right ones/difficult to get it right).

It's also the only obvious way to get Visual Studio older than 2019 at the moment, that I'm aware of. Some projects hit corner cases and bugs with Visual Studio 2019 and need the older 2017 or 2015 that one can get with `windows-build-tools`.

As a Windows user, I'm much more interested in the C++ compiler toolchain (from Visual Studio). Python installation is coincidentally included, but if it could be disabled in `windows-build-tools` I personally would disable it. For what it's worth: If you already have Python 2.7 on your system, it usually skips installing another copy.

rvagg (approved):
This change lgtm, the question of whether that package is a good recommendation is another issue that we could deal with separately I think.

cclauss (approved):

richardlau (approved):

rvagg:
merged @ e98bb16

PR #2127, opened 2020-05-13
ci: switch to GitHub Actions

codebytere:
##### Description of change

This PR switches test builds from Travis to GitHub Actions.

cc @rvagg @cclauss 

##### Checklist

- [ ] `npm install && npm test` passes
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

PR #2210, opened 2020-08-31
ci: switch to GitHub Actions

### LINKED ISSUE #2209: Do we need Travis CI anymore or are GitHub Actions good enough?
https://travis-ci.com/github/nodejs/node-gyp vs. https://github.com/nodejs/node-gyp/actions

* **Node Version**: 
* **Platform**: 
* **Compiler**: 
* **Module**: 

```
Paste your log here, between the backticks. It can be:
  - npm --verbose output,
  - or contents of npm-debug.log,
  - or output of node-gyp rebuild --verbose.
Include the command you were trying to run.

This should look like this:

>npm --verbose
npm info it worked if it ends with ok
npm verb cli [
npm verb cli   'C:\\...\\node\\13.9.0\\x64\\node.exe',
npm verb cli   'C:\\...\\node\\13.9.0\\x64\\node_modules\\npm\\bin\\npm-cli.js',
npm verb cli   '--verbose'
npm verb cli ]
npm info using npm@6.13.7
npm info using node@v13.9.0

Usage: npm <command>
(...)
```

</details>

imatlopez:
##### Checklist

##### Description of change

This PR switches test builds from Travis to GitHub Actions.

closes #2127 closes #2209 

cc: @codebytere @rvagg @cclauss


imatlopez:
There's a serious bottleneck with macOS nodes being scarce and slow so I combined both workflows into one, but otherwise just had to revert @codebytere's change for msbuild to get all passing (probably something that should be tackled on its own)

package.json
```diff
@@ -45,6 +45,6 @@
   },
   "scripts": {
     "lint": "standard */*.js test/**/*.js",
-    "test": "npm run lint && tap --timeout=120 test/test-*"
```
rvagg, line 48: So, I'd really like this to stay for `npm t` because it's a good dev workflow. _If_ this is useful to separate out into CI maybe we need a separate `"ci"` target to use instead?

imatlopez: Added `test:ci` because personally I find that having testing steps separated helps find root causes for failures easier.

rvagg: and reverted? was it too complicated?

imatlopez: it broke windows tests because it added to variables to the environment. Rather go for a smaller changes than go changing things left and right.

rvagg (approved):
this seems good to me, @codebytere, @cclauss are you able to cast an eye over it?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
+# TODO: Line 20, enable python-version: 3.5
```
cclauss, line 1: Given that Py35 reaches end-of-life in less than two weeks, let's just remove support.
https://devguide.python.org/#status-of-python-branches

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+# TODO: Line 69, enable pytest --doctest-modules
```
cclauss, line 2: This seems quite important to me.  What are the remaining blockers?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+
+name: Tests
```
cclauss, line 4: I find the output of Actions less confusing when line 4 (currently Tests) and line 15 (currently test) match.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
cclauss, line 12:
```suggestion
```
I usually run tests on all pull requests on all branches.  Is there a reason to avoid doing that?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        node: [10.x, 12.x, 14.x]
+        python: [3.6, 3.7, 3.8]
```
cclauss, line 20: Please do just one test run on `3.9.0-rc.1` on your own to make sure that we are ready for the Py39 release in one month.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+        env:
+          PYTHON_VERSION: ${{ matrix.python }}
+      - name: Install Dependencies
+        run: |
+          npm install --no-progress
+          pip install flake8 pytest
+      - name: Lint python
```
cclauss, line 40:
```suggestion
      - name: Lint Python
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+      - name: Run node tests (Linux and macOS)
+        if: matrix.os != 'windows-latest'
+        run: |
+          npm test
+      - name: Run python tests (Linux and macOS)
```
cclauss, line 51:
```suggestion
      - name: Run Python tests (Linux and macOS)
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+        if: matrix.os != 'windows-latest'
+        run: |
+          python -m pytest
+      - name: Run node tests (Windows)
```
cclauss, line 55:
```suggestion
      - name: Run Node tests (Windows)
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,66 @@
...
+        if: matrix.os == 'windows-latest'
+        shell: bash
+        run: |
+          GYP_MSVS_VERSION=2015 GYP_MSVS_OVERRIDE_PATH="C:\\Dummy" npm test
+      - name: Run python tests (Windows)
```
cclauss, line 60:
```suggestion
      - name: Run Python tests (Windows)
```

cclauss (changes requested):
This is looking good.  Thanks for doing it.  What are the blockers to enable `pytest --doctest-modules` ?

.github/workflows/tests.yml
imatlopez: tested it in cc9942e and it passes (only the doctest-modules fails so I reverted it back to a TODO)

imatlopez:
Seems test-options.js is flaky on windows

rvagg:
That's more than flaky, it's broken, something to do with charmaps and non-ascii characters. Would @bzoz @joaocgreis or someone else with windows expertise mind having a look? Maybe there's an env var that we need to set for Actions to run these tests properly?

imatlopez:
@rvagg @cclauss added a patch to `test-options.js` which helps with the stubborn env flags on windows.

wrt the charmap test, the error you see is just uncaptured spew, it is part of the test, at least it is a skip criteria for the test

test/test-options.js
```diff
@@ -3,13 +3,19 @@
...
 
-test('options in environment', function (t) {
+test('options in environment', t => {
   t.plan(1)
 
   // `npm test` dumps a ton of npm_config_* variables in the environment.
   Object.keys(process.env)
-    .filter(function (key) { return /^npm_config_/.test(key) })
-    .forEach(function (key) { delete process.env[key] })
+    .filter(key => /^npm_config_/.test(key))
```
rvagg, line 11: the pattern in this project is always-parens for arrow functions, you have a few in this file that need changing

rvagg (approved):
this seems OK to me, is it ready to roll?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,51 @@
+# TODO: Line 47, enable pytest --doctest-modules
+
+name: Tests
+on: [push, pull_request]
+jobs:
+  Test:
```
cclauss, line 6:
```suggestion
  Tests:
```

cclauss (approved):
LGTM.  Nice work!

rvagg:
awesome, thanks @imatlopez!

landed in d1176d7, Shelly gets primary authorship on that squashed commit but I've added you in the metadata

PR #2212, opened 2020-09-01
Update README.md

enes4949:
Changed $npm install -g node-gyp to #npm install -g node-gyp because it gives an permission error.

##### Checklist

- [ ] tests are included 

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ etc.), regardless of what version of Node.js is actually installed on your syste
 You can install `node-gyp` using `npm`:
 
 ``` bash
-$ npm install -g node-gyp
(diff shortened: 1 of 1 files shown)


rvagg:
well.. it's fairly common to put a prompt character in there, I think most people know to remove the character. Personally I prefer to _not_ include anything that's not going to be copied and pasted and the prompt character usually gets in the way, especially for multi-line bash.

`#` is wrong, so -1 on this particular change, it just turns it into a comment and the syntax highlighting is just going to make it grey. But I'd be fine with removing the `$` completely but you'd need to do it for every instance of script copypasta in the README, and there's a lot of them.

PR #2213, opened 2020-09-02
Update README.md

enes4949:
Not running npm install -g node-gyp as root/admin gives permission error.

##### Checklist

- [ ] tests are included 

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,10 @@ etc.), regardless of what version of Node.js is actually installed on your syste
 
 ## Installation
 
-You can install `node-gyp` using `npm`:
+You can install `node-gyp` using `npm`(run as root/admin):
 
 ``` bash
-$ npm install -g node-gyp
+npm install -g node-gyp
 ```
 
 Depending on your operating system, you will need to install:
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Install the current version of Python from the [Microsoft Store package](https:/
 
 #### Option 1
 
-Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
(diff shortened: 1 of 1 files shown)


rvagg (changes requested):
You need a space before `(` and this isn't quite correct - many people (myself included) will set up their system so that their global node_modules is owned by their regular dev user account so `sudo` never needs to get involved. I'm not comfortable recommending root or `sudo` usage, although a suggestion about permissions and the need to overcome these problems might be welcome.

Something like:

> If your user account doesn't have access to the global `node_modules` (recommended) then this may need to be run as the root user or with `sudo`.

README.md
```diff
@@ -21,10 +21,10 @@ etc.), regardless of what version of Node.js is actually installed on your syste
 
 ## Installation
 
-You can install `node-gyp` using `npm`:
+You can install `node-gyp` using `npm`(run as root/admin):
 
 ``` bash
-$ npm install -g node-gyp
```
rvagg, line 27: also this change should be reverted or you're going to need to do it for all of the bash blocks in the README (which I'm not opposed to)

README.md
```diff
@@ -49,7 +49,7 @@ Install the current version of Python from the [Microsoft Store package](https:/
 
 #### Option 1
 
-Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global --production windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
+Install all the required tools and configurations using Microsoft's [windows-build-tools](https://github.com/felixrieseberg/windows-build-tools) using `npm install --global windows-build-tools` from an elevated PowerShell or CMD.exe (run as Administrator).
```
cclauss, line 52: Does windows-build-tools still install legacy Python?  If so then it should be removed in favor of installing Python 3 from the Microsoft App Store.

cclauss:
Closing.  Out of sync with the current README.md.

PR #2220, opened 2020-09-13
lib: migrate requests to fetch

### LINKED ISSUE #2047: request has gone into maintenance mode. Maybe replace it.

imatlopez:
##### Checklist

- [x] tests are included 

##### Description of change

`requests` is deprecated, this gets rid of it and modernizes parts of install.js

resolves #2047
closes #2198
closes #2222 

cc @rvagg


imatlopez:
@rvagg the tests are failing when downloading headers because the untar is not working with the streams here. Wonder if you might be able to find where?

lib/install.js
```diff
@@ -79,277 +84,214 @@ function install (fs, gyp, argv, callback) {
...
-          log.verbose('extracted file from tarball', path)
-          extractCount++
-        } else {
-          // invalid
-          log.silly('ignoring from tarball', path)
-        }
-        return isValid
+    } catch (err) {
+      if (err.code === 'EACCES') {
+        eaccesFallback(err)
```
rvagg, line 138: This is going to result in a double call to `cb()` because you've at once handed off flow control to the returning promise while also attempting to regain control here. `eaccesFallback` calls `cb` but you also have `then(cb, cb)` so 🤷. I'm not really sure how best to resovle this without pulling it apart even further, the middle-place with half Promises and half callbacks is very unpleasant. (See also https://github.com/nodejs/node-gyp/pull/2140 which perhaps you'd be interested in contributing to if you're interested in this challenge).

lib/install.js
```diff
@@ -79,277 +84,214 @@ function install (fs, gyp, argv, callback) {
...
-          var targetLibPath = path.resolve(dir, release.name + '.lib')
-          var libUrl = release[arch].libUrl
-          var libPath = release[arch].libPath
-          var name = arch + ' ' + release.name + '.lib'
-          log.verbose(name, 'dir', dir)
-          log.verbose(name, 'url', libUrl)
-
-          fs.mkdir(dir, { recursive: true }, function (err) {
-            if (err) {
-              return done(err)
```
rvagg, line 314: just noticing all of the potential double-callbacks for errors in this old code, yikes!

rvagg:
I can't see anything obvious that would make all windows and some macos to fail (the macos failure could suggest a flaky?). The windows version has to do a lot more parallel downloads in the `downloadNodeLibs()` so maybe there's something in there going awry. Maybe print some debug output in there that will show up in the Actions output - print what's going to get downloaded and what you eventually successfully get downloaded. Maybe there's a missing stream drain or missed callback in there somewhere that's not obvious.

oliverjessner:
lets merge this?

imatlopez:
@rvagg should this get a `semver-major` tag?

imatlopez:
@rvagg @cclauss since python2 is being dropped, which is a major change, is now the time to get this in?

rvagg:
Yeah, semver-major to be safe, too much can go wrong with this critical piece, although it's great to see make-fetch-happen being used so we're more likely to have parity with the same environments as npm as long as we're using it roughly the same. I don't have time to review this unfortunately, do we have another collaborator here that can look over and give it a :thumbsup:? I'm happy to ship this as a semver-major and maintain two majors for a little while (and probably ask npm not to hop on the new one too quickly).

gengjiawen (approved):
LGTM.

gengjiawen:
LGTM. I have tested on linux with some example from https://github.com/nodejs/node-addon-examples, works good too.

package.json
```diff
@@ -25,9 +25,9 @@
     "env-paths": "^2.2.0",
     "glob": "^7.1.4",
     "graceful-fs": "^4.2.3",
+    "make-fetch-happen": "^8.0.9",
```
indolent-developer, line 28: I guess the fetch requirement if pretty basic. Should we just use native fetch function. No need to again upgrade in couple of years :)

Trott: There is (at the time of this writing) no native fetch function in Node.js.

imatlopez: make-fetch-happen also deals with proxies which is a big thing for corporate users.

indolent-developer: ok proxy support would be critical requirement. No point implementing that.

Trott:
@nodejs/node-gyp Anyone else comfortable reviewing this?

gengjiawen:
> @nodejs/node-gyp Anyone else comfortable reviewing this?

@rvagg @cclauss

rvagg:
Yeah, from my brief look a few months ago this seemed fine, I'm mainly concerned about the huge potential for breakage but using the npm http client does alleviate concerns and we can publish this as a semver-major. I really don't have the bandwidth to give this a proper review, but I give it a very general :thumbsup: (maybe that's a conceptual :thumbsup:). @gengjiawen if you're happy with the code then that might be enough.

gengjiawen:
Thanks for the great contribution 👍

PR #2222, opened 2020-09-15
feat: promisify install

imatlopez:
##### Checklist

- [x] tests are included 

##### Description of change

An effort to modernize the codebase, handover control of installation
to promises.

cc: @rvagg

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -4,6 +4,8 @@ const fs = require('graceful-fs')
 const os = require('os')
 const tar = require('tar')
 const path = require('path')
+const util = require('util')
+const stream = require('stream')
 const crypto = require('crypto')
 const log = require('npmlog')
 const semver = require('semver')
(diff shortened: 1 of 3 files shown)


rvagg:
@imatlopez would you mind trying to rebase your branches rather than merging master back in? when they land they'll need to be merge-commit-free which will end up requiring awkward cherry-picking and conflict resolving. If you could do that early then it might save work later on.

I'm also thinking about how to bite all of this off. I'm keen to get some of this work in but it's (a) a lot of work to review carefully, so takes time, and (b) high-risk, as with any major change here since our test suite is so sparse but the user-base so massive. So I hope you understand the delays here, I just wish I had a clear strategy for it. Maybe we should try something like merging into a `next` branch and doing pre-releases or something. It's difficult to get real use of versions that aren't bundled with npm, but perhaps we could try at least.

There's also #2140 which is in a similar limbo and has similar change scope.

Issue #2203, opened 2020-08-28
Could not find any Visual Studio installation to use after trying all the tips I could find online


kakyoism:
* **Node Version**: `node -v`=v12.18.3 and `npm -v`=6.14.8
* **Platform**: `systeminfo | findstr /B /C:"OS Name" /C:"OS Version" /C:"System Type"` (Windows)
```
OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.18362 N/A Build 18362
System Type:               x64-based PC
```

* **Compiler**: `msbuild /version & cl` (Windows)

```
Microsoft (R) Build Engine version 15.9.21+g9802d43bc3 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

15.9.21.664Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27043 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]
```
* **Module**: zeromq

<details><summary>Verbose output (from npm or node-gyp):</summary>

```
me@me-PC0 D:\Desktop\myapp
# node-gyp configure --msvs_version=2017
gyp info it worked if it ends with ok
gyp info using node-gyp@7.1.0
gyp info using node@12.18.3 | win32 | x64
gyp info find Python using Python version 3.8.5 found at "C:\Python\Python38\python.exe"
gyp ERR! find VS
gyp ERR! find VS msvs_version was set from command line or npm config
gyp ERR! find VS - looking for Visual Studio version 2017
gyp ERR! find VS running in VS Command Prompt, installation path is:
gyp ERR! find VS "C:\Program Files (x86)\Microsoft Visual Studio\2017"
gyp ERR! find VS - will only use this version
gyp ERR! find VS could not use PowerShell to find Visual Studio 2017 or newer
gyp ERR! find VS looking for Visual Studio 2015
gyp ERR! find VS - found in "C:\Program Files (x86)\Microsoft Visual Studio 14.0"
gyp ERR! find VS - could not find MSBuild in registry for this version
gyp ERR! find VS not looking for VS2013 as it is only supported up to Node.js 8
gyp ERR! find VS msvs_version does not match this VS Command Prompt or the
gyp ERR! find VS installation cannot be used.
gyp ERR! find VS
gyp ERR! find VS **************************************************************
gyp ERR! find VS You need to install the latest version of Visual Studio
[...]

PR #2229, opened 2020-09-29
lib: better log message when ps fails

initram:
##### Checklist

##### Description of change
I have changed the log message, that is printed when PowerShell fails to run Find-VisualStudio.cs, to include a tip to run the command again with --loglevel silly. This is do guide people to solve their problem them selves instead of spending hours googling the printed error messages (like I and others have done before). See fx. #2203. 

Many people having issues with node-gyp are not familiar with node log-levels, and do not see that as a path to solving their problems.


lib/find-visualstudio.js
```diff
@@ -151,7 +151,7 @@ VisualStudioFinder.prototype = {
 
     const failPowershell = () => {
       this.addLog(
-        'could not use PowerShell to find Visual Studio 2017 or newer')
+        'could not use PowerShell to find Visual Studio 2017 or newer, run command with \'--loglevel silly\' to figure out why')
```
rvagg, line 154: It's a bit too definite, you may not figure out why. Maybe:

> "try re-running with `--loglevel=silly` for more details"

rvagg:
This seems reasonable to me, any input from Windows contributors?

bzoz (approved):

rvagg:
landed in 3e7f8ccafc

PR #2261, opened 2020-11-17
workflows: replace depracted set-env with github envs

### LINKED ISSUE #2231: GitHub Actions: The `set-env` command is deprecated and will be disabled soon.
The `set-env` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

As seen at https://github.com/nodejs/node-gyp/actions

* **Node Version**: 
* **Platform**: 
* **Compiler**: 
* **Module**: 

```
Paste your log here, between the backticks. It can be:
  - npm --verbose output,
  - or contents of npm-debug.log,
  - or output of node-gyp rebuild --verbose.
Include the command you were trying to run.

This should look like this:

>npm --verbose
npm info it worked if it ends with ok
npm verb cli [
npm verb cli   'C:\\...\\node\\13.9.0\\x64\\node.exe',
npm verb cli   'C:\\...\\node\\13.9.0\\x64\\node_modules\\npm\\bin\\npm-cli.js',
npm verb cli   '--verbose'
npm verb cli ]
npm info using npm@6.13.7
npm info using node@v13.9.0

Usage: npm <command>
(...)
```

</details>

MadMango:
Fixes #2231 

##### Checklist

- [ ] tests are included 

##### Description of change

set-env is deprecated and builds will fail, this should fix it

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,8 +32,8 @@ jobs:
       - name: Set Windows environment
         if: matrix.os == 'windows-latest'
         run: |
-          echo 'GYP_MSVS_VERSION=2015' >> $GITHUB_ENV
-          echo 'GYP_MSVS_OVERRIDE_PATH=C:\\Dummy' >> $GITHUB_ENV
+          Set-Variable -Name "GYP_MSVS_VERSION" -Value "2015"
+          Set-Variable -Name "GYP_MSVS_OVERRIDE_PATH" -Value "C:\\Dummy"
       - name: Lint Python
         if: matrix.os == 'ubuntu-latest'
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,7 @@ jobs:
           pip install flake8 pytest
       - name: Set Windows environment
         if: matrix.os == 'windows-latest'
-        run:
+        run: |
           echo 'GYP_MSVS_VERSION=2015' >> $GITHUB_ENV
           echo 'GYP_MSVS_OVERRIDE_PATH=C:\\Dummy' >> $GITHUB_ENV
       - name: Lint Python
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,6 +32,7 @@ jobs:
       - name: Set Windows environment
         if: matrix.os == 'windows-latest'
         run: |
+          import os
           os.environ["GYP_MSVS_VERSION"] = "2015"
           os.environ["GYP_MSVS_OVERRIDE_PATH"] = "C:\\Dummy"
         shell: python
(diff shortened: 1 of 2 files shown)


MadMango:
Just realised this is for powersheel or cmd and my unix fix won't do here, I'll try come up with something else but can someone confirm if this will run in CMD or powershell?

.github/workflows/tests.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       - name: Set Windows environment
         if: matrix.os == 'windows-latest'
         run:
```
richardlau, line 34: Looks like an existing bug in the workflow, but this should be:

```suggestion
        run: |
```
to preserve the line break between the two subsequent `echo ...` lines.

richardlau:
> Just realised this is for powersheel or cmd and my unix fix won't do here, I'll try come up with something else but can someone confirm if this will run in CMD or powershell?

Since we don't override the default I believe actions on Windows run in powershell by default: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

MadMango:
Tried to help but I'm out of ideas I'm afraid, setting these variables through powershell or python doesn't seem to make a difference:

image
image

In both cases when trying to print these variables they are not defined:

image

Any other ideas?

gyp/pylib/gyp/MSVSVersion.py
```diff
@@ -549,6 +549,9 @@ def SelectVisualStudioVersion(version="auto", allow_fallback=True):
         "2017": ("15.0",),
         "2019": ("16.0",),
     }
+    print('ENV OVERRIDES: \n')
+    print(os.environ.get("GYP_MSVS_OVERRIDE_PATH"))
+    print(os.environ.get("GYP_MSVS_VERSION"))
```
cclauss, line 554:
```suggestion
    for key in ("GYP_MSVS_OVERRIDE_PATH", "GYP_MSVS_VERSION"):
        print("{}: {}".format(key, os.environ.get(key)))
```

gengjiawen:
I think I find the right way in https://github.com/nodejs/node-gyp/pull/2285, can you cherrypick it here ?

MadMango:
> I think I find the right way in #2285, can you cherrypick it here ?

Well you got it working in that one, let's just close this one and go with yours :) 
I would've never quessed the `$Env:GITHUB_ENV` syntax though 🤯

gengjiawen:
Ok， I find the other part is still useful, I will try to import to my patch.

Thanks for taking time on this :)

> I would've never quessed the $Env:GITHUB_ENV syntax though 🤯

Yeap, github documentation is useless in this case. I searched lots of code find this out.

PR #2306, opened 2021-01-19
gyp: update gyp.el to change case to cl-case

liuchong:
##### Checklist
- [x] commit message follows [commit guidelines]

##### Description of change
‘case’ is an obsolete alias (as of 27.1); use ‘cl-case’ instead.


rvagg:
this file is managed by https://github.com/nodejs/gyp-next and vendored in here without modification so you'd best file this issue over there (although I'm not sure that file is actively used by _anything_)

Issue #893, opened 2016-03-13
default values for environment variables


rimmartin:
Hi,

Is there a way to default a variable in binding.gyp yet allow a different user externally set a different value.

https://github.com/HDF-NI/hdf5.node/blob/master/binding.gyp#L3
Some people have a hard coded path to native dependencies while others like a more app packaged deployment and/or the ability to have more than one version installed

```
"hdf5_home_linux": "/usr/local"
```
How to keep a default yet make it adaptable for other choices?

bnoordhuis:
In theory, you can pass custom defines to node-gyp like this:
```
$ node-gyp -Dhdf5_home_linux=/foo/bar rebuild
```
You need to declare the variable as configurable in your binding.gyp, like this:
```
'variables': { 'hdf5_home_linux%': '/usr/local' }
```
Having said that, I had a look at the code that implements argument parsing and it looks broken to me.  I'll see if I can come up with a PR.

Also, how to get npm to pass the `-D` option to node-gyp is something I'll have to leave as an exercise to the reader because I don't know. :-)

rimmartin:
Thank you for looking into this.  I'll test further

rimmartin:
I found that this does work to make npm (node-v5.0.0) pick up the variable as a switch.
```
npm install  --hdf5_home_linux=<pathtoyourhdf5install>
```
for example where this was a dependent project that used my module

bnoordhuis:
For anyone coming here through a search engine, you can alternatively pass options through the GYP_DEFINES environment variable:
```
$ env GYP_DEFINES="-Dname=value" npm install
```
That should work with any node-gyp version.

shrinktofit:
> For anyone coming here through a search engine

The `GYP_DEFINES` is enabled by command line option "--ignore-environment", which by default is `true`.

m4heshd:
> `$ env GYP_DEFINES="name=value name2=value" npm install`

Anybody know where's the documentation for this? 🙄

PR #2319, opened 2021-01-27
configure: app_root_dir for the application root directory

tveronezi:
##### Checklist

##### Description of change

I have a `binding.gyp` looks like (redacted):

```
{
  "targets": [
    {
      "target_name": "my_target",
      "sources": [
        "src/a.cc",
        "src/b.cc",
        "src/c.cc",
        "src/d.cc"
      ],
...
      "conditions": [
        ['OS == "linux"',
          {
            'variables': {
              'USE_MUSL': '<!(ldd --version 2>&1 | head -n1 | grep "musl" | wc -l)'
            },
            'conditions': [
              ['<(USE_MUSL) == 1',
                {
                  "libraries": [
                    "<(module_root_dir)/dir_a/dir_b/my-musl.so",
                  ]
                },
              # else
                {
                  "libraries": [
                    "<(module_root_dir)/dir_a/dir_b/my.so",
                  ]
                }
              ]
            ]
          }
        ],
...
      ],
...
}
``` 

The problem is that in a particular installation I have (azure app service), `module_root_dir` is not where I can find the `dir_a/dir_b/*.so` files. 

This PR creates a new variable `app_root_dir` that can be overwritten with an environment variable (`APP_ROOT_DIR`).  This way, no matter what Azure does to install a nodejs application, I can always point to the right directory. 

I know I could use `"<!(echo $MY_AWESOME_VARIABLE)/dir_a/dir_b/my.so"` but it is not cross-platform.

Is this OK or is there an existing solution I can use?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ test/.node-gyp
 .ncu
 .nyc_output
 package-lock.json
+.idea
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -317,6 +317,11 @@ function configure (gyp, argv, callback) {
       }
       argv.push('-Dnode_lib_file=' + nodeLibFile)
       argv.push('-Dmodule_root_dir=' + process.cwd())
+      argv.push('-Dapp_root_dir=' + (
+        process.env.APP_ROOT_DIR && process.env.APP_ROOT_DIR.trim() !== ''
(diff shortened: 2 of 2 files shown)


tveronezi:
I guess I've just found a solution. :) 
https://github.com/nodejs/node-gyp/issues/893#issuecomment-215406837

Issue #2286, opened 2020-12-18
consider release-please to improve release process


gengjiawen:
https://github.com/google-github-actions/release-please-action It's has been using in `gyp-next` and works pretty great so far as I observed (https://github.com/nodejs/gyp-next/blob/master/.github/workflows/release-please.yml
). 

Maybe it's also good here too.

@rvagg What do you think ? 

cc @targos @richardlau

rvagg:
Maybe, I'm not totally opposed to it. What we're doing now doesn't scale very well because it's basically coming down to me managing the releases.

_But_ the install base of node-gyp is very large. 6M weekly downloads. And it's in critical path in CI systems for a very large number of installs. So we have security and stability concerns that should give us pause. And this is all outside of npm where they at least have some stability gating when pulling node-gyp in. So my concern would be the ease with which releases can be made and pushed into people's production pipelines. The write-access on this repo is fairly broad right now because we can gate releases with npm write access.

targos:
GitHub Actions now have environment protection rules and environment secrets: https://github.blog/changelog/2020-12-15-github-actions-environments-environment-protection-rules-and-environment-secrets-beta/
Maybe that can be used to protect an npm token and only allow a subset of collaborators to publish the package?

gengjiawen:
> The write-access on this repo is fairly broad right now because we can gate releases with npm write access.

Maybe only use it for changelog generate and release PR. remove npm publish step from the workflow.

rvagg:
> Maybe only use it for changelog generate and release PR. remove npm publish step from the workflow.

yes, I was thinking that this might be a good way to get started, gate publish on a human for now and see how we go.

I'd be ok with exploring this switch, it'll help with getting things merged and maintaining a changelog, but we're all going to have to be educated in the commit message pattern (I'm not _quite_ there yet but have been trying to adopt it in other projects). We're clearly not scaling by sticking with the clean merge patterns and metadata of nodejs/node.

DeeDeeG:
@rvagg here is a similar thing to `release-please-action`... but producing the Node style of CHANGELOG formatting, and working well with Node-style commit prefixes (`lib:`, `doc:`, `gyp:`, and so on) rather than conventional commits prefixes (`chore:`, `fix:`, `feat:`): [workflow.yml](https://github.com/DeeDeeG/node-gyp/blob/9fbccb181d732b5be09c1ef120642548c0288a05/.github/workflows/new-workflow.yml)

[...]

PR #2322, opened 2021-01-30
lib: avoid changing process.config

targos:
##### Checklist

##### Description of change

It is deprecated and will emit a warning in Node.js 16.

Refs: https://github.com/nodejs/node/pull/36902

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -96,7 +96,7 @@ function configure (gyp, argv, callback) {
 
     log.verbose('build/' + configFilename, 'creating config file')
 
-    var config = process.config || {}
+    var config = Object.assign({}, process.config)
     var defaults = config.target_defaults
     var variables = config.variables
 


richardlau (approved):

rvagg (approved):

targos:
I'll let someone from @nodejs/node-gyp land this.

rvagg:
@gengjiawen I don't suppose you want to help us get set up with the Actions suggested in #2286 so we can reduce this friction?

gengjiawen:
> @gengjiawen I don't suppose you want to help us get set up with the Actions suggested in #2286 so we can reduce this friction?

Actually I can help with it. I am home for Chinese spring festival, get more free time :)

gengjiawen (approved):

DeeDeeG:
Hi, folks.

I was testing `node-gyp` with NodeJS nightly v16. I noticed `node-gyp` still changes `process.config`, and the warning is still emitted on Node 16. I am working on submitting a fix as a PR.

I believe this is because `Object.assign()` does not deep-copy nested objects under the object being copied. It only creates a reference to the nested object. So we can't modify the new object's nested objects without modifying `process.config`.

<details><summary>Details of the problem, with annotated code snippet to illustrate (click to expand)</summary>

We do the following, which mutates the `process.config.target_defaults` object nested under `process.config`:

```js
// lib/configure.js

    var config = Object.assign({}, process.config) // This is not a deep-copy operation!
    var defaults = config.target_defaults // defaults is a pointer to process.config.target_defaults!

    // ... snipped ...

    // don't inherit the "defaults" from node's `process.config` object.
    defaults.cflags = [] // this mutates process.config.target_defaults.cflags !
    defaults.defines = []
    defaults.include_dirs = []
    defaults.libraries = []
```

</details>

I think it's a bit suboptimal to stringify to JSON and parse it back out again, but I am not aware of any other simple way to deep copy an object. `deepCopyOfObject = JSON.parse(JSON.stringify(originalObject))`.

There are libraries specifically for deep copying objects, but I'm not sure it's worth adding another library just to deep copy `process.config`.

rvagg:
I wouldn't be opposed to `JSON.parse(JSON.stringify())`, it's simple and easy and avoids another dep. `process.config` is pretty small and simple so it shouldn't cost much. It also wouldn't be hard to write a manual deep copy but JSON might just be easier. Or maybe we can work around it entirely by being a bit smarter about what we pull out?

DeeDeeG:
Exploring options: we could also do this:

```js
var defaults = Object.assign({}, process.config.target_defaults)
var variables = Object.assign({}, process.config.variables)
var config = { variables, target_defaults: defaults }
```

(Since `process.config` is only two layers deep and has not many children, we can directly copy both child objects out.)

I personally find it slightly more confusing to read than the `JSON.stringify()` approach.

And we have to be confident neither of `process.config.target_defaults` nor `process.config.variables` will contain a deeper-nested object in the future, or we could end up mutating `process.config` again.

<details><summary>microbenchmarks (either fix is about 0.1 - 0.4 milliseconds slower than current master branch) (click to expand)</summary>

Original as of this PR: ~0.10 - 0.13 milliseconds (outlier slow runs: ~0.14 - 0.15 milliseconds) _**(NOTE: deprecated)**_
Object.assign() fix: ~0.25 - 0.31 milliseconds (outlier slow runs: ~ 0.35 - 0.52 milliseconds)
JSON.stringify() fix: ~0.26 - 0.35 milliseconds (outlier slow runs: ~ 0.46 - 0.56 milliseconds)

These times are from a microbenchmark of just assigning `defaults`, `variables` and `config` with each approach, and doing `perf_hooks.performance.now()` immediately before and after object assignment. it's not reliably or perceptibly faster than the `JSON.stringify()` approach. They're both "twice as slow" as doing `Object.assign()` once, with references instead of deep copying. So yeah, there's a "performance hit"... a performance hit of only about 0.1 to 0.2 milliseconds! (Up to 0.4 milliseconds slower in a few outlier runs.)

</details>

I personally lean toward `JSON.parse(JSON.stringify)` over that. But in general I think these are close to, if not _the_ two optimal solutions for deep copying today's `process.config`.

targos:
Note that when I opened this, I knew that the could would still change deep parts of `process.config`. I only cared about avoiding the warning and I validated it by changing the code directly inside npm/node_modules so I'm surprised that it can still be triggered.

DeeDeeG:
@targos I have some steps to reproduce. I think these are pretty real-world equivalent:

- Use [`n`](https://github.com/tj/n) to install Node 16 nightly. (command: `n nightly`). macOS 10.15.7 Catalina.
  - (Effectively: download and extract the [Node 16 nightly tarball](https://nodejs.org/download/nightly/v16.0.0-nightly2021041040ace47396/), ensure its extracted `bins` dir is on the `PATH`)
- Clone some package repo that builds with `node-gyp` (I used the `superstring` package: `git clone https://github.com/atom/superstring`), `npm install` its dependencies as normal.
- In the module directory `(/Users/[me]/superstring`), do `npx node-gyp configure`.

To get the full trace stack for the warning, I did:
- `npm install -g node-gyp`
- `node --trace-deprecation $(command -v node-gyp) configure` inside a module directory that build with `node-gyp` (again, `superstring` in my case.)

targos:
Thanks for the detailed explanation. I just realized that the deprecation was applied deeply using proxies, so yeah we do need to create a deep copy...
I think the JSON stringify/parse approach is fine in this case.

PR #2324, opened 2021-02-07
fix build.js remove comments issue

diegozhu:
##### Checklist
- [x ] `npm install && npm test` passes
- [] tests are included 
- [] documentation is changed or added

##### Description of change

some user may use git Windows-style line-ending which will result in JSON.parse error.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -46,7 +46,7 @@ function build (gyp, argv, callback) {
         }
         return
       }
-      config = JSON.parse(data.replace(/#.+\n/, ''))
+      config = JSON.parse(data.replace(/#.+[\n\r]+/, ''))
 
       // get the 'arch', 'buildType', and 'nodeDir' vars from the config
       buildType = config.target_defaults.default_configuration


lib/build.js
```diff
@@ -46,7 +46,7 @@ function build (gyp, argv, callback) {
         }
         return
       }
-      config = JSON.parse(data.replace(/#.+\n/, ''))
+      config = JSON.parse(data.replace(/#.+[\n\r]+/, ''))
```
rvagg, line 49: the `[]` is making this do more than is intended, let's be minimal and just turn `\n` into `\r?\n` and that should solve the Windows problem

rvagg (changes requested):

lib/build.js
lukekarrys:
```suggestion
      config = JSON.parse(data.replace(/#.+\r?\n/, ''))
```

lukekarrys:
I committed the suggested change and rebased against main to get CI running and tests are now failing.

PR #2333, opened 2021-02-15
lib: drop Python 2 support in find-python.js

DeeDeeG:
##### Checklist

- [x] tests are included _(I updated existing tests)_ 

##### Description of change

Follow-up to https://github.com/nodejs/node-gyp/pull/2300 and https://github.com/nodejs/node-gyp/pull/2318.

(Now that `gyp-next` no-longer supports Python 2, `node-gyp` should stop finding and trying to use Python 2 to run `gyp-next` with.)

This Pull Request:
- Updates the accepted versions of Python in `lib/find-python.js` from `'2.7.x || >=3.5.0'` to just `'>=3.6.0'`
- Removes checks attempting to find Python 2 (on all platforms)
- Updates the fallback paths (default Python install paths) to check on Windows
- Uses the `-3` flag with the `py.exe` launcher (on Windows), so that it looks for/executes only a Python 3 executable
- Updates the tests to reflect the above changes


DeeDeeG:
Notes for reviewers:

- The posture of `gyp-next` and `node-gyp` overall is to support Python `>= 3.6`, not `>= 3.5`. I left it at Python `>= 3.5` to avoid making any changes beyond dropping Python 2.
  - Please let me know if this should be changed to `>= 3.6` in this PR, or left as `>= 3.5`, to be addressed in another PR.
  - Prior discussion elsewhere in the NodeJS ecosystem:
    - https://github.com/nodejs/node/pull/36691#issuecomment-752651399
    - https://github.com/nodejs/node/pull/36691#issuecomment-768994805
    - https://github.com/nodejs/node-gyp/pull/2300#issuecomment-759142662

---

- Adding the `-3` flag for the `py.exe` launcher is mostly redundant to the "main change," the updated `>= 3.5.0` semver range of accepted Python versions. If the user has Python 3 available, the `py.exe` launcher (even without flags) will already run Python 3. But I consider The `-3` flag beneficial for performance in the "no Python 3 available" case.
  - Performance benefit: When the user has Python 2 but not Python 3, the `-3` flag skips having the launcher pointlessly execute Python 2 just to have the semver check reject it.
  - Hypothetical drawback: If the `py.exe` launcher ever supports "Python 4", and at such a time as `node-gyp` supports "Python 4", this will need to be updated.

lib/find-python.js
```diff
@@ -18,16 +18,15 @@ PythonFinder.prototype = {
   log: logWithPrefix(log, 'find Python'),
   argsExecutable: ['-c', 'import sys; print(sys.executable);'],
   argsVersion: ['-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);'],
-  semverRange: '2.7.x || >=3.5.0',
+  semverRange: '>=3.5.0',
```
cclauss, line 21:
```suggestion
  semverRange: '>=3.6.0',
```
See README.md

lib/find-python.js
```diff
@@ -18,16 +18,15 @@ PythonFinder.prototype = {
...
 
   // These can be overridden for testing:
   execFile: cp.execFile,
   env: process.env,
   win: win,
   pyLauncher: 'py.exe',
   winDefaultLocations: [
-    path.join(process.env.SystemDrive || 'C:', 'Python37', 'python.exe'),
-    path.join(process.env.SystemDrive || 'C:', 'Python27', 'python.exe')
+    path.join(process.env.SystemDrive || 'C:', 'Python37', 'python.exe')
```
cclauss, line 29: 36, 38, 39 are also valid.

mfonville: I think those default locations are not valid for Python 3.5 and higher anymore. See e.g. https://stackoverflow.com/questions/44749636/what-path-to-install-python-3-6-to-on-windows

For system-wide installs they are now in `%ProgramFiles%` and per user in `%LocalAppData%\Programs`.
Maybe you can look in both those directories and loop with decreasing the python version number, so that you find the highest installed version?
Though it gets complicated, so maybe drop it completely and just tell the user to set the path to python correctly?

DeeDeeG: > For system-wide installs they are now in %ProgramFiles% and per user in %LocalAppData%\Programs. 👍

cclauss (approved):
LGTM!  Thanks.

lib/find-python.js
DeeDeeG: @mfonville is right. `'C:\Python37'\python.exe` isn't the real format for the default install path for Python 3.

I can confirm from testing the installers, that the format would be one of the following:

- `C:\Users\[user]\AppData\Local\Programs\Python\Python39`
- `C:\Program Files\Python39`
- `C:\Users\[user]\AppData\Local\Programs\Python\Python39-32`
- `C:\Program Files (x86)\Python39-32`

Here is an example branch using those paths for all supported Python versions (3.6-3.9): https://github.com/DeeDeeG/node-gyp/compare/find-python-only-find-python-3...DeeDeeG:ae4e0b4c7a1e9a5c9a0b877c2ed737fe5c6332cc

This is not much slower at all, so I think it would be worth adding.

(Unfortunately I can't figure out how to adjust the tests to pass with this change. If anyone can assist with that, it would be much appreciated.)

DeeDeeG: I have fixed the tests when using the real Python 3 default install paths. ([Diff](https://github.com/DeeDeeG/node-gyp/compare/find-python-only-find-python-3...update-Python-fallback-paths-for-Windows), [CI](https://github.com/DeeDeeG/node-gyp/actions/runs/576058726)).

@cclauss should I add that to this PR? (Or perhaps wait for a follow-up PR?) Thanks.

Also: Python 3.10 will be out [in October](https://www.python.org/dev/peps/pep-0619/#release-schedule), and the [alpha release](https://www.python.org/downloads/release/python-3100a5/) seems to work with `node-gyp` in my testing. I could add 3.10 install paths as well.

cclauss: Please put those changes for the supported Pythons (3.6-3.9) in this PR but let's hold off on 3.10 until it is released.  We do not want to support any Python 3.10 breakage until it is officially released.  Thanks massively for your awesome work!

lib/find-python.js
```diff
@@ -26,7 +26,22 @@ PythonFinder.prototype = {
...
+    path.join(process.env.SystemDrive || 'C:', 'Program Files (x86)', 'Python38', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Users', process.env.USERNAME || process.env.USER, 'AppData', 'Local', 'Programs', 'Python', 'Python37', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Program Files', 'Python37', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Users', process.env.USERNAME || process.env.USER, 'AppData', 'Local', 'Programs', 'Python', 'Python37-32', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Program Files (x86)', 'Python37', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Users', process.env.USERNAME || process.env.USER, 'AppData', 'Local', 'Programs', 'Python', 'Python36', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Program Files', 'Python36', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Users', process.env.USERNAME || process.env.USER, 'AppData', 'Local', 'Programs', 'Python', 'Python36-32', 'python.exe'),
+    path.join(process.env.SystemDrive || 'C:', 'Program Files (x86)', 'Python36', 'python.exe')
```
cclauss: This seems like a lot of boilerplate code with tons of possibilities for typos.  Would there be a way to use a list-driven approach to remove the repetitive code.  `for python in ["Python39", "Python38", "Python37", "Python36"]:`

mfonville: I just wanted to comment the same thing, but with `'39', '38', '37', '36'` :-)

lib/find-python.js
DeeDeeG: The system-wide Python install path for 32-bit Windows is `C:\Program Files\Python39-32`. (There is no `Program Files (x86)` on 32-bit Windows, only `Program Files`.)

 That's one more path to check per minor version.

lib/find-python.js
DeeDeeG: @cclauss @mfonville, sure. It should be as simple as constructing the array programmatically somewhere else in the file, assigning that to a variable, then assigning the exported array from the variable. (Instead of directly exporting all this verbose/hard-coded stuff.)

lib/find-python.js
```diff
@@ -1,13 +1,26 @@
...
 const extend = require('util')._extend // eslint-disable-line
 const win = process.platform === 'win32'
 const logWithPrefix = require('./util').logWithPrefix
 
+const systemDrive = process.env.SystemDrive || 'C:'
+const username = process.env.USERNAME || process.env.USER
+const winDefaultLocationsArray = []
+
+for (const majorMinor of ['39', '38', '37', '36']) {
+  winDefaultLocationsArray.push(
```
mfonville, line 15: Instead of fixed `Users` and `Program Files` shouldn't this be with the environment variables `%ProgramFiles%`, `%ProgramFiles(x86)%`, `%ProgramW6432%` and `%LocalAppData%`?

Because these folders can be set to be on other disks than the system drive, and avoids potential localization issues in the path.

gengjiawen (approved):

lib/find-python.js
DeeDeeG: @mfonville Thanks for the suggestion. I did implement your suggestion on another branch here: https://github.com/DeeDeeG/node-gyp/commit/e57fdcc5443e57be5c96107aab8ffafc3c2d5321

It does work. It's also a bit more involved. After some quick research and testing, I am somewhat doubtful whether there are circumstances where the paths actually change based on localization or user customization. See my thoughts below. If you have examples or more information about when/if this approach is necessary, please let me know.

At the moment I am inclined to believe this is not necessary. But I would not mind being proved wrong. Thanks.

---

My thoughts on this: I wish I knew in what circumstances the "Program Files" or "AppData" folders are localized or moved, if there are any such circumstances. I do not think the paths are truly localized on-disk, nor movable, in Windows 10. It's unclear to me if they've ever been truly localized on-disk, or movable in previous Windows releases.

_(Paths were slightly different in Windows XP, but Windows XP is not compatible with the latest `node-gyp`. Python 3.4 is the last version supporting Windows XP. `node-gyp` master branch now requires Python 3.6 and up, so no XP support anymore. And only Windows 10 is in "mainstream" support by Microsoft now.)_

Regarding custom locations for Program Files or AppData: Is anyone here aware of a proper way to move these folders? _(I am aware that folders such as Documents and Videos can be easily moved to another drive. But can "Program Files", user home folders, or "AppData" be moved off of the system drive? I have looked this up online, and the overall impression I am getting is that there is no official way to move these, and that moving them may break things.)_

Regarding localization, or Windows in languages other than English: I don't see the paths being actually localized on the disk. (On Windows 10 in Spanish, I see in the File Explorer, "C: > Usuarios > [user]`, but the moment you try to edit the path in the address bar, it changes to `C:\Users\[user]`. The env var `LOCALAPPDATA` is still `C:\Users\[user\\AppData\Local`. The env var `ProgramFiles` is still `C:\Program Files`. Even while this shows as "C:\Archivos de programa` in File Explorer, the moment you try to edit it, it changes to `C:\Program Files`.

mfonville: If XP is out of the picture, then I do think localization is probably indeed not an issue anymore. Windows 10 does this nicely.

But about custom paths, to be honest I am not working with Windows daily anymore (switched to Linux many years ago). But when a couple of years ago I did do some volunteering with managing some Windows systems, we could set custom locations for these paths through policies. And also I have seen customized paths when computers were upgraded from a HDD to a SSD, where these paths were customized to optimize performance.
I am not sure how Windows 10 handles it these days, but imho those environment variables are more reliable than any hardcoded path. So better to use the variables :-)

DeeDeeG: Thanks for that explanation. I also found this official Microsoft document about moved user folders: https://docs.microsoft.com/en-us/troubleshoot/windows-server/user-profiles-and-logon/relocation-of-users-and-programdata-directories

So apparently user directories _can_ be moved to a custom location. (Maybe this applies only to Windows Server? I'd say `node-gyp` is not any less important for Windows Server users.)

I'm a little more inclined to do this (https://github.com/DeeDeeG/node-gyp/commit/e57fdcc5443e57be5c96107aab8ffafc3c2d5321) now. Truth be told, it's not much more complex. Given even a mild justification for end-user benefit, I'd want to do it.

test/test-find-python.js
```diff
@@ -10,19 +10,15 @@ const PythonFinder = findPython.test.PythonFinder
...
-      if (/Python 2/.test(stderr)) {
-        t.strictEqual(stdout, '')
-        t.ok(/Python 2/.test(stderr))
-      } else {
-        t.ok(/Python 3/.test(stdout))
-        t.strictEqual(stderr, '')
-      }
+      t.notOk(/Python 2/.test(stderr))
+      t.ok(/Python 3/.test(stdout))
+      t.strictEqual(stderr, '')
```
DeeDeeG, line 21: I just noticed I kept redundant checks in this test:

`t.notOk(/Python 2/.test(stderr))` is redundant to
`t.strictEqual(stderr, '')` two lines later.

With permission from the maintainers/reviewers, I'd want to delete one of them, preferably the first one.

<details>

For context:
- `python2 -V` outputs the Python version on `stderr`, like this: `'Python 2.7.16'`.
- `python3 -V` outputs the version on `stdout` like this: `'Python 3.8.2'`.

This test is just checking version strings when printing the version of Python that `findPython()` found.

Example script, showing that Python 2 and Python 3 print to `stderr` and `stdout` respectively:

```js
const execFile = require('child_process').execFile

execFile('python2', ['-V'], function (err, stdout, stderr) {
  console.log('--- python2 -V output ---')
  console.log('stdout is: ' + stdout)
  console.log('stderr is: ' + stderr)
})

execFile('python3', ['-V'], function (err, stdout, stderr) {
  console.log('--- python3 -V output ---')
  console.log('stdout is: ' + stdout)
  console.log('stderr is: ' + stderr)
})
```

Output:

```
--- python2 -V output ---
stdout is: 
stderr is: Python 2.7.16

--- python3 -V output ---
stdout is: Python 3.8.2

stderr is: 
```

</details>

lib/find-python.js
```diff
@@ -1,13 +1,26 @@
...
-const path = require('path')
 const log = require('npmlog')
 const semver = require('semver')
 const cp = require('child_process')
 const extend = require('util')._extend // eslint-disable-line
 const win = process.platform === 'win32'
 const logWithPrefix = require('./util').logWithPrefix
 
+const systemDrive = process.env.SystemDrive || 'C:'
+const username = process.env.USERNAME || process.env.USER
```
DeeDeeG, line 11: I think this should fall back to assigning some string to `username`. Even something not very useful like `''` or `'User'`.

(It's possible both `process.env.USERNAME` and `process.env.USER` are `undefined`. Having a not very useful path to check is better than causing a `ReferenceError`.)

cclauss: I actually like the error message better than unsuccessfully trying to sweep the problem under the carpet.  The message makes it clear what the problem is.

Please do not worry about the Python 4 case.  We will cross that bridge later.

cclauss: Thinking a bit more...

If username is undefined and we can find a globally installed acceptable Python then we should use that without complaint.  If we fail to find an acceptable Python and username is undefined then we should raise the error.

lib/find-python.js
mfonville: And most of all, there is no downside to the change I think.

cclauss: Go for it.  This code is easy to read/ understand.

cclauss: I know that is the Swiss in me but...

Please sort your const declarations in the order that they are used to improve readability.

lib/find-python.js
cjihrig: You can also get the username via `os.userInfo()`.

lib/find-python.js
DeeDeeG: More research: We definitely don't have to worry about Windows XP or Vista.

Node v6 refuses to run on anything older than Windows 7. Node v5 is the newest Node that will run on Windows Vista. _Node v5 cannot even run the latest `node-gyp`_, as `node-gyp` user newer JavaScript features that aren't supported in Node 5. In conclusion: modern `node-gyp` won't run at all on Windows Vista, regardless of this PR.

DeeDeeG: > environment variables [. . .] `%ProgramW6432%` [. . .]

@mfonville thanks for mentioning that. I believe I will need to switch from `process.env.ProgramFiles` to `process.env.ProgramW6432` to get the "not (x86)" Program Files.  

<details>

It appears we have to use `process.env.ProgramW6432` to get the not-architecture-specific `Program Files` location. From 32-bit Node on a 64-bit host, `process.env.ProgramFiles` returns `[driveLetter]:\Program Files (x86)`. As such, `ProgramW6432` is the only easy way to get `[driveLetter]:\Program Files` from the environment across all combinations of host arch/node arch.

(Official documentation says that the env var `ProgramW6432` only exists as of Windows 7 or newer, but that's okay. `node-gyp` doesn't even run on Windows Vista or older.)

</details>

See:
- https://docs.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details?redirectedfrom=MSDN#environment-variables
- https://serverfault.com/questions/413320/programfiles-differences-on-64bit-windows

test/test-find-python.js
cclauss: You do not need to wait for permission.  If the test is redundant that it can be eliminated.

lib/find-python.js
DeeDeeG: Okay, I committed this in the PR now. `Program Files` and `AppData\Local` are set from environment variables now.

lib/find-python.js
DeeDeeG: Thanks for that suggestion @cjihrig. `os.userInfo()` should always return a username string. I committed that as a fallback. I think this can reasonably be expected to prevent `Uncaught ReferenceError: username is not defined`.

DeeDeeG:
I think the `py.exe` launcher works really well. I'd like to suggest it be tried before all these `winDefaultPaths` entries. I'm not sure if that's out-of-scope for this PR though.

(I noticed while testing this pull request that the `py.exe` launcher is run _after_ checking all these potential paths from `winDefaultPaths`.)

cclauss:
@cjihrig @mfonville Are you folks satisfied with the current state of this PR?  There is nice work here and it would be good to have more PR approvals besides @gengjiawen and myself.

mfonville:
I am happy with its current state. You could still consider @DeeDeeG's last suggestion to try `py.exe` before all the generated entries. That would probably be the logical thing to do.

cclauss:
@mfonville
1. https://github.com/nodejs/node-gyp/pull/2333/files
2. Review changes
3. Approve

mfonville (approved):

mfonville (approved):

lib/find-python.js
DeeDeeG: Note: There was apparently no risk of `Uncaught ReferenceError` after all.

<details>

If you write `.someNotDefinedThing` off of an object or array, you get `undefined` as the result, but no `ReferenceError`. Since `process.env` returns an object, doing `process.env.THIS_IS_NOT_DEFINED` does not produce a `ReferenceError`.

For example:

```
console.log(`Here is my undefined var: ${{}.some_undefined_var}`)
```

Outputs the string:

```
Here is my undefined var: undefined
```

So we would construct a Python path resembling this: `C:\\Users\\undefined\\AppData\\Local\\Programs\\Python\\Python39\\python.exe`

</details>

I'm happy with the `os.userInfo()` fallback. It makes it pretty much guaranteed we can find the username. This is just an interesting finding from my tests.

DeeDeeG:
Reminder: I think it might be neat to prioritize trying the `py.exe` launcher, _before_ checking all the default install paths, on Windows.

I have an implementation of this ready to go. (https://github.com/DeeDeeG/node-gyp/commit/c39ad57e2f0e9c094ed67fc831180577059d8625) Please let me know if this is preferred. And/or if you want that, but as another PR.

rvagg:
I'm happy to include this in v8.0.0 but there's been some additional commits since @gengjiawen and @cclauss approved, would either of you mind taking a look over the current state?
I'll squash it into a single commit for merging, unless anyone thinks there's separate things going on here worth individuating.

gengjiawen:
@cclauss Still looks good to you ?

cclauss:
This PR already has my approval.

lib/find-python.js
beltran-rubo: Related to this, still an issue when the username is not defined, only the user ID e.g. that is common in 'non-root' containers for OpenShift https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html#openshift-container-platform-specific-guidelines.

```
I have no name!@52545ce360b8$ node-gyp --help
os.js:272
    throw new ERR_SYSTEM_ERROR(ctx);
    ^

SystemError [ERR_SYSTEM_ERROR]: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
    at Object.userInfo (os.js:272:11)
    at Object.<anonymous> (/opt/bitnami/node/lib/node_modules/node-gyp/lib/find-python.js:11:76)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/opt/bitnami/node/lib/node_modules/node-gyp/lib/configure.js:11:18)
    at Module._compile (internal/modules/cjs/loader.js:999:30) {
  code: 'ERR_SYSTEM_ERROR',
  info: {
    errno: -2,
    code: 'ENOENT',
    message: 'no such file or directory',
    syscall: 'uv_os_get_passwd'
  },
  errno: [Getter/Setter],
  syscall: [Getter/Setter]
}
```

Python is in the path so a good option could be to add a fallback or not return an error for the line

```
const username = process.env.USERNAME || process.env.USER || require('os').userInfo().username
```

ytbigvinny27 (commented):
.

PR #2346, opened 2021-03-18
Release proposal: v8.0.0

rvagg:
I think this needs a dep update. If someone would like to help out on that - go look at https://github.com/npm/cli/blob/latest/package.json and try and match the overlapping semver specifiers to theirs. I tend to keep them roughly in sync to avoid duplication in their releases. Whether or not they'll take a v8.0.0 is another matter, and we may want to recommend they hold off for a little while until this is tested in the wild.

* [[`15a5c7d45b`](https://github.com/nodejs/node-gyp/commit/15a5c7d45b)] - **ci**: migrate deprecated grammar (#2285) (Jiawen Geng) [#2285](https://github.com/nodejs/node-gyp/pull/2285)
* [[`cc1cbce056`](https://github.com/nodejs/node-gyp/commit/cc1cbce056)] - **doc**: update macOS\_Catalina.md (#2293) (iMrLopez) [#2293](https://github.com/nodejs/node-gyp/pull/2293)
* [[`6287118fc4`](https://github.com/nodejs/node-gyp/commit/6287118fc4)] - **doc**: updated README.md to copy easily (#2281) (மனோஜ்குமார் பழனிச்சாமி) [#2281](https://github.com/nodejs/node-gyp/pull/2281)
* [[`66c0f04467`](https://github.com/nodejs/node-gyp/commit/66c0f04467)] - **doc**: add missing `sudo` to Catalina doc (Karl Horky) [#2244](https://github.com/nodejs/node-gyp/pull/2244)
* [[`a78b584236`](https://github.com/nodejs/node-gyp/commit/a78b584236)] - **(SEMVER-MAJOR)** **gyp**: remove support for Python 2 (#2300) (Christian Clauss) [#2300](https://github.com/nodejs/node-gyp/pull/2300)
* [[`c3c510d89e`](https://github.com/nodejs/node-gyp/commit/c3c510d89e)] - **gyp**: update gyp to v0.8.0 (#2318) (Christian Clauss) [#2318](https://github.com/nodejs/node-gyp/pull/2318)
* [[`9e1397c52e`](https://github.com/nodejs/node-gyp/commit/9e1397c52e)] - **(SEMVER-MAJOR)** **gyp**: update gyp to v0.7.0 (#2284) (Jiawen Geng) [#2284](https://github.com/nodejs/node-gyp/pull/2284)
[...]

PR #2348, opened 2021-03-22
deps: sync mutual dependencies with npm

DeeDeeG:
##### Checklist

##### Description of change

Sync semver ranges for packages defined in `package.json` to match the ranges `npm` uses. Keeping these in sync ensures `npm` can maximally dedupe its mutual dependencies with `node-gyp`. Leads to smaller install size for `npm` and most distributions of `node`. (As mentioned here: https://github.com/nodejs/node-gyp/pull/2346#issue-595221072)

This particular sync is probably not strictly necessary, since none of these were semver major dependency bumps, but it doesn't hurt either.

<details>

(Theoretically, this PR prevents older versions of `node-gyp`'s dependencies from being installed and not deduping with `npm`'s slightly stricter semver ranges with a higher minimum version. In practice, the deduping algorithm should prioritize the parent `npm` package's stricter range requirement, and redundant packages would still be avoided. But making these exactly match doesn't hurt.

As far as I can tell, maximal deduping is already achieved in practice at the `npm` repository, even without this PR, but I'm posting this Pull Request anyway, as @rvagg expressed interest in still syncing the dependencies.)

</details>

I opted for caret `^` semver rather than tilde `~` semver for the `npmlog` dependency, as matching their tilde `~` semver range would make `node-gyp` _more_ likely to go out of sync with `npm`, rather than less likely. Discussed previously here: https://github.com/nodejs/node-gyp/pull/2346#issuecomment-803465549 / https://github.com/nodejs/node-gyp/pull/2346#issuecomment-803756827 (Meaning the `npmlog` dependency is already as much in sync as we want it to be, so it saw no changes in this PR).


gengjiawen (approved):

rvagg:
merged @ 06ddde2

PR #2225, opened 2020-09-19
feat: promisify remove

imatlopez:
##### Checklist

- [x] `npm install && npm test` passes
- [ ] tests are included 
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

An effort to modernize the code

PR #2357, opened 2021-03-31
feat: promisify clean

imatlopez:
##### Checklist

- [ ] tests are included 

##### Description of change

Promisifying the clean action such that it is similar to #2220 and #2225

### DIFF
--- a/lib/clean.js
+++ b/lib/clean.js
@@ -2,14 +2,17 @@
 
 const rm = require('rimraf')
 const log = require('npmlog')
+const util = require('util')
 
-function clean (gyp, argv, callback) {
+async function clean (gyp, argv) {
   // Remove the 'build' dir
-  var buildDir = 'build'
+  const buildDir = 'build'
 
   log.verbose('clean', 'removing "%s" directory', buildDir)
-  rm(buildDir, callback)
+  return util.promisify(rm)(buildDir)
 }
 
-module.exports = clean
+module.exports = function (gyp, argv, callback) {
+  clean(gyp, argv).then(callback.bind(undefined, null), callback)
+}
 module.exports.usage = 'Removes any generated build files and the "out" dir'
--- a/lib/clean.js
+++ b/lib/clean.js
@@ -13,6 +13,7 @@ async function clean (gyp, argv) {
 }
 
 module.exports = function (gyp, argv, callback) {
-  clean(gyp, argv).then(callback.bind(undefined, null), callback)
+  // no response expected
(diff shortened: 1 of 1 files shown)


lib/clean.js
```diff
@@ -2,14 +2,17 @@
...
+  const buildDir = 'build'
 
   log.verbose('clean', 'removing "%s" directory', buildDir)
-  rm(buildDir, callback)
+  return util.promisify(rm)(buildDir)
 }
 
-module.exports = clean
+module.exports = function (gyp, argv, callback) {
+  clean(gyp, argv).then(callback.bind(undefined, null), callback)
```
DeeDeeG, line 16: Hi, this PR looks good to me other than a small thing!

There is a gray line `undefined` output to the console with this PR.

Example output:

Before:

```
% /Users/[me]/node-gyp/bin/node-gyp.js clean
gyp info using node-gyp@7.1.2
gyp info using node@10.24.1 | darwin | x64
gyp info ok 
```

After:

```
% /Users/[me]/node-gyp/bin/node-gyp.js clean
gyp info using node-gyp@7.1.2
gyp info using node@10.24.1 | darwin | x64
undefined
gyp info ok 
```

I think this line I am commenting on is implicitly returning a variable that exists but has value `undefined` to the callback. Whereas before the callback was simply called with no argument.

It's the difference between `console.log()` and `var a; console.log(a)`. One logs nothing, the other logs `undefined`.

My suggestion is to add parentheses to indicate calling the callback with no argument whatsoever if the Promise succeeded/was fulfilled. See code suggestion here:

```suggestion
  clean(gyp, argv).then(callback.bind(undefined, null)(), callback)
```

DeeDeeG: P.S. I admit I don't understand the need to `bind` the callback, and the output seems the same to me (i.e. there is no output) with or without using `.bind()`.

If it's possible to omit the `.bind()` in the Promise succeeded/fulfilled case, then I would personally think that's better for making the code simpler and more readable.

imatlopez: thanks, made the onFulfilled function simply call callback no args. Goal of mine is to replace the callbacks with promises altogether as the backbone.

DeeDeeG (approved):
FWIW (though I'm not a maintainer here) I approve because:

- It's functionally equivalent
- It uses Promises and modern `const` instead of callbacks and `var`

lib/clean.js
DeeDeeG: Thanks for the explanation!

Issue #2351, opened 2021-03-24
Python syntax error in dynamically generated 'build/gyp-mac-tool' file when python3 on the PATH is < 3.6


DeeDeeG:
* **Node Version**:  Latest 10, 12, 14 and 15 (10.24.0, 12.21.0, 14.16.0, 15.12.0)
* **Platform**:  macOS "Catalina" 10.15.7
* **Compiler**:  Apple clang version 12.0.0 (clang-1200.0.32.29)
* **Module**:  git-utils, leveldown, superstring

<details><summary>Verbose output (from node-gyp rebuild --verbose):</summary>

```
% /Users/[user]/node-gyp/bin/node-gyp.js rebuild --verbose
gyp info it worked if it ends with ok
gyp verb cli [
gyp verb cli   '/Users/[user]/n-prefix/bin/node',
gyp verb cli   '/Users/[user]/node-gyp/bin/node-gyp.js',
gyp verb cli   'rebuild',
gyp verb cli   '--verbose'
gyp verb cli ]
gyp info using node-gyp@7.1.2
gyp info using node@15.12.0 | darwin | x64
gyp verb command rebuild []
gyp verb command clean []
gyp verb clean removing "build" directory
gyp verb command configure []
gyp verb find Python Python is not set from command line or npm configuration
gyp verb find Python Python is not set from environment variable PYTHON
gyp verb find Python checking if "python3" can be used
gyp verb find Python - executing "python3" to get executable path
gyp verb find Python - executable path is "/Library/Frameworks/Python.framework/Versions/3.10/bin/python3"
gyp verb find Python - executing "/Library/Frameworks/Python.framework/Versions/3.10/bin/python3" to get version
gyp verb find Python - version is "3.10.0"
gyp info find Python using Python version 3.10.0 found at "/Library/Frameworks/Python.framework/Versions/3.10/bin/python3"
gyp verb get node dir no --target version specified, falling back to host node version: 15.12.0
gyp verb command install [ '15.12.0' ]
gyp verb install input version string "15.12.0"
gyp verb install installing version: 15.12.0
gyp verb install --ensure was passed, so won't reinstall if already installed
gyp verb install version is already installed, need to check "installVersion"
gyp verb got "installVersion" 9
gyp verb needs "installVersion" 9
gyp verb install version is good
gyp verb get node dir target node version installed: 15.12.0
gyp verb build dir attempting to create "build" dir: /Users/[user]/github/node_modules/superstring/build
[...]

PR #2355, opened 2021-03-29
gyp: update gyp to v0.8.1

DeeDeeG:
##### Checklist

- [x] `npm install && npm test` passes
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

Ran `./update-gyp.py v0.8.1` to pull in `gyp` from `gyp-next` v0.8.1

- See the CHANGELOG: https://github.com/nodejs/gyp-next/releases/tag/v0.8.1
- and see the commit-for-commit comparison: https://github.com/nodejs/gyp-next/compare/v0.8.0...v0.8.1

In terms of actual source code changes, this is strictly updating the shebang lines of various Python source code files from `#!/usr/bin/env python` to `#!/usr/bin/env python3`... and a couple of functionally identical whitespace/indentation changes. (Other changes are version updates to GitHub actions in workflows for that repository, and a new entry in `gyp`'s `CHANGELOG.md` -- not functionally impactful to `node-gyp`.)

Partially addresses https://github.com/nodejs/node-gyp/issues/2351. (Addresses https://github.com/nodejs/node-gyp/issues/2351 for users with Python 3 >=3.6 as the first `python3` on their PATH.)

PR #2254, opened 2020-11-06
Add support for Unicode characters in paths

owl-from-hogvarts:
##### Checklist

- [x] `npm install && npm test` passes (`npm i` pass but `npm test` not, becouse test was written not for my changes)
- [x] tests are included 
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

On Windows 10 i discovered an issue that if there are
non-english letters or symbols in path
for python, find-python.js script can't find it.
This bug is couse by encoding issue which (i hope)
I have fixed. At least this bug fix works for me.

Have changed:
--modified:   gyp/pylib/gyp/easy_xml.py
----python 2.7 handle xml_string in wrong way
----becouse of encoding (line 128)
--modified:   gyp/pylib/gyp/input.py
----if "encoding:utf8" on line 240 doesn't marked cause to error
--new file:   lib/find-python-script.py
----i have created this file for convience.
----Script which can handle non-eanglish
----letters and symbols cann't °fit one lne
----becouse it is necessary to specify
----instarctions for several versions of python
--modified:   lib/find-python.js
----to make js understand python (line 249)

PR #2362, opened 2021-04-08
Add Python symlink to path (for non-Windows OSes only)

DeeDeeG:
##### Checklist

- [ ] tests are included 

##### Description of change

An attempt to fix https://github.com/nodejs/node-gyp/issues/2351 for the rest of users not yet addressed by https://github.com/nodejs/node-gyp/pull/2355.

- Symlink the valid Python binary found by `lib/find-python.js` at `build/node_gyp_bins/python3`
- Add `build/node_gyp_bins` to the `PATH` just before building
  - Ensures any of the build-time Python scripts (such as `build/gyp-mac-tool` or `build/gyp-flock-tool`) will be run with the Python binary validated by`lib/find-python.js`

##### More notes/thoughts/explanation

**The corner case/bug is not applicable to Windows. So this PR changes nothing on Windows.**

<details>

This PR changes nothing on Windows. The bug being fixed is Not Applicable to Windows; There are no Python scripts to run at build time on Windows, per my testing, nor as far as I can deduce from research.

_(If someone can prove that there are ever Python scripts at build-time on Windows, or the maintainers prefer to do something on Windows just in case, I can work on a solution involving writing a batch script to `build/node_gyp_bins/python3.bat` that runs the validated Python binary from `lib/find-python.js`. That should work just as well. Non-administrator users cannot create symlinks by default on Windows, so in my opinion we can't rely on symlinks on Windows. But again, I don't think there's any Python to run during the build phase on Windows. So this all should be moot!)_

</details>

**Who is this for?**

Specifically, folks with older or somehow invalid `python3` (< 3.6) first on their `PATH`, who have nevertheless managed to supply a newer, valid Python binary to `node-gyp`, probably by command-line `--python=/some/python3` or as an env var `PYTHON=/some/python3`.

<details>

_(Note that Debian [Jessie](https://packages.debian.org/jessie/python3) and [Stretch](https://packages.debian.org/jessie/python3), as well as Ubuntu [Xenial](https://packages.ubuntu.com/xenial/python3) all ship `python3` older than 3.6. Perhaps these users will get newer Python3 and opt not to put it early on the PATH?)_

</details>

**What does this PR achieve?** A user with older Python on their PATH who manages to supply a valid Python version to `node-gyp` and make it through the configure stage (including `lib/find-python.js`) should not abruptly hit errors at the build stage. This PR aims to solve that corner case. (See: https://github.com/nodejs/node-gyp/issues/2351 for details.)

**Where did the idea come from?** This was inspired by the approach that `nodejs/node` takes when setting up the build for NodeJS itself. See `configure.py` at that repository:

- https://github.com/nodejs/node/blob/v15.14.0/configure.py#L1826-L1857 and
- https://github.com/nodejs/node/blob/v15.14.0/configure.py#L1960-L1965

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -185,6 +185,13 @@ function build (gyp, argv, callback) {
       }
     }
 
+    if (!win) {
+      // Add build-time dependency symlinks (such as Python) to PATH
+      const buildBinsDir = path.resolve('build', 'node_gyp_bins')
+      process.env.PATH = `${buildBinsDir}:${process.env.PATH}`
+      log.verbose('bin symlinks', `adding symlinks (such as Python), at "${buildBinsDir}", to PATH`)
+    }
+
     var proc = gyp.spawn(command, argv)
     proc.on('exit', onExit)
   }
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -100,13 +100,13 @@ function configure (gyp, argv, callback) {
 
     fs.unlink(symlinkDestination, function (err) {
       if (err && err.code !== 'ENOENT') {
-        log.warn('python symlink', 'error when attempting to remove existing symlink')
-        log.warn('python symlink', err.stack, 'errno: ' + err.errno)
+        log.verbose('python symlink', 'error when attempting to remove existing symlink')
+        log.verbose('python symlink', err.stack, 'errno: ' + err.errno)
       }
       fs.symlink(python, symlinkDestination, function (err) {
         if (err) {
-          log.warn('python symlink', 'error when attempting to create Python symlink')
-          log.warn('python symlink', err.stack, 'errno: ' + err.errno)
+          log.verbose('python symlink', 'error when attempting to create Python symlink')
+          log.verbose('python symlink', err.stack, 'errno: ' + err.errno)
         }
       })
     })
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -100,11 +100,13 @@ function configure (gyp, argv, callback) {
 
     fs.unlink(symlinkDestination, function (err) {
       if (err && err.code !== 'ENOENT') {
-        log.warn('python symlink', 'error when attempting to remove existing symlink\n', err)
+        log.warn('python symlink', 'error when attempting to remove existing symlink')
+        log.warn('python symlink', err.stack)
       }
       fs.symlink(python, symlinkDestination, function (err) {
         if (err) {
-          log.warn('python symlink', 'error when attempting to create Python symlink\n', err)
+          log.warn('python symlink', 'error when attempting to create Python symlink')
+          log.warn('python symlink', err.stack)
         }
       })
     })
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -98,8 +98,15 @@ function configure (gyp, argv, callback) {
 
     log.verbose('python symlink', `creating symlink to "${python}" at "${symlinkDestination}"`)
 
-    fs.unlink(symlinkDestination, function () {
-      fs.symlink(python, symlinkDestination, function () {})
+    fs.unlink(symlinkDestination, function (err) {
+      if (err && err.code !== 'ENOENT') {
+        log.warn('python symlink', 'error when attempting to remove existing symlink\n', err)
+      }
+      fs.symlink(python, symlinkDestination, function (err) {
+        if (err) {
+          log.warn('python symlink', 'error when attempting to create Python symlink\n', err)
+        }
+      })
     })
   }
 
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -101,12 +101,12 @@ function configure (gyp, argv, callback) {
     fs.unlink(symlinkDestination, function (err) {
       if (err && err.code !== 'ENOENT') {
         log.warn('python symlink', 'error when attempting to remove existing symlink')
-        log.warn('python symlink', err.stack)
+        log.warn('python symlink', err.stack, 'errno: ' + err.errno)
       }
       fs.symlink(python, symlinkDestination, function (err) {
         if (err) {
           log.warn('python symlink', 'error when attempting to create Python symlink')
-          log.warn('python symlink', err.stack)
+          log.warn('python symlink', err.stack, 'errno: ' + err.errno)
         }
       })
     })
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -17,6 +17,7 @@ if (win) {
 function configure (gyp, argv, callback) {
   var python
   var buildDir = path.resolve('build')
+  var buildBinsDir = path.join(buildDir, 'node_gyp_bins')
   var configNames = ['config.gypi', 'common.gypi']
   var configs = []
   var nodeDir
@@ -73,7 +74,9 @@ function configure (gyp, argv, callback) {
 
   function createBuildDir () {
     log.verbose('build dir', 'attempting to create "build" dir: %s', buildDir)
-    fs.mkdir(buildDir, { recursive: true }, function (err, isNew) {
+
+    const deepestBuildDirSubdirectory = win ? buildDir : buildBinsDir
+    fs.mkdir(deepestBuildDirSubdirectory, { recursive: true }, function (err, isNew) {
       if (err) {
         return callback(err)
       }
@@ -84,8 +87,19 @@ function configure (gyp, argv, callback) {
         findVisualStudio(release.semver, gyp.opts.msvs_version,
           createConfigFile)
       } else {
+        createPythonSymlink()
         createConfigFile()
       }
+    })
+  }
+
+  function createPythonSymlink () {
+    const symlinkDestination = path.join(buildBinsDir, 'python3')
+
+    log.verbose('python symlink', `creating symlink to "${python}" at "${symlinkDestination}"`)
+
+    fs.unlink(symlinkDestination, function () {
+      fs.symlink(python, symlinkDestination, function () {})
     })
   }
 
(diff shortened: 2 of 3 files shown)


imatlopez (approved):

DeeDeeG:
Hi folks, just wanted to mention the other solution I found for fixing this bug. It involves patching `gyp`'s Makefile generator to embed the absolute path to the Python binary we want into the generated Makefile. (And check if that binary still exists at build time, with `command -v /path/to/python3`, otherwise fall back to whatever's first on the PATH again.)

See this commit for the alternative fix: https://github.com/DeeDeeG/node-gyp/commit/a68d184c75faf5cb36653d19dda8db85ab034f9f

<details><summary>Weighing the pros and cons of this alternate approach. (Click to expand)</summary>

Pros:
- I think the Makefile generator patch is more surgically precise than adding a new directory to the PATH.

Cons:
- but maintaining `node-gyp`-specific patches to `gyp` is more involved, and the patch might or might not be suitable to upstream into `gyp-next`. That's a a whole additional conversation that would need to be had before going with that approach.
   - (Where to land this? How to maintain the patch if not upstreamed? Consider any potential implications for various `gyp` consumers such a `nodejs/node` itself and various users of `gyp-next`, e.g. via the pip package?)
- If we ever enable the Ninja generator with `node-gyp`, a similar approach would be needed patching that generator  as well. Ditto for the various generators at `gyp/pylib/gyp/generator/`, although none of the other generators seem as likely to be brought up with `node-gyp` as the ninja one does. (See https://github.com/nodejs/node-gyp/pull/2171 for tentative/draft ninja support).

I am open to either approach, but this PR currently has the solution that is the more general one (no need to patch each individual generator, just set the environment up the way we want and it all works after that), and one which is easier to land without considering implications for `gyp-next` and its consumers.

</details>

DeeDeeG:
I added a couple of commits just now to pare down redundant information in the warning output and improve its formatting. (Although, of course, in most cases I expect there will be no warnings at all.)

<details><summary>Before (click to expand)</summary>

```
# ...
gyp info find Python using Python version 3.10.0 found at "/usr/local/bin/python3"
gyp WARN python symlink Error: EACCES: permission denied, unlink '/Users/[user]/superstring/build/node_gyp_bins/python3'
gyp WARN python symlink  error when attempting to remove existing symlink
gyp WARN python symlink  [Error: EACCES: permission denied, unlink '/Users/[user]/superstring/build/node_gyp_bins/python3'] {
gyp WARN python symlink   errno: -13,
gyp WARN python symlink   code: 'EACCES',
gyp WARN python symlink   syscall: 'unlink',
gyp WARN python symlink   path: '/Users/[user]/superstring/build/node_gyp_bins/python3'
gyp WARN python symlink }
gyp WARN python symlink Error: EEXIST: file already exists, symlink '/usr/local/bin/python3' -> '/Users/[user]/superstring/build/node_gyp_bins/python3'
gyp WARN python symlink  error when attempting to create Python symlink
gyp WARN python symlink  [Error: EEXIST: file already exists, symlink '/usr/local/bin/python3' -> '/Users/[user]/superstring/build/node_gyp_bins/python3'] {
gyp WARN python symlink   errno: -17,
gyp WARN python symlink   code: 'EEXIST',
gyp WARN python symlink   syscall: 'symlink',
gyp WARN python symlink   path: '/usr/local/bin/python3',
gyp WARN python symlink   dest: '/Users/[user]/superstring/build/node_gyp_bins/python3'
gyp WARN python symlink }
gyp info spawn /usr/local/bin/python3
# ...
```

</details>

<details><summary>After (click to expand)</summary>

```
# ...
gyp info find Python using Python version 3.10.0 found at "/usr/local/bin/python3"
gyp WARN python symlink error when attempting to remove existing symlink
gyp WARN python symlink Error: EACCES: permission denied, unlink '/Users/[user]/superstring/build/node_gyp_bins/python3' errno: -13
gyp WARN python symlink error when attempting to create Python symlink
gyp WARN python symlink Error: EEXIST: file already exists, symlink '/usr/local/bin/python3' -> '/Users/[user]/superstring/build/node_gyp_bins/python3' errno: -17
gyp info spawn /usr/local/bin/python3
# ...
```

</details>

I don't want these warnings to be too scary or attract too much attention in case we go on to successfully build the module. I just want them to be there to let folks know what's going on behind the scenes, and in case of an eventual build failure along the lines of https://github.com/nodejs/node-gyp/issues/2351.

DeeDeeG:
**Question for reviewers:** Should this PR use `rimraf` instead of `fs.unlink()` to remove existing symlinks?

<details>

I have gone with `fs.unlink()` for simplicity, and because it looked like the suitable option from browsing the Node `fs` API documentation. But I notice elsewhere in the repository the code uses `rimraf` to remove files/dirs. Apparently `rimraf` is a more sophisticated and reliable way to remove a file? Hopefully the symlink is not in use at `configure` time, so the chances of errors are lower. But I can use `rimraf` if we want to make a stronger effort to remove the symlink, and to be consistent with the rest of the project, at the cost of bringing a slightly more heavy-duty tool to bear on a pretty simple and usually optional task. I would go either way, but sticking with my first instinct `fs.unlink()` unless advised otherwise. Thanks for considering.

</details>

imatlopez:
I'd stick with rimraf, symlinks in windows are finicky at best so I'd go for the solution that has worked so far.

DeeDeeG:
`rimraf()` vs `fs.unlink()`: I suppose `rimraf()` is still very fast to execute. I wonder if it is slower to load...? Perf impact is probably low, I just have a hard time benchmarking load time. The "optimize it" part of me wants to use `fs.unlink()`, but the hyper-reliability part of me wants to use `rimraf()`.

@imatlopez thanks for the feedback. But this PR isn't doing symlinks on Windows, only on other OSes.

<details><summary>More on why symlinks are weird on Windows, but this is a bit off-topic</summary>

Yeah, I agree symlinks on Windows are weird from what I've heard. And they require Administrator permissions or a special "Developer Mode" toggle in the Settings app, so the vast majority of people probably can't use symlinks on Windows.

</details>

owl-from-hogvarts:
> Non-administrator users cannot create symlinks by default on Windows, so in my opinion we can't rely on symlinks on Windows

But in new tests in #2254 (`test/test-find-python.js:321`) there is check that make symlink (to be honest `junction`) to python's directory and it works well

DeeDeeG:
@cclauss @gengjiawen this would help users with Python 3.5 first on the PATH and newer Python 3.x located somewhere else. (Debian 9 users maybe?)

(Without this PR, for those users, `gyp` will ignore the compatible/new enough version of Python; Instead, it would use the first `python3` on the PATH, due to simply interpreting `gyp`'s `#!/usr/bin/env python3` shebang lines.)

Thoughts? Please take a look, thank you.

cclauss:
Please fix the conflicts.  Our minimum Python is now 3.7 in alignment with README.md and https://devguide.python.org/#status-of-python-branches.  How does this code ensure we are not building a symlink to an invalid `python3`?

I like this approach because it is JavaScript-based.  We improve maintainability when we remove Python code from gyp.

lib/build.js
```diff
@@ -185,6 +185,13 @@ function build (gyp, argv, callback) {
       }
     }
 
+    if (!win) {
+      // Add build-time dependency symlinks (such as Python) to PATH
+      const buildBinsDir = path.resolve('build', 'node_gyp_bins')
+      process.env.PATH = `${buildBinsDir}:${process.env.PATH}`
+      log.verbose('bin symlinks', `adding symlinks (such as Python), at "${buildBinsDir}", to PATH`)
```
cclauss, line 192:
```suggestion
      log.verbose('bin symlinks', `adding symlinks (such as Python 3), at "${buildBinsDir}", to PATH`)
```

lib/configure.js
```diff
@@ -81,11 +84,31 @@ function configure (gyp, argv, callback) {
...
+        createPythonSymlink()
         createConfigFile()
       }
     })
   }
 
+  function createPythonSymlink () {
+    const symlinkDestination = path.join(buildBinsDir, 'python3')
+
+    log.verbose('python symlink', `creating symlink to "${python}" at "${symlinkDestination}"`)
```
cclauss, line 96:
```suggestion
    log.verbose('python symlink', `creating symlink to "${python3}" at "${symlinkDestination}"`)
```

cclauss: repeat below?

DeeDeeG: In this string interpolation, `python` is a JS variable holding the path on disk to the Python binary chosen by `node-gyp`'s own `find-python.js`.

https://github.com/nodejs/node-gyp/blob/v9.0.0/lib/configure.js#L29

This variable's name doesn't functionally matter -- changing it would be only a cosmetic change, strictly for readers of this repo's JS code. We could change this variable in this file to be named `python3`, but would not change what is being logged. What is actually being logged (path to a usable Python binary) is determined by the logic in `find-python.js`.

lib/build.js
DeeDeeG: Feels like bikeshedding, but I don't like how specific this is. Implicitly, all Python in this version of `node-gyp` and beyond would be Python 3. There will never be an ambiguity in which major version of Python was symlinked.

Unless the Pythonistas unleash Python 4 on us, at which point heaven save us.

DeeDeeG:
>  Our minimum Python is now 3.7 in alignment with README.md and https://devguide.python.org/#status-of-python-branches. How does this code ensure we are not building a symlink to an invalid `python3`?

That is a matter handled in `node-gyp/lib/find-python.js`.

(This PR makes a symlink to whatever copy of Python `find-python.js` found.)

cclauss (approved):

Issue #2101, opened 2020-04-22
TypeError: 'cmp' is an invalid keyword argument for sort()


xxoo:
* **Node Version**: 13.13.0
* **Platform**: macos 10.15
* **Compiler**: XCode 11.4.1

<details><summary>Verbose output (from npm or node-gyp):</summary>

```
% node-gyp configure -- -f xcode 
gyp info it worked if it ends with ok
gyp info using node-gyp@6.1.0
gyp info using node@13.13.0 | darwin | x64
gyp info find Python using Python version 3.7.3 found at "/Volumes/data/Applications/Xcode.app/Contents/Developer/usr/bin/python3"
gyp http GET https://nodejs.org/download/release/v13.13.0/node-v13.13.0-headers.tar.gz
gyp http 200 https://nodejs.org/download/release/v13.13.0/node-v13.13.0-headers.tar.gz
gyp http GET https://nodejs.org/download/release/v13.13.0/SHASUMS256.txt
gyp http 200 https://nodejs.org/download/release/v13.13.0/SHASUMS256.txt
gyp info spawn /Volumes/data/Applications/Xcode.app/Contents/Developer/usr/bin/python3
gyp info spawn args [
gyp info spawn args   '/Volumes/data/lib/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'xcode',
gyp info spawn args   '-I',
gyp info spawn args   '/Volumes/data/Projects/mem/mem/build/config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/Volumes/data/lib/node_modules/node-gyp/addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/Users/xx/Library/Caches/node-gyp/13.13.0/include/node/common.gypi',
gyp info spawn args   '-Dlibrary=shared_library',
gyp info spawn args   '-Dvisibility=default',
gyp info spawn args   '-Dnode_root_dir=/Users/xx/Library/Caches/node-gyp/13.13.0',
gyp info spawn args   '-Dnode_gyp_dir=/Volumes/data/lib/node_modules/node-gyp',
gyp info spawn args   '-Dnode_lib_file=/Users/xx/Library/Caches/node-gyp/13.13.0/<(target_arch)/node.lib',
gyp info spawn args   '-Dmodule_root_dir=/Volumes/data/Projects/mem/mem',
[...]

PR #2383, opened 2021-04-24
gyp: use keys in calls to sorted

nwhetsell:
##### Checklist

- [ ] tests are included

##### Description of change

This replaces uses of the cmp parameter of sorted (which was removed in Python 3) with either the key parameter, or no parameter where key seems redundant. This also encodes Unicode strings before hashing them (which is needed in Python 3).

I’m honestly not sure if this is merge-able since this is a change to a dependency. I’m happy to attempt to write a test for this, but I’d need some guidance on where to put this.

Fixes: https://github.com/nodejs/node-gyp/issues/2101

### DIFF
--- a/gyp/pylib/gyp/xcodeproj_file.py
+++ b/gyp/pylib/gyp/xcodeproj_file.py
@@ -138,7 +138,9 @@ a project file is output.
 """
 
 import gyp.common
+from functools import cmp_to_key
 import hashlib
+from operator import attrgetter
 import posixpath
 import re
 import struct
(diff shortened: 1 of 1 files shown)


DeeDeeG:
Hi @nwhetsell,

> I’m honestly not sure if this is merge-able since this is a change to a dependency. I’m happy to attempt to write a test for this, but I’d need some guidance on where to put this.

The Node ecosystem uses its own fork of `gyp` that is more up-to-date than the largely unmaintained Google upstream project. It is developed as `gyp-next`, and the repo is located here: https://github.com/nodejs/gyp-next

Changes to the `gyp` dependency in `node-gyp` go to `gyp-next` first, and that get synced into `node-gyp` as the maintainers see fit.

nwhetsell:
Thanks for that information; I‘ve opened https://github.com/nodejs/gyp-next/pull/106 to address this. Sorry for the noise!

PR #2428, opened 2021-06-04
Brand new `find-python.js`

owl-from-hogvarts:
##### Checklist

- [x] tests are included 
- [x] in code documentation is changed or added
- +- (old no, but new yes) commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

## Description of change
**Note: i preserved old api** 

So... This changes are the logical continuation of my previos pull request #2254.
The purpose of changes is to modernize `find-python.js` namely:
 - improve code readability
 - use modern way to work with *async* stuff (using *promises* instead of *callback hell*)
 - improve cooments
 - improve tests (in fact the test for #2254 are cherry-picked from here)
 - modernize logging system
 - make log output more clear
 - colorize output
 - make `find-python.js` easly extencible (thanks to promises and new architecture)

And as I think i have achived that)

### Architecture

***assuming that #2254 is merged into master***

The common idea of checking haven't changed because old method is working pretty well

The main change in architecture is all about working with async logic. Before `find-python.js` used callbacks to make deal wiht async stuff. It was not so convinent because of error handling. It was necessary to path error callback down to the call tree. That mean that the programmer need to take eyes on that and even minimal carelessness could lead to error not to be handled. So this approach isn't save enough. We have wonderful `Promises` which make work with async very convinent as we can no longer care to pass appropriate callback to matching function. More on this topic later.
One more not really big but important change is moving to ES6 syntax with with all its classes etc. This syntax sugar is well playing in terms of code readability.

#### runChecks()

As before we getting the arrays of checks via `getChecks()` function. Now `runChecks` use promises due the reasons discussed above. *By the way promises give as more extesible (because code is easier to read and i hope understand :smile: ).* Checks are running *one by one* (see note). As previosly we running `before` (also called `pretask`. For any check you can specify special function which will be executed just before check itself. Useful when checks need to be skiped under a certain condition) function on checks which have it (it is optional). *Have added default logging procedure for checks.* ***Do not use `before` function for logging only. If `before` is specified, then no standard logging procedure will be executed so you need to make logging by your hand***.

Note: I think checks can be easly make running simultaneously via `Promise.allSeatled()`. But this functionality need to be polifilled because node.js v10 does not support it :cry:. 

#### more on Promises

The main thing from where promises are comming is `PythonFinder.prototype.run`. It is wrapper around `childProcess.execFile`. We have a bit of logging stuff here and a huge comment on possible problems. Check it please. I think it need to be discussed. Important thing is that `find-python-script.py` mustn't write anything to `stderr` as it wil be considered as error.
I'd like to make the module inside more modular. Now we have three or so main check funtions. They do not act i would like: like a small functions with only one resposability which can be easily chaned into  conveyor. It whould be really cool change. I am already took some steps in this direction: distinguish and generalize logging and move as many error handling to one single place (more on that later).

Note to discuss: generally we can use  `utils.promisify` in `PythonFinder.prototype.run`???

#### The rudiments of the conveyor approach
`PythonFinder.prototype.checkCommand` and `PythonFinder.prototype.checkPyLauncher` became really thin by using conveyor approach: chaning operations one by one.

#### New error handling system
Now all errors are handled individually for each `check` and in one single place `PythonFinder.prototype.catchErrors`. So you can just throw error and it will be handled and logged properly (default logging mechanism is implemented). If you need some special logging for you error you can make *custom error* class which **should be** the successor of `ErrorWithData`. Later whould be really cool to move to typescript and use here common interface for logging.

#### Logging
As mentioned before we have almost aoutomotive logging so no need to make custom spikes. Errors are logged in `PythonFinder.prototype.catchErrors` either by default logger or by implemented `log` method of *custom error*, logging which idicates that the check is processing resides in `PythonFinder.prototype.runChecks`. 
The logger itself is a separate class. I decided to move it away from `PythonFinder` because logging is not responsability of `PythonFinder` so for logging stuff nothing to do there. `Logger` is injected in `PythonFinder` in constructor so can be easely used in it. In my opinion it whould be greate to make the `Logger` fully separated identity as now there prefix hardcoded in it. 
The `Logger` also include logic to highlight text for terminal 

<hr>
Edited:
Also pls see these PRs: owl-from-hogvarts/node-gyp/pull/13, owl-from-hogvarts/node-gyp/pull/12, owl-from-hogvarts/node-gyp/pull/11 as they contain changes that need to be discussed

<hr>

I hope i stated all my thoughts)

@rvagg @DeeDeeG @joaocgreis you are welcom to review this)


lukekarrys:
Thanks for this work. Unfortunately this file has recently been modernized in #2925 so there are a lot of conflicts here. I'm going to close this but feel free to open any new PRs to refactor, those are always appreciated.

Issue #2435, opened 2021-06-17
Document how to contribute to node-gyp's wiki


owl-from-hogvarts:
If someone wants to contribute to node-gyp wiki, how it should be done? Unfortunatly, github doesn't provide any prefered way for this.

Quick googling resulting to this: https://gist.github.com/larrybotha/10650410

So i think that link may be added to `CONTRIBUTION.md` to provide explicit and unified way to contribute to wiki

rvagg:
@owl-from-hogvarts the wiki kind of sucks because it's out-of-sight out-of-mind, and yet it's a valuable resource for people.

How about we fix it by importing the docs into this repo itself and allowing editing via PR? We could have a docs/ directory that initially pulls in the Wiki docs and leaves tombstones on each of those pages pointing to the new content.

Anyone else got thoughts on this?

owl-from-hogvarts:
As I understand, you, @rvagg, suggesting to move whole documentation to `docs/` folder and in wiki leave references to the actual files in `docs/` folder where the documentation will be stored? If so, I am encourage it!

rvagg:
See https://github.com/nodejs/node-gyp/pull/2440

owl-from-hogvarts:
Can i close this issue as it seems that it is resolved)

Issue #2272, opened 2020-12-02
Upgrading node-gyp independently of npm and handling multiple versions


rvagg:
_(This could be made into a top-level doc, or integrated into the README, it comes up enough. Please edit this post or comment below if you have improvement suggestions. If someone wants to turn this into a PR then you're welcome to.)_

## Wot?

npm ships with its own version of node-gyp. Older versions of npm have older versions of node-gyp. npm has caught up recently to node-gyp's release cycle but it's expected we'll get out of sync and there will be cases where you need an old npm but a newer node-gyp, or npm doesn't ship yet with a node-gyp that we've released and it has fixes you need.

Generally, npm's library files are installed inside your global "node_modules", where npm is installed (run `npm prefix` and add `lib/node_modules`, or just `node_modules` for Windows [I think, Windows person - please confirm and edit this]). There are some exceptions to this. Inside this global `node_modules/` there will be an `npm/` directory and inside this you'll find a `node_modules/node-gyp/` directory. So it may look something like `/usr/local/lib/node_modules/npm/node_modules/node-gyp/`. This is the version of node-gyp that ships with npm.

When you install a _new_ version of node-gyp with outside of npm, it'll go into your global node_modules, but not under the `npm/node_modules`. So that may look like `/usr/local/lib/node_modules/node-gyp/`. It'll have the `node-gyp` executable linked into your `PATH` so running `node-gyp` will use this version.

The catch is that npm won't use this version unless you tell it to, it'll keep on using the one you have installed. You need to instruct it to by setting the `node_gyp` config variable (which goes into your `~/.npmrc`). You do this by running the `npm config set` command as below. Then npm will use the command in the path you supply whenever it needs to build a native addon.

**Important**: You also need to remember to unset this when you upgrade npm with a newer version of node-gyp, or you have to manually keep your globally installed node-gyp to date. See "Undo" below.

## Linux and macOS

```
npm install --global node-gyp@latest
npm config set node_gyp $(npm prefix -g)/lib/node_modules/node-gyp/bin/node-gyp.js
```

`sudo` may be required for the first command if you get a permission error.

## Windows

@joaocgreis' Windows instructions from https://github.com/nodejs/node-gyp/issues/1753#issuecomment-501827267

### Windows Command Prompt

```
npm install --global node-gyp@latest
for /f "delims=" %P in ('npm prefix -g') do npm config set node_gyp "%P\node_modules\node-gyp\bin\node-gyp.js"
```

### Powershell

```
[...]

PR #2440, opened 2021-06-21
Merge wiki content into doc/ directory

rvagg:
Ref: #2435

The proposal here is to deprecate the wiki and replace the current docs with tombstones that point to their copies in the docs/ directory. Then documentation can be managed, organised and edited properly through the GitHub process, like we've done (quite effectively) with the macOS_Catalina.md (h/t to @cclauss for the great work on that).

GitHub wiki's are garbage for broad collaboration and mostly out-of-sight for regular contributors. We have the problem of having a wiki with some docs that are regularly viewed by users thanks to search results but not much effort goes into maintaining them.

This PR pulls all of the wiki content into docs/ including the entire commit history of the wiki, amending each commit with `doc(wiki): `. No other editing has taken place, the proposal is that the first step is to just import & tombstone, then editing and curation can take place after that (editing, removal, renaming, adding, etc.).


docs/Common-issues.md
```diff
@@ -0,0 +1,14 @@
+## Python Issues OSX
+
+Make sure you are using the native Python version in OSX.  If you use a MacPorts of HomeBrew version, you may run into problems.
```
cclauss, line 3:
```suggestion
Make sure you are using the native Python version in OSX.  If you use a MacPorts or HomeBrew version, you may run into problems.
```
https://www.macports.org and https://brew.sh are two separate macOS package managers.

I think we should remove these two paragraphs because Apple what’s stated that native Python will be removed from macOS.  I have been using brew Python from the beginning with zero issues and have seen no one complaining about `execvp`.

docs/Common-issues.md
```diff
@@ -0,0 +1,14 @@
...
+
+If you have issues with `execvp`, be sure to check your `$PYTHON` environment variable.  If it is not set to the native version, unset it and try again.
+
+Notes: https://gist.github.com/erichocean/5177582
```
cclauss, line 7: This is about node-gyp v0.91 yet the current node-gyp is v8.1.  Is there any evidence that this problem still exists?

owl-from-hogvarts:
Also would be cool to tell in readme where to find docs

rvagg:
I've added a brief note in the README pointing to this directory.

I didn't really want to get into the weeds editing the content, but perhaps some of it just needs to be retired completely. Perhaps the best way to do this is to just edit my branch directly, or PR against my branch for any changes that you think are necessary?

Home.md could probably be renamed to README.md. We could probably improve the filenames (the `"` in the binding.gyp file and the casualness of the names are probably not ideal). I'm not sure the "undeclared identifier" doc is necessary. Common-issues is ... maybe not so common. Maybe we even move the Catalina doc into this directory? (But do we leave a tombstone with a link to the new location for a while since it's probably got some link-fu out in the web).

tbh I don't really have time to make a big job of this but would appreciate help from others to make this land-able and retire the wiki if others see that as a worthy goal.

docs/Common-issues.md
cclauss: Let’s agree to land this first and then resolve these issues in future PR.

cclauss (approved):
Nice!  Thanks for doing this.

richardlau (approved):
SGTM. No objections to the plan to migrate off the wiki.

I think these docs will end up being included in node-gyp releases, which I don't personally have an issue with but feel like I should point out in case anyone else does.

owl-from-hogvarts:
I think we can somwhow force releas please to ingore them

rvagg:
Well, if you want to help get them up to scratch before we cut another release then it should be ok to merge now work on them immediately afterward. My preference would be to include them in releases, just not necessarily in their current form since they're not all helpful.

cclauss:
> I think these docs will end up being included in node-gyp releases

`.gitignore` to the rescue?

richardlau:
> > I think these docs will end up being included in node-gyp releases
> 
> `.gitignore` to the rescue?

I imagine if we're adding these docs to the repo so they can be edited via git commits/GitHub pull requests we do not want them git ignored :slightly_smiling_face:.

rvagg:
doh! the CI failure is real:

```
  "C:\Program Files\Git\bin\git.exe" checkout --progress --force -B rvagg/wiki-to-doc refs/remotes/origin/rvagg/wiki-to-doc
  Error: error: invalid path 'docs/"binding.gyp"-files-out-in-the-wild.md'
  Error: The process 'C:\Program Files\Git\bin\git.exe' failed with exit code 128
```

will have to change that filename before merging

rvagg:
renamed 3 docs to not have quotes, removed the TypedArray doc which is unhelpful

owl-from-hogvarts:
i think https://github.com/nodejs/node-gyp/issues/2272 mast be added to docs

cclauss:
The idea was to land this PR without further mods and then rebase PRs like #2272 afterwards.

gengjiawen (approved):

owl-from-hogvarts:
what is needed to land this PR?

rvagg:
@owl-from-hogvarts just waiting after my last push to see if there were any objections. I don't see any so let's make this happen and get on with it!

rvagg:
Merged and all Wiki pages updated to point to that directory. The wiki still remains editable by people with commit access here if anyone thinks there's a better way to communicate the move. I didn't point to specific documents, just the directory, so we have the freedom in the short term to rename and move things around.

Let's hold off on cutting a new release and shipping half baked docs until we're slightly more comfortable with them. But **please contribute your excellent documentation to help users of this project and so we have better docs to point people to when they show up asking questions on a daily basis**.

PR #2375, opened 2021-04-12
lib: fail gracefully if we can't find the username

DeeDeeG:
##### Checklist

- [x] `npm install && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

In `lib\find-python.js`...

- Only attempt to get `os.userInfo().username` in a `try {}` block, since it's not 100% guaranteed to succeed
- If we can't find the `Appdata\Local` directory, skip checking paths under `Appdata\Local`.

Fixes https://github.com/nodejs/node-gyp/issues/2373

Issue #2152, opened 2020-06-07
`node-gyp configure` and `node-gyp install` fail with Node < v10.12.0


DeeDeeG:
* **Node Version**:  Node >= v6.0.0, < Node v10.12
* **Platform**:  Any
* **Compiler**: Any
* **Module**: Any

<details><summary>Verbose output (from npm or node-gyp):</summary>

When using an unsupported version of Node (< 10.12.0), error happens every time running `node-gyp configure`, or every other time running `node-gyp install`.

```console
$ npx node-gyp@7 configure
npx: installed 99 in 4.99s
gyp info it worked if it ends with ok
gyp info using node-gyp@7.0.0
gyp info using node@8.17.0 | linux | x64
gyp info find Python using Python version 3.8.2 found at "/usr/bin/python3"
gyp ERR! configure error 
gyp ERR! stack Error: EEXIST: file already exists, mkdir '/home/[user]/nodegypreq/native-module/build'
gyp ERR! System Linux 5.4.0-33-generic
gyp ERR! command "/home/[user]/n-prefix/bin/node" "/home/[user]/.npm/_npx/217155/bin/node-gyp" "configure"
gyp ERR! cwd /home/[user]/nodegypreq/native-module
gyp ERR! node -v v8.17.0
gyp ERR! node-gyp -v v7.0.0
gyp ERR! not ok 

```

```console
$ npx node-gyp@7 install
gyp info it worked if it ends with ok
gyp info using node-gyp@7.0.0
gyp info using node@8.17.0 | linux | x64
gyp WARN install got an error, rolling back install
gyp ERR! install error 
gyp ERR! stack Error: EEXIST: file already exists, mkdir '/home/[user]/.cache/node-gyp/8.17.0'
gyp ERR! System Linux 5.4.0-33-generic
gyp ERR! command "/home/[user]/n-prefix/bin/node" "/home/[user]/.npm/_npx/216921/bin/node-gyp" "install"
gyp ERR! cwd /home/[user]/nodegypreq/native-module
gyp ERR! node -v v8.17.0
gyp ERR! node-gyp -v v7.0.0
gyp ERR! not ok 

```

</details>

#2123 switched from the `mkdirp` module to native Node API's `fs.mkdir` with option `{recursive: true}`.

This feature (native recursive mkdir) is only available in Node v10.12.0 and up.

[...]

PR #2441, opened 2021-06-21
Update wrong catch block

liviarett:
##### Checklist

- [ ] tests are included 

### DIFF
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -36,8 +36,7 @@ for (const majorMinor of ['39', '38', '37', '36']) {
 function getOsUserInfo () {
   try {
     return require('os').userInfo().username
-  } catch (e) {
-  }
+  } catch (e) {}
 }
 
 function PythonFinder (configPython, callback) {
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -36,7 +36,9 @@ for (const majorMinor of ['39', '38', '37', '36']) {
 function getOsUserInfo () {
   try {
     return require('os').userInfo().username
-  } catch {}
+  } catch (e) {
+    throw e
+  }
 }
 
 function PythonFinder (configPython, callback) {
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -37,7 +37,6 @@ function getOsUserInfo () {
   try {
     return require('os').userInfo().username
   } catch (e) {
-    throw e
   }
 }
 


cclauss:
https://stackoverflow.com/questions/3335376/why-do-good-programmers-sometimes-silently-swallow-exceptions

rvagg:
Yeah, sorry, this is intentional (IIRC), there's some OS' where this fails, I can't remember the details, it was some trimmed-down Linux, I'm sure `git blame` would point the way to the PR that introduced the quiet fail.

Is this causing problems for you?

liviarett:
hmm I had a few other issues that were causing me to get to this point anyway, but this is disguising the error, as it gives me 

```
gyp ERR! UNCAUGHT EXCEPTION
gyp ERR! stack /Users/liviarett/.nvm/versions/node/v8.17.0/lib/node_modules/node-gyp/lib/find-python.js:44
gyp ERR! stack   } catch {}
gyp ERR! stack           ^
gyp ERR! stack
gyp ERR! stack SyntaxError: Unexpected token {
gyp ERR! stack     at createScript (vm.js:80:10)
gyp ERR! stack     at Object.runInThisContext (vm.js:139:10)
gyp ERR! stack     at Module._compile (module.js:617:28)
gyp ERR! stack     at Object.Module._extensions..js (module.js:664:10)
gyp ERR! stack     at Module.load (module.js:566:32)
gyp ERR! stack     at tryModuleLoad (module.js:506:12)
gyp ERR! stack     at Function.Module._load (module.js:498:3)
gyp ERR! stack     at Module.require (module.js:597:17)
gyp ERR! stack     at require (internal/module.js:11:18)
gyp ERR! stack     at Object.<anonymous> (/Users/liviarett/.nvm/versions/node/v8.17.0/lib/node_modules/node-gyp/lib/configure.js:11:18)
```

but not a big problem :)

lib/find-python.js
```diff
@@ -36,7 +36,9 @@ for (const majorMinor of ['39', '38', '37', '36']) {
 function getOsUserInfo () {
   try {
     return require('os').userInfo().username
-  } catch {}
+  } catch (e) {
+    throw e
```
rvagg, line 40:
```suggestion
```

rvagg (changes requested):
ookkaayy ... my memory is failing me on this one, I see this particular change was introduced @ fca4795512c67dc8420aaa0d913b5b89a4b147f3 and that I must be thinking of another silent catch that we introduced eons ago.. So this is pretty new, and a legit parse error.

:thumbsup: so let's fix the parse error you're getting, but not force the throw.

rvagg:
ahha! I knew I wasn't going _that_ senile: https://github.com/nodejs/node-gyp/pull/1835 but it's in install.js, while this same code is now in find-python.js which was just recently refactored.

liviarett:
ah alright! nice, so we can at least fix the unexpected token thing

DeeDeeG:
@cclauss

The `catch {}` syntax is valid on our supported versions of Node (10.12 and newer). I honestly hadn't anticipated this would every run on Node older than that, but in hindsight that's a little optimistic.

The `catch {}` I added in https://github.com/nodejs/node-gyp/pull/2375 is admittedly a syntax error on older Node. Older node wants the error to be passed into the catch block: `catch (e) {}`.

[MDN says](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#browser_compatibility) this is  a change as of Node 10/Chrome 66, but I'm finding it hard to confirm that anywhere in NodeJS or Chrome's documentation/blogs. Maybe this NodeJS PR? https://github.com/nodejs/node/pull/20708 (landed for Node 10.2 and newer).

I tried logging the error (https://github.com/nodejs/node-gyp/pull/2375#issuecomment-818217760 --> [implementation](https://github.com/DeeDeeG/node-gyp/compare/master...DeeDeeG:fail-gracefully-winDefaultLocations-bigger-refactor-with-logging-and-comments#diff-389dd7e63af69fe94fc71e79916ed76b0114dfdf49efe26112a13efa9e6531f6R349-R359)), but that added complexity and size of code seemed a bit overkill, given we intend to fail gracefully and just move on. It is hard to properly add logging at file-level scope (outside of `PythonFinder`), and I found refactoring this into `PythonFinder` tricky or impossible, since this data needs to be computed for `winDefaultLocations` to exist when `PythonFinder` is made... That's as well as I know how to explain it, given my limited JS proficiency. (I am self-taught with JS, and `PythonFinder` honestly confuses me a bit.)

DeeDeeG (approved):
:+1: Avoids a syntax error on older Node (Node 9.x or older).

_(Note: Those versions of Node are technically not supported by the latest `node-gyp`, because of https://github.com/nodejs/node-gyp/pull/2123 --> https://github.com/nodejs/node-gyp/issues/2152. But I think not causing a syntax error in the program for older Node is still an improvement.)_

Due to https://github.com/nodejs/node-gyp/issues/2152, users should really be on Node 10.12 or newer when using the latest `node-gyp`, and ideally they would be on the latest Node 12.x or 14.x or 16.x. But I don't think fixing this syntax error for older Node (at the cost of three characters of JS) hurts anything.

rvagg:
:thumbsup: thanks @DeeDeeG, it's good to lean on the conservative side for node-gyp since it's in such wide usage. When people show up using older versions of Node.js (for whatever reason - sometimes they can't change) and well tell them that their old version of node-gyp doesn't work for their use-case, it's not fun to tell them they can't get the newer one cause their Node.js is old. So where the change is painless, like this one, let's opt for the conservative choice.

DeeDeeG:
Update: I found out that trying to use `node-gyp` 8 on older Node (Node < 10) is still rather busted, due to https://github.com/nodejs/node-gyp/pull/2220.

Here's the details and a workaround...

https://github.com/nodejs/node-gyp/pull/2220/ broke `node-gyp install` (and, by extension, broke `node-gyp rebuild`) on Node older than version 10, by relying on the [`stream.pipeline` Node API function](https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback) that was only added in Node v10.0.0. And by switching a bunch of `fs.x` to `fsPromises.x`. (The `fsPromises` API was also added in Node 10.0.0).

<details><summary>Workaround if you really need to run the latest <code>node-gyp</code> on Node 8.x would be:</summary>

Caveat/warning: All this only applies to `node-gyp`@8.1.0. Only tested somewhat working by running `node-gyp rebuild` via `npm rebuild`... Using `npm` 6.13.4 and Node 8.17.0. Only tested on Linux. Any other combination may not work, and none of this is actually officially supported by `node-gyp` or `npm` or Node, etc. Try at your own risk.

- Install older Node's development headers with an older `node-gyp` that works on older Node, such as Node 8.x.
  - (With the older Node version, run `npx node-gyp@6 install`)
- Then, go in with a text editor and manually edit the copy of `node-gyp` you want to run; Edit `node-gyp`'s `lib/find-python.js` file like this PR does; update the `catch {}` block in `lib/find-python.js` to pass in the error: `catch (e) {}`.
  - (Patch the file [here](https://github.com/nodejs/node-gyp/blob/v8.1.0/lib/find-python.js#L39).)
- Then in `lib/install.js`...
  -  Go in with a text editor and manually replace `const streamPipeline = util.promisify(stream.pipeline)` near the top of the file with an empty function: `const streamPipeline = function () {}`.
      - ([here](https://github.com/nodejs/node-gyp/blob/v8.1.0/lib/install.js#L15).)
  - And replace `fs.promises.stat()` with `fs.statSync()`, and replace `fs.promises.readFile()` with `fs.readFileSync()`, and delete the `await` immediately preceding these two `fs` API calls, since we're using basic synchronous code now, not Promises.
    - (Only needed in two places. [here](https://github.com/nodejs/node-gyp/blob/v8.1.0/lib/install.js#L57) and [here](https://github.com/nodejs/node-gyp/blob/v8.1.0/lib/install.js#L75).)

</details>

PR #2448, opened 2021-06-28
fix: doc how to update node-gyp independently from npm

owl-from-hogvarts:
##### Checklist

- [ ] tests are included 

##### Description of change

https://github.com/nodejs/node-gyp/issues/2272

### DIFF
--- a/docs/Updating-npm-bundled-node-gyp.md
+++ b/docs/Updating-npm-bundled-node-gyp.md
@@ -6,9 +6,9 @@ not running a [current version of node-gyp](https://github.com/nodejs/node-gyp/r
 `npm` bundles its own, internal, copy of `node-gyp`. This internal copy is independent of any globally installed copy of node-gyp that
 may have been installed via `npm install -g node-gyp`.
 
-Generally, npm's library files are installed inside your global "node_modules",where npm is installed (run `npm prefix` and add `lib/node_modules`, or just `node_modules` for Windows). There are some exceptions to this. Inside this global `node_modules/`there will be an `npm/` directory and inside this you'll find a `node_modules/node-gyp/` directory. So it may look something like `/usr/local/lib/node_modules/npm/node_modules/node-gyp/`. This is the version of node-gyp that ships with npm.
+Generally, npm's library files are installed inside your global "node_modules", where npm is installed (run `npm prefix` and add `lib/node_modules`, or just `node_modules` for Windows). There are some exceptions to this. Inside this global `node_modules/` there will be an `npm/` directory and inside this you'll find a `node_modules/node-gyp/` directory. So it may look something like `/usr/local/lib/node_modules/npm/node_modules/node-gyp/`. This is the version of node-gyp that ships with npm.
 
-When you install a _new_ version of node-gyp with outside of npm, it'll go into your global node_modules, but not under the `npm/node_modules`. So that may look like `/usr/local/lib/node_modules/node-gyp/`. It'll have the `node-gyp` executable linked into your `PATH` so running `node-gyp` will use this version.
+When you install a _new_ version of node-gyp outside of npm, it'll go into your global node_modules, but not under the `npm/node_modules`. So that may look like `/usr/local/lib/node_modules/node-gyp/`. It'll have the `node-gyp` executable linked into your `PATH` so running `node-gyp` will use this version.
 
 The catch is that npm won't use this version unless you tell it to, it'll keep on using the one you have installed. You need to instruct it to by setting the `node_gyp` config variable (which goes into your `~/.npmrc`). You do this by running the `npm config set` command as below. Then npm will use the command in the path you supply whenever it needs to build a native addon.
 
@@ -23,7 +23,6 @@ npm config set node_gyp $(npm prefix -g)/lib/node_modules/node-gyp/bin/node-gyp.
 `sudo` may be required for the first command if you get a permission error.
 
 ## Windows
-@joaocgreis' Windows instructions from [#1753 (comment)](https://github.comnodejs/node-gyp/issues/1753#issuecomment-501827267)
 
 ### Windows Command Prompt
 ```
@@ -38,7 +37,7 @@ npm prefix -g | % {npm config set node_gyp "$_\node_modules\node-gyp\bin\node-gy
 ```
 
 ## Undo
-**Beware** if you don't unset the `node_gyp` config option, npm will continue to use the globally installed version of node-gyp rather than the one it ships with,which may end up being newer.
+**Beware** if you don't unset the `node_gyp` config option, npm will continue to use the globally installed version of node-gyp rather than the one it ships with, which may end up being newer.
 
 ```
 npm config delete node_gyp
--- a/docs/Updating-npm-bundled-node-gyp.md
+++ b/docs/Updating-npm-bundled-node-gyp.md
@@ -6,63 +6,41 @@ not running a [current version of node-gyp](https://github.com/nodejs/node-gyp/r
 `npm` bundles its own, internal, copy of `node-gyp`. This internal copy is independent of any globally installed copy of node-gyp that
 may have been installed via `npm install -g node-gyp`.
 
-This means that while `node-gyp` doesn't get installed into your `$PATH` by default, npm still keeps its own copy to invoke when you
-attempt to `npm install` a native add-on.
+Generally, npm's library files are installed inside your global "node_modules",where npm is installed (run `npm prefix` and add `lib/node_modules`, or just `node_modules` for Windows). There are some exceptions to this. Inside this global `node_modules/`there will be an `npm/` directory and inside this you'll find a `node_modules/node-gyp/` directory. So it may look something like `/usr/local/lib/node_modules/npm/node_modules/node-gyp/`. This is the version of node-gyp that ships with npm.
 
-Sometimes, you may need to update npm's internal node-gyp to a newer version than what is installed. A simple `npm install -g node-gyp`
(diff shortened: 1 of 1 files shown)


cclauss:
My sense is that we should have a separate docs for Windows and Linux/macOS.  Most of the open issues and complexity is on the Windows side.  The Linux/macOS doc would be more streamlined.

owl-from-hogvarts:
IMO what you are suggesting should be done in separate pull request. This one is only for incldujg important notes to docs.

After including that we can start work on separating docs by platforms.

docs/Updating-npm-bundled-node-gyp.md
```diff
@@ -1,63 +1,40 @@
...
-$ npm explore npm/node_modules/npm-lifecycle -g -- npm install node-gyp@latest
+npm install --global node-gyp@latest
+npm config set node_gyp $(npm prefix -g)/lib/node_modules/node-gyp/bin/node-gyp.js
 ```
 
-If the command fails with a permissions error, please try `sudo` and then the command.
+`sudo` may be required for the first command if you get a permission error.
 
 ## Windows
+@joaocgreis' Windows instructions from [#1753 (comment)](https://github.comnodejs/node-gyp/issues/1753#issuecomment-501827267)
```
rvagg, line 20: not sure we need this line if you've captured the essence of it below

rvagg:
Seems fine to me but can you please go over the text and look for missing spaces, there's a bunch of run-ins in many of the lines. I did start a suggestion review but it impacts most of the first half of this at least. Maybe a copy/paste problem?

Also I'm fine with just getting this into a good state and then we could fork it into separate docs if it gets large enough.

docs/Updating-npm-bundled-node-gyp.md
```diff
@@ -1,63 +1,40 @@
-`npm` bundles its own, internal, copy of `node-gyp`. This internal copy is independent of any globally installed copy of node-gyp that
-you may have installed via `npm install -g node-gyp`.
+npm ships with its own version of node-gyp. Older versions of npm have olderversions of node-gyp. npm has caught up recently to node-gyp's release cycle butit's expected we'll get out of sync and there will be cases where you need an oldnpm but a newer node-gyp, or npm doesn't ship yet with a node-gyp that we'vereleased and it has fixes you need.
```
cclauss, line 1: My sense is that this sentence is more meandering and less clear than the original.

owl-from-hogvarts: agree

owl-from-hogvarts:
@cclauss can this be merged?

docs/Updating-npm-bundled-node-gyp.md
```diff
@@ -1,63 +1,41 @@
 `npm` bundles its own, internal, copy of `node-gyp`. This internal copy is independent of any globally installed copy of node-gyp that
 you may have installed via `npm install -g node-gyp`.
 
-This means that while `node-gyp` doesn't get installed into your `$PATH` by default, npm still keeps its own copy to invoke when you
-attempt to `npm install` a native add-on.
+Generally, npm's library files are installed inside your global "node_modules",where npm is installed (run `npm prefix` and add `lib/node_modules`, or just `node_modules` for Windows). There are some exceptions to this. Inside this global `node_modules/`there will be an `npm/` directory and inside this you'll find a `node_modules/node-gyp/` directory. So it may look something like `/usr/local/lib/node_modules/npm/node_modules/node-gyp/`. This is the version of node-gyp that ships with npm.
```
rvagg, line 4:
```suggestion
Generally, npm's library files are installed inside your global "node_modules", where npm is installed (run `npm prefix` and add `lib/node_modules`, or just `node_modules` for Windows). There are some exceptions to this. Inside this global `node_modules/` there will be an `npm/` directory and inside this you'll find a `node_modules/node-gyp/` directory. So it may look something like `/usr/local/lib/node_modules/npm/node_modules/node-gyp/`. This is the version of node-gyp that ships with npm.
```

docs/Updating-npm-bundled-node-gyp.md
```diff
@@ -1,63 +1,41 @@
...
 
-Sometimes, you may need to update npm's internal node-gyp to a newer version than what is installed. A simple `npm install -g node-gyp`
-_won't_ do the trick since npm will still continue to use its internal copy over the global one.
+When you install a _new_ version of node-gyp with outside of npm, it'll go into your global node_modules, but not under the `npm/node_modules`. So that may look like `/usr/local/lib/node_modules/node-gyp/`. It'll have the `node-gyp` executable linked into your `PATH` so running `node-gyp` will use this version.
```
rvagg, line 6:
```suggestion
When you install a _new_ version of node-gyp outside of npm, it'll go into your global node_modules, but not under the `npm/node_modules`. So that may look like `/usr/local/lib/node_modules/node-gyp/`. It'll have the `node-gyp` executable linked into your `PATH` so running `node-gyp` will use this version.
```

docs/Updating-npm-bundled-node-gyp.md
```diff
@@ -1,63 +1,41 @@
...
-$ npm explore npm/node_modules/npm-lifecycle -g -- npm install node-gyp@latest
+npm install --global node-gyp@latest
+npm config set node_gyp $(npm prefix -g)/lib/node_modules/node-gyp/bin/node-gyp.js
 ```
 
-If the command fails with a permissions error, please try `sudo` and then the command.
+`sudo` may be required for the first command if you get a permission error.
 
 ## Windows
+@joaocgreis' Windows instructions from [#1753 (comment)](https://github.comnodejs/node-gyp/issues/1753#issuecomment-501827267)
```
rvagg, line 21:
```suggestion
```

docs/Updating-npm-bundled-node-gyp.md
```diff
@@ -1,63 +1,41 @@
...
-cd node_modules\npm\node_modules\npm-lifecycle
+npm install --global node-gyp@latest
+npm prefix -g | % {npm config set node_gyp "$_\node_modules\node-gyp\bin\node-gypjs"}
 ```
 
-Finish by running:
-```bash
-$ npm install node-gyp@latest
+## Undo
+**Beware** if you don't unset the `node_gyp` config option, npm will continue to use the globally installed version of node-gyp rather than the one it ships with,which may end up being newer.
```
rvagg, line 36:
```suggestion
**Beware** if you don't unset the `node_gyp` config option, npm will continue to use the globally installed version of node-gyp rather than the one it ships with, which may end up being newer.
```

rvagg (changes requested):

docs/Updating-npm-bundled-node-gyp.md
owl-from-hogvarts: On windows global `node_modules` and *node's* `node_modules` are two diffrent dirs, aren't they? Thats why 
> Generally, npm's library files are installed inside your global "node_modules", where npm is installed

can really mislead.

I propose this:
> So npm's internal copy of node-gyp isn't stored inside *global* `node_modules` and thus isn't available for use as a standalone package. Npm uses it's internal copy of node-gyp to automatically build native addon when you are trying to it install via `npm install`

cclauss:
Conflicting files `docs/Updating-npm-bundled-node-gyp.md`

Are we convinced that the majority of users are going to remember to undo/redo when upgrading?

owl-from-hogvarts:
When updating npm or node-gyp?

As i understand `npm config` writes to `.npmrc` which i asume should persist when updating npm. And also does path to global `node-gyp` really breaks when updating?

owl-from-hogvarts:
Does my own addition to this file should be done as separate PR?

owl-from-hogvarts:
@rvagg @cclauss Can this finally be merged, pls?

rvagg (approved):

rvagg:
yes, it can! sorry, I didn't see that you'd updated based on the review, thanks @owl-from-hogvarts!

DeeDeeG:
Hi folks.

This advice is not applicable to npm 7 or 8 at the moment. npm authors inadvertently broke the `node_gyp` npm config functionality, starting in npm 7, and it has not been fixed since. Still broken in the latest npm 8.1.3 as I write this comment. npm maintainers have acknowledged the issue.

See these open issues I filed at the npm repo: https://github.com/npm/cli/issues?q=is%3Aissue+is%3Aopen+node-gyp+author%3ADeeDeeG

The advice in this file, as of this PR, boils down to setting that config var. So it won't work on npm 7 or npm 8.

> The catch is that npm won't use this version unless you tell it to, it'll keep on using the one you have installed. You need to instruct it to by setting the `node_gyp` config variable (which goes into your `~/.npmrc`). You do this by running the `npm config set` command as below. Then npm will use the command in the path you supply whenever it needs to build a native addon.

That advice is unfortunately not true on npm 7 and 8 at the moment.

DeeDeeG:
Also: The title and filename of this do not match the content anymore, as of this PR. This advice is more about "Having npm use an independent copy of node-gyp instead of its bundled copy". 

The title ("Updating the npm-bundled version of node-gyp") matches the previous content.

I would request that the previous content of this file be hosted somewhere in the docs... Preferably under this filename, to make the edit history follow-able. (In other words, please consider restoring the previous content of this file, and moving the new content to another file.)

I would humbly suggest to move the updated content to a new file, under an appropriate filename and title, with the understanding that it applies ONLY to npm 6 and lower, and saying so somewhere in this file, so users don't get surprised when it doesn't work in npm 7 or higher.

The method described as of this PR is a more durable, maybe less hacky fix for npm 6 users. So I do like this method. It's just a minor disappointment it doesn't work on newer npm right now.

PR #2449, opened 2021-07-02
Fix: findVisualStudio with pwsh constrained lang. mode

lucascloarec:
'Add-Type' cannot be used in powershell with constrained language mode. Language mode does not exist in PowerShell 2.0 engine. To force powershell to use 2.0 engine allows the usage of msvc 2017 or newer in environment with constrained language mode.
See : 
https://devblogs.microsoft.com/powershell/powershell-constrained-language-mode/
https://docs.microsoft.com/en-us/powershell/scripting/windows-powershell/starting-the-windows-powershell-2.0-engine?view=powershell-7.1

##### Checklist

- [ ] tests are included 

##### Description of change
I added the option '-Version 2' in the powershell command which retrieves the installation data of Visual Studio 2017 (or newer). This option allows you to ignore the constrained language mode since the user does not necessarily have control over this parameter in his environment. This language mode does not allow the use of 'Add-Type', but this powershell applet is currently required to run C # code.

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -129,6 +129,8 @@ VisualStudioFinder.prototype = {
       'WindowsPowerShell', 'v1.0', 'powershell.exe')
     var csFile = path.join(__dirname, 'Find-VisualStudio.cs')
     var psArgs = [
+      '-Version',
+      '2',
       '-ExecutionPolicy',
       'Unrestricted',
       '-NoProfile',


samuelmaddock (changes requested):
This solution also has problems when PS 2.0 isn't installed
```
C:\Users\smaddock>powershell.exe -Version 2
Encountered a problem reading the registry.  Cannot find registry key SOFTWARE\Microsoft\PowerShell\1\PowerShellEngine. The Windows PowerShell 2 engine is not installed on this computer.
```

lucascloarec:
> This solution also has problems when PS 2.0 isn't installed
> 
> ```
> C:\Users\smaddock>powershell.exe -Version 2
> Encountered a problem reading the registry.  Cannot find registry key SOFTWARE\Microsoft\PowerShell\1\PowerShellEngine. The Windows PowerShell 2 engine is not installed on this computer.
> ```

I modified the pull request so that the use of the '-Version 2' option is only done when necessary.
If the usage is needed, but version 2 is not installed, then the script will crash, it's not a regression since the initial command wouldn't have worked either.
If the use is not necessary because powershell's language mode is FullLanguage, then the '-Version 2' option is not added to the command.

jarig:
I found that Get-VSSetupInstance works well - https://github.com/microsoft/vssetup.powershell

Here is my gist, but it's very rough with few hard-codes there (to be generalized):
https://gist.github.com/jarig/0328a439765e68cd966d168677f6bb5a

Mapuppy09 (changes requested):
Best effort!

cclauss:
> Conflicting files: `lib/find-visualstudio.js`

Mapuppy09:
Ok fix not used visual studio . Thanks !

Mayra delapaz

bshah007:
Can anyone fix this issue as it has a huge impact while building other libraries?

jarig:
@bshah007 
If constrained language mode is an issue, they try installing `Install-Module VSSetup -Scope CurrentUser`, then it will start using VSSetup module that won't hit constrained mode limitations. These instructions mentioned in the README and also should be pasted to the stdout if the module not found.

The change comes from the ticket mentioned above #2957 (merged)

PR #2512, opened 2021-10-05
feat: adds environment variable to allow PowerShell executable to be specified

brianpursley:
##### Checklist

##### Description of change
The `Add-Type` powershell command in `findVisualStudio2017OrNewer()` can fail with early versions of PowerShell (<=5.x) in a restricted corporate environment due to permission denied on a temp folder in `~\AppData\Local\Temp`.  This is not a problem in node-gyp itself, but rather a limitation of legacy versions of PowerShell in a restricted environment.

Newer versions of PowerShell (7.0 for example) do not have this problem and are able to perform the `Add-Type` command without any problems in a restricted environment.

This PR adds a new check for a new environment variable (`NODE_GYP_POWERSHELL_PATH`) which can be used to specify a different PowerShell executable to use, if you want.

Example usage:
```
$env:NODE_GYP_POWERSHELL_PATH = "C:\Program Files\PowerShell\7\pwsh.exe"
npm i
```

---

NOTE: This is different than the problem addressed by https://github.com/nodejs/node-gyp/pull/2449.  I'm looking for a way to use Powershell 7, which is an entirely different executable than powershell.exe.

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -127,6 +127,10 @@ VisualStudioFinder.prototype = {
   findVisualStudio2017OrNewer: function findVisualStudio2017OrNewer (cb) {
     var ps = path.join(process.env.SystemRoot, 'System32',
       'WindowsPowerShell', 'v1.0', 'powershell.exe')
+    if (process.env.NODE_GYP_POWERSHELL_PATH) {
+      ps = path.resolve(process.env.NODE_GYP_POWERSHELL_PATH)
+      this.addLog('PowerShell path has been set by NODE_GYP_POWERSHELL_PATH:', ps)
+    }
     var csFile = path.join(__dirname, 'Find-VisualStudio.cs')
     var psArgs = [
       '-ExecutionPolicy',


cclauss:
Are there new security exposures as a result of the proposed changes?

brianpursley:
> Are there new security exposures as a result of the proposed changes?

I don’t think there are any *new* security exposures, as this is similar to [how node-gyp uses the `PYTHON` environment variable](https://github.com/nodejs/node-gyp/blob/0a67dcd1307f3560495219253241eafcbf4e2a69/lib/find-python.js#L116).

It does allow someone to potentially put any arbitrary executable’s path in this variable and node-gyp will execute it while searching for visual studio.  I don’t think it would be an executable they can’t already execute themselves though.

Would you consider this to be a vulnerability?

As an alternative, I could change the new environment variable to something like `NODE_GYP_POWERSHELL_VERSION` and let you set it to something like `6` or `7` instead of the full path, and then the code would only look for the executable in a well known, whitelisted, location.  A little bigger change, but it would be more restrictive.

Another alternative might be to not add any new environment variable and change node-gyp to search for powershell like it does for Python, prefer a newer version of powershell if it is found, and fallback to the current powershell.exe if nothing is found.  I don’t see a problem with this approach personally, but I was hesitant to suggest this approach as it would change the default behavior of node-gyp.

brianpursley:
The more I think about it, the more I wonder if this should just be some sort of automatic fallback instead.

I originally thought the environment variable would be a good idea because it would let people decide if they want to use a different version of powershell, but node-gyp already has quite a few environment variables and it does make things more complicated for people.

What do you think about something like this?

1. Try to execute the command using `powershell.exe` like it does currently
2. If `powershell.exe` fails, then automatically try to execute the same thing using `pwsh.exe`
3. If both fail, then print the message, like it does today when `powershell.exe` fails.

PR #2547, opened 2021-11-08
fix: _ in npm_config_ env variables

zcbenz:
##### Checklist

- [x] tests are included 

##### Description of change

This PR fixes 2 things on `npm_config_` env variables:

1. Make `npm_config_force_process_config=true` env work by automatically converting `_` to `-` when parsing `npmConfigPrefix`.
2. Recognize `--disturl` as option in `create-config-gypi.js`, it is not documented anywhere but users have been relying on `npm_config_disturl` to pass `--dist-url` in env.

### DIFF
--- a/lib/create-config-gypi.js
+++ b/lib/create-config-gypi.js
@@ -19,7 +19,8 @@ async function getBaseConfigGypi ({ gyp, nodeDir }) {
   // try reading $nodeDir/include/node/config.gypi first when:
   // 1. --dist-url or --nodedir is specified
   // 2. and --force-process-config is not specified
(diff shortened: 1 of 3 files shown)


rvagg:
> users have been relying on npm_config_disturl

ugh, searching the web confirms lots of advice on this, and we do this in process-release.js already

Re "converting `_` to `-`" - I think the reason we've not had to deal with this previously is because we don't touch any environment variables that have `_` in them (and maybe why we're accepting `disturl`). So this seems fine, it's just surprising that we've not had to deal with this before.

rvagg (approved):

Issue #65, opened 2012-05-16
Build failure when builddir contains spaces.


jasondavies:
See also: brianmcd/contextify#30.

For example, a home directory of `Brad Old` causes the build to fail:

```
> contextify@0.1.2 install /Users/Brad Old/Desktop/DEMO_jason/node_modules/d3/node_modules/jsdom/node_modules/contextify
> node-gyp rebuild

info it worked if it ends with ok 
info downloading: http://nodejs.org/dist/v0.6.17/node-v0.6.17.tar.gz 
spawn python [ '/Users/Brad Old/.node-gyp/0.6.17/tools/gyp_addon',
  'binding.gyp',
  '-I/Users/Brad Old/Desktop/DEMO_jason/node_modules/d3/node_modules/jsdom/node_modules/contextify/build/config.gypi',
  '-f',
  'make' ]
spawn make [ 'BUILDTYPE=Release', '-C', 'build' ]
  CXX(target) Release/obj.target/contextify/src/contextify.o
clang: error: no such file or directory: 'Old/.node-gyp/0.6.17/src'
clang: error: no such file or directory: 'Old/.node-gyp/0.6.17/deps/uv/include'
clang: error: no such file or directory: 'Old/.node-gyp/0.6.17/deps/v8/include'
make: *** [Release/obj.target/contextify/src/contextify.o] Error 1
ERR! Error: `make` failed with exit code: 2
    at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:209:23)
    at ChildProcess.emit (events.js:70:17)
    at maybeExit (child_process.js:362:16)
    at Process.onexit (child_process.js:398:5)
ERR! not ok
npm WARN optional dependency failed, continuing contextify@0.1.2
```

Notice the `-I` parameter is not escaped.

Moter8:
Bump I guess? This still is an issue afaik.

Mike-Dunton:
I am still having this issue here is an log output. 
```
lwip@0.0.5 install /Users/michael dunton/projects/optimlol/node_modules/css-sprite/node_modules/lwip
> node-gyp rebuild

  CXX(target) Release/obj.target/lwip_decoder/src/decoder/init.o
clang: error: no such file or directory: 'dunton/.node-gyp/0.10.36/src'
clang: error: no such file or directory: 'dunton/.node-gyp/0.10.36/deps/uv/include'
[...]

PR #2556, opened 2021-11-25
fix: failure when build directory contains spaces (#65)

davej:
##### Checklist

- [ ] tests are included 

##### Description of change
Before this fix, if the build directory had a space in the path then it would fail because spaces were not escaped properly. Happy to add a test for this if anyone has a suggestion for how I should test this? Had a brief look through the existing tests and it wasn't obvious to me.

### DIFF
--- a/gyp/pylib/gyp/generator/make.py
+++ b/gyp/pylib/gyp/generator/make.py
@@ -1328,7 +1328,7 @@ $(obj).$(TOOLSET)/$(TARGET)/%%.o: $(obj)/%%%s FORCE_DO_CMD
             includes = config.get("include_dirs")
             if includes:
                 includes = [Sourceify(self.Absolutify(i)) for i in includes]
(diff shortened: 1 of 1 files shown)


cclauss:
I would strongly suggest the creation of a GitHub Action that fails because of a space in the pathname.  Then we can demonstrate that this code fixes the failing test.

gjsjohnmurray:
Getting this merged and into a new release should fix an issue seen on macOS in https://github.com/mtxr/vscode-sqltools/issues/921 where the [sqlite3](https://www.npmjs.com/package/sqlite3) package gets downloaded into a subfolder of ~/Library/Application Support/ and then has to build its binaries, uses node-gyp to do this, and fails because of the space in the path.

recrsn:
@davej The node-gyp project simply vendors the gyp-next so maybe we should raise a PR against https://github.com/nodejs/gyp-next/blob/main/pylib/gyp/generator/make.py

PR #2153, opened 2020-06-07
deps: increase "engines" to "node" : ">= 10.12.0"

DeeDeeG:
Makes npm warn users if they are using an unsupported Node version.

Resolves #2152.

##### Checklist

- [x] `npm install && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/nodejs/node/blob/master/doc/guides/contributing/pull-requests.md#commit-message-guidelines)

##### Description of change

Updates the [`"engines"` field of `package.json`](https://docs.npmjs.com/files/package.json#engines) to `"Node" : ">= 10.12.0"`. This is done to reflect a change in #2123, which requires Node v10.12.0 or above.

PR #2601, opened 2022-01-27
deps: increase "engines" to "node" : "^12.22 || ^14.13 || >=16"

wraithgar:
Makes npm warn users if they are using an unsupported Node version.

##### Checklist

##### Description of change
Commit message follows the pattern from the last time this field was updated: https://github.com/nodejs/node-gyp/pull/2153

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "which": "^2.0.2"
   },
   "engines": {
-    "node": ">= 10.12.0"
+    "node": "^12.22 || ^14.13 || >=16"
   },
   "devDependencies": {
     "bindings": "^1.5.0",


richardlau (approved):

richardlau:
I think we usually treat dropping support for a version (in this case Node.js 10) as semver-major.

wraithgar:
That was my first instinct but then I checked the git history on that line and that was not done last time.  In the cli we are treating this kind of change as a semver major, and would happily amend my commit.

The current testing matrix only includes 12.x, 14.x and 16.x which I believe only uses the latest versions in those lines.

richardlau:
I don't feel strongly enough about this to argue either way so I'll remove the label -- this specific change is is advisory (at least for npm, I think some (all?) versions of yarn enforced engines by default). 

Given that we don't actively test against Node.js 10 we can't even say that we aren't already broken there (albeit I'm sure we'd have received plenty of reports if that been the case).

If other node-gyp maintainers disagree, feel free to put back the semver-major label.

rvagg:
let's do semver-major `!`, it's not a big deal to bump that, releases are relatively cheap and signalling is important

@wraithgar do you have any urgency about getting a release for this out for consumption upstream?

wraithgar:
@rvagg I'd say medium low urgency? It's mostly about allowing node-gyp to stay current w/ the latest make-fetch-happen. v10 has a couple of bug fixes in it, nothing very big (yet). It would also shrink the npm cli bundle size a bit by allowing us to dedupe make-fetch-happen up out of node-gyp.

PR #2616, opened 2022-02-17
removing breaking space

icculp:
space between -- and npm_config_global cause the following error:
```
'npm_config_global' is not recognized as an internal or external command,
```

##### Checklist

- [ ] tests are included 

### DIFF
--- a/docs/Updating-npm-bundled-node-gyp.md
+++ b/docs/Updating-npm-bundled-node-gyp.md
@@ -31,12 +31,12 @@ Unix is easy. Just run the following command.
 
 If your npm is version ___7 or 8___, do:
 ```bash
-$ npm explore npm/node_modules/@npmcli/run-script -g -- npm_config_global=false npm install node-gyp@latest
+$ npm explore npm/node_modules/@npmcli/run-script -g --npm_config_global=false npm install node-gyp@latest
 ```
 
 Else if your npm is version ___less than 7___, do:
 ```bash
-$ npm explore npm/node_modules/npm-lifecycle -g -- npm install node-gyp@latest
+$ npm explore npm/node_modules/npm-lifecycle -g --npm install node-gyp@latest
 ```
 
 If the command fails with a permissions error, please try `sudo` and then the command.


cclauss:
@owl-from-hogvarts @DeeDeeG Your reviews, please.

owl-from-hogvarts (changes requested):
@icculp what is your OS and shell? For `fish` and `bash` on Linux unmodified command works well. I can't approve your changes since they modify semantics of the original command. Let me explain.

`npm_config_global` is environment variable. We pass it to `npm install` command to not install specified dependencies into global `node_modules` (because we want install it as dependency of particular package). The [npm-explore](https://docs.npmjs.com/cli/v8/commands/npm-explore) (look carefully at command syntax on top of the page) docs says that there *should be* space between command and `--`. This is common way to pass entire commands in unix world. Because if there is no space, `npm_config_global` considered not as a part of command to pass to subshell, but rather and option for `npm explore` itself. Thus `npm install` will not cancel inherited `-g` flag and will install `node-gyp` as *global* module which is not what we want. And since `npm explore` does not take any options so this one just ignored. 

My thought is, if env var cause any problems with shell we may just use `--no-global` flag which does exactly same thing but will not cause any problems with shell syntax.

DeeDeeG:
Hi, thanks for the ping. I have to say the commands modified in this PR _should_ work on Unix-like OSes. Linux and macOS, mainly.

For what it's worth, this looks like Windows cmd or Powershell to me:

> `XYZ is not recognized as an internal or external command`

Windows' cmd and powershell do not understand the "`command -- positional arguments here`" syntax.

The commands in this section of the document are only meant to work on Unix-like OSes (Linux or macOS mainly). There is a separate section of the instructions for Windows elsewhere in the document, which should work okay on Windows. I suggest trying those as written currently.

Furthermore, I agree with all of what @owl-from-hogvarts said above.

---

To add to that...

In plain English, the command updated by this PR is meant to 
- Go into the global install dir (`npm explore` with `-g` flag)
- From there go into the subdirectory `npm/node_modules/@npmcli/run-script` (npm 7 or 8) or `npm/node_modules/npm-lifecycle` (npm 6 or older)
- Set the environment variable `npm_config_global=false` (npm 7 or 8)
- Finally, run the command after the `--`, namely: `npm install node-gyp@latest`

So in summary, we are temporarily moving into the internal dependencies of the globally installed `npm` and updating its copy of `node-gyp`.

The "`command -- positional arguments here`" syntax convention is sometimes used in Unix-style command-line tools to have a main command `x` and a `--` and then some potentially complex or ambiguous argument `y` after. The `--` allows to clarify that whatever follows is a "positional argument", where being after the `--` tells you the meaning.

As @owl-from-hogvarts mentioned above, `npm explore` reserves a special meaning for anything after the `--`. Specifically: **For `npm explore`, anything after the `--` is a command to run after moving to the specified directory.**

Changing it to `--something` makes what follows a flag rather than a positional argument. So it would instead tack a flag on to `npm explore`. This would probably not be a flag `npm explore` is programmed to recognize or respond to, so the flag would likely be ignored... And the `npm install node-gyp@latest` would _not_ be interpreted as a separate command to run.

So this PR would remove the crucial step of actually installing newer `node-gyp`.

Again I suggest trying the steps in the Windows section of the document and seeing if those work, if you are indeed on Windows.

lukekarrys:
If this continues to be an issue, please try following the guidance above and if that does not work, open a new issue.

PR #2617, opened 2022-02-18
Update README.md by adding missing formatings

0xdw:
##### Checklist

- [ ] tests are included 

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,13 +94,13 @@ different build configurations from the official Node.js distribution, you
 should use `--dist-url` or `--nodedir` flags to specify the headers of the
 runtime to build for.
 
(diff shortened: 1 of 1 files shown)


rvagg:
So .. while this does increase the consistency, I personally dislike that we have `node-gyp` throughout the readme where it's just referring to the name rather than showing CLI or programmatic usage. I think we've gone back and forth on this question but my vote is to rip them all out.

Not a hard blocker if someone else feels strongly about this.

Thanks for the thoughtful contribution though @0xdw.

0xdw:
@rvagg Agreed! What I thought is that it's better to have a single way of stating the name. Also, I prefer to have the module name highlighted :).

PR #2647, opened 2022-04-21
chore: update `glob` package

Semigradsky:
It's just dependency updating.

There are no breaking changes, `glob@8` just doesn't support `node < 10` more: https://github.com/isaacs/node-glob/blob/main/changelog.md#80

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "main": "./lib/node-gyp.js",
   "dependencies": {
     "env-paths": "^2.2.0",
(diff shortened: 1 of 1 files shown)


richardlau:
We probably need to be careful with this one as there are reports of breakages on Windows with glob@8.

Refs: https://github.com/isaacs/node-glob/issues/467
Refs: https://github.com/nodejs/nodejs.org/pull/4551#issuecomment-1101829886

Issue #2645, opened 2022-04-14
Some tests fail because certificate expired


targos:
See https://github.com/nodejs/gyp-next/runs/6006035910?check_suite_focus=true

```
    # Subtest: download over https with custom ca
        1..3
        ok 1 - should be equal
        not ok 2 - request to https://localhost:49373/ failed, reason: certificate has expired
          ---
          stack: |
            ClientRequest.<anonymous> (node_modules/minipass-fetch/lib/index.js:130:14)
          at:
            line: 130
            column: 14
            file: node_modules/minipass-fetch/lib/index.js
            function: ClientRequest.<anonymous>
          type: system
          code: CERT_HAS_EXPIRED
          errno: CERT_HAS_EXPIRED
          tapCaught: returnedPromiseRejection
          test: download over https with custom ca
          source: |
            reject(new FetchError(`request to ${request.url} failed, reason: ${
          ...
```

cclauss:
Would [`github-token: ${{ secrets.GITHUB_TOKEN }}`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) work or do we need something different?

gengjiawen:
Looks like ca.crt expired.
```
gitpod /workspace/node-gyp/test/fixtures (master) $ cat ca.crt | openssl x509 -noout -enddate
notAfter=Apr 11 06:22:33 2022 GMT
```

gengjiawen:
> Looks like ca.crt expired.
> 
> ```
> gitpod /workspace/node-gyp/test/fixtures (master) $ cat ca.crt | openssl x509 -noout -enddate
> notAfter=Apr 11 06:22:33 2022 GMT
> ```

@bnoordhuis Any idea how to regenerate this CA. I am not familiar with openssl.

rvagg:
https://nodejs.org/en/knowledge/HTTP/servers/how-to-create-a-HTTPS-server/

PR #2662, opened 2022-05-11
fix: new ca & server certs, bundle in .js file and unpack for testing

rvagg:
New certs, expiry of 9999 days which should give us some breathing room eh? Hopefully node-gyp won't exist by then.

Bundling in certs.js rather than including the raw files should avoid some false positives that low-quality security scanners keep on complaining about .. I hope, but for all I know they're also looking for `--- BEGIN .*` in any file and might complain about that.

fix https://github.com/nodejs/node-gyp/issues/2645


gengjiawen (approved):

gengjiawen:
Not sure why this start to fail 

```console
# Subtest: test/test-addon.js
    # Subtest: build simple addon
        1..3
        # test count(0) != plan(3)
        # failed 1 test
    1..0 # no tests found
    1..0 # no tests found
not ok 1 - test/test-addon.js # time=1[20](https://github.com/nodejs/node-gyp/runs/6381965496?check_suite_focus=true#step:6:20)007.527ms
```

rvagg:
I doubled the timeout and still got ~half of the windows machines failing, so I bumped it up to 10 minutes and now I have all greens

that's pretty garbage, maybe GitHub's runners are slower? But it's green at least so I'm calling this good.

PR #2650, opened 2022-04-25
feat: add support for native windows arm64 build tools

dennisameling:
##### Checklist

- [x] `npm install && npm test` passes
- [x] tests are included 
- [x] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change

Visual Studio 2022 17.4 adds a native C++ compiler for Windows on ARM. This allows arm64 devices to leverage native build tools, leading to a 35% (or more) speed increase.

Related announcement from Microsoft: https://devblogs.microsoft.com/visualstudio/arm64-visual-studio-is-officially-here/

Building [Signal Desktop](https://github.com/signalapp/signal-desktop)'s native dependencies on a Surface Pro X (arm64 device):

```
x86 build tools (emulation): Done in 108.97s.
arm64 build tools (native): Done in 69.91s.
That's ~35.8% faster with the native tools.
```

PR #2680, opened 2022-06-11
fix: Find and use ARM64 tools on win32_arm64 platform

edolnx:
##### Checklist

- [x] tests are included 

##### Description of change
This fixes building on win32_arm64 platform by finding the correct VC tools for the platform. The documentation was already correct in terms of necessary tools to install, but the code would not find/use the ARM64 version of the tools, or worse build x86 code instead. This corrects that by looking for the ARM64 tools when the platform is arm64, otherwise default to the x86.x84 tools. This also allows for future platforms to be added in the if tree later if needed. All of the existing tests apply, and the log messages now show which version of the tools it's looking for in the event of a failure.

This fix works with VS2019 and VS2022. ARM64 support is not available on older versions of VS. This supersedes #2650 since VS2022 has shipped and the pre-release version is no longer needed.

Happy to make any adjustments in order to get this, or a similar patch, merged so that more NodeJS packages are available on the win32_arm64 platform.

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -288,11 +288,18 @@ VisualStudioFinder.prototype = {
 
   // Helper - process toolset information
   getToolset: function getToolset (info, versionYear) {
-    const pkg = 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64'
+    var pkg
+    if (process.arch === 'arm64') {
+      this.addLog('- Looking for ARM64 Tools')
+      pkg = 'Microsoft.VisualStudio.Component.VC.Tools.ARM64'
+    } else {
+      this.addLog('- Looking for x86.x64 Tools')
+      pkg = 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64'
+    }
     const express = 'Microsoft.VisualStudio.WDExpress'
 
     if (info.packages.indexOf(pkg) !== -1) {
-      this.log.silly('- found VC.Tools.x86.x64')
+      this.log.silly('- found VC.Tools')
     } else if (info.packages.indexOf(express) !== -1) {
       this.log.silly('- found Visual Studio Express (looking for toolset)')
     } else {


edolnx:
I don't think these failures are due to my change, but I'm unsure. Could someone give me some guidance?

dennisameling:
@edolnx did you see https://github.com/nodejs/node-gyp/pull/2650? I think we're both trying to accomplish the same 😄

dennisameling:
I actually think we need a combination of this PR and https://github.com/nodejs/node-gyp/pull/2650. In my testing, the 32-bit version of MSBuild is still triggered because `node-gyp` [explicitly references](https://github.com/nodejs/node-gyp/blob/d7687d55666fa77928cce270b8991b8e819c5094/lib/find-visualstudio.js#L272) `MSBuild\Current\Bin\MSBuild.exe`, which is 32-bit/x86:

image

The native MSBuild for ARM64 is located in `MSBuild\Current\Bin\arm64\MSBuild.exe`, which I added to my PR. 

As far as I can see, your PR only touches the _toolset detection logic_, which only checks if the ARM64 tools are _installed_, but does not actually use them.

Were you able to trigger the native ARM64 MSBuild by using this PR only (can be checked in Task Manager when building a project)?

edolnx:
@dennisameling You are correct that the x86 MSBuild is used, since I'm only using the GA releases of VS2019 and VS2022 on my system ARM64 based system. I did not try the VS2022 Preview which has ARM64 native MSBuild, however even with the x86 MSBuild the binaries created are ARM64 native since I only have those compilers installed. Otherwise, it will default to the x86.x86 compilers, as you noted. The reason why I didn't go down the path of trying with VS2022 is that most other dependencies still require VS2019 so sticking with the existing toolset which doesn't have a native MSBuild seems to make more sense. Also while the VS2022 compilers and tools are installed, the build system seems to prefer the VS2019 versions, probably because that is what build the Python 3.11b1 ARM64 release I used for this (which I only used because while the Python 3.10 code will compile for ARM64, the installer will not build and backporting the updated installer code failed). I'm relatively new to the NodeJS/npm ecosystem so I'll dig in a bit more and see if I can merge in some changes and I'll upgrade my VS2022 from the GA release to the Preview release with ARM64 native tooling.

edolnx:
Ah! After reviewing your code a bit mode @dennisameling your PR is in better shape than mine. There are some things that need to be addressed (mainly that most code won't compile on VS2022 currently because of compiler changes) but I'll put comments and make suggestions based off yours instead.

PR #2718, opened 2022-08-11
fix: update example to reflect proper argument usage

itsknob:
The example had a space instead of an equals sign (=) where the table below specifies that the argument should utilize an equals sign (=) for it's key-value flags.

##### Checklist

- [ ] tests are included 

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ version `node-gyp` should use in one of the following ways:
 1. by setting the `--python` command-line option, e.g.:
 
 ``` bash
-node-gyp <command> --python /path/to/executable/python
+node-gyp <command> --python=/path/to/executable/python
 ```
 
 2. If `node-gyp` is called by way of `npm`, *and* you have multiple versions of


favelado-web (approved):

cclauss:
Does it work either way or does only one way work?  On Linux, macOS, and Windows?

itsknob:
Since I opened this a while ago, my memory is a little fuzzy. I think the best information I have for you is this screenshot. This is MacOS on an M1 Chip. After specifying the Python 2.7 path as an arg without an equals sign (=), it seems as though the node-gyp utilized the Python 3 installation. 

image

All other arguments in the docs used an equals sign. I'm not entirely sure that the cause for using the wrong python version was due to the lack of the equals sign, but this MR was more just to standardize the way the arguments are laid out in the documentation.

cclauss:
Node-gyp is using Python 3 because Python 2 died more than 1,000 days ago on 1/1/2020.

itsknob:
Legacy software still exists, I was installing node-gyp 3.8 for an application running node 12. While I understand that it's not supported, this isn't a bug report. It's a docs consistency update.

If this doesn't seem relevant or worth the effort, feel free to close this.

cclauss:
`node-gyp 3.8` loves legacy Python.

PR #2719, opened 2022-08-19
fix: node.js debugger adds stderr (but exit code is 0) -> shouldn't throw

FuPeiJiang:
Debugger attached.
Waiting for the debugger to disconnect...

##### Checklist

- [ ] tests are included 

### DIFF
--- a/gyp/pylib/gyp/input.py
+++ b/gyp/pylib/gyp/input.py
@@ -961,13 +961,13 @@ def ExpandVariables(input, phase, variables, build_file):
                     # Fix up command with platform specific workarounds.
                     contents = FixupPlatformCommand(contents)
                     try:
-                        p = subprocess.Popen(
+                        # stderr will be printed no matter what
+                        result = subprocess.run(
                             contents,
-                            shell=use_shell,
                             stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE,
-                            stdin=subprocess.PIPE,
+                            shell=use_shell,
                             cwd=build_file_dir,
+                            check=False
                         )
                     except Exception as e:
                         raise GypError(
@@ -975,20 +975,12 @@ def ExpandVariables(input, phase, variables, build_file):
                             % (e, contents, build_file)
                         )
 
-                    p_stdout, p_stderr = p.communicate("")
-                    p_stdout = p_stdout.decode("utf-8")
-                    p_stderr = p_stderr.decode("utf-8")
-
-                    if p_stderr:
-                        sys.stderr.write(p_stderr)
-                    if p.wait() != 0:
-                        # Simulate check_call behavior, since check_call only exists
-                        # in python 2.5 and later.
+                    if result.returncode > 0:
                         raise GypError(
                             "Call to '%s' returned exit status %d while in %s."
-                            % (contents, p.returncode, build_file)
+                            % (contents, result.returncode, build_file)
                         )
-                    replacement = p_stdout.rstrip()
+                    replacement = result.stdout.decode("utf-8").rstrip()
 
                 cached_command_results[cache_key] = replacement
             else:
--- a/gyp/pylib/gyp/input.py
+++ b/gyp/pylib/gyp/input.py
@@ -979,8 +979,9 @@ def ExpandVariables(input, phase, variables, build_file):
                     p_stdout = p_stdout.decode("utf-8")
                     p_stderr = p_stderr.decode("utf-8")
 
-                    if p.wait() != 0 or p_stderr:
+                    if p_stderr:
                         sys.stderr.write(p_stderr)
+                    if p.wait() != 0:
                         # Simulate check_call behavior, since check_call only exists
                         # in python 2.5 and later.
                         raise GypError(


gyp/pylib/gyp/input.py
```diff
@@ -979,7 +979,7 @@ def ExpandVariables(input, phase, variables, build_file):
                     p_stdout = p_stdout.decode("utf-8")
                     p_stderr = p_stderr.decode("utf-8")
 
-                    if p.wait() != 0 or p_stderr:
+                    if p.wait() != 0:
                         sys.stderr.write(p_stderr)
                         # Simulate check_call behavior, since check_call only exists
                         # in python 2.5 and later.
```
cclauss, line 985:
```suggestion
                    if p_stderr:
                        sys.stderr.write(p_stderr)
                    if p.wait() != 0:
```

cclauss (changes requested):

gyp/pylib/gyp/input.py
cclauss: Is there a simple way that this code can detect the presence of the debugger?

cclauss: If we are going to make changes here, it is probably best to upgrade to [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html) to remove a lot of clutter.

FuPeiJiang: keep comments ? (keep/delete) ?
```
                        # Simulate check_call behavior, since check_call only exists
                        # in python 2.5 and later.
```

this is better behavior, no loss of stderr output (possible useful debug log)

cclauss: We do not need the comment about Py25...  https://docs.python.org/3/library/subprocess.html#subprocess.check_call is fully available to us on all supported versions of Python and the even simpler `subprocess.run()` is too.

Will _all_ users have the debugger present?  If not, what is the effect of the proposed changes on those users?

gyp/pylib/gyp/input.py
```diff
@@ -979,8 +979,9 @@ def ExpandVariables(input, phase, variables, build_file):
                     p_stdout = p_stdout.decode("utf-8")
                     p_stderr = p_stderr.decode("utf-8")
 
-                    if p.wait() != 0 or p_stderr:
+                    if p_stderr:
                         sys.stderr.write(p_stderr)
+                    if p.wait() != 0:
                         # Simulate check_call behavior, since check_call only exists
                         # in python 2.5 and later.
```
cclauss, line 986:
```suggestion
```
Remove.

gyp/pylib/gyp/input.py
FuPeiJiang: > what is the effect of the proposed changes on those users?

`<!(node -p \"require('node-addon-api').gyp\")`

oof, the scope is Any command..

___

> Is there a simple way that this code can detect the presence of the debugger?

there may be another way: silence the:
```
Debugger attached.
Waiting for the debugger to disconnect...
```

https://github.com/nodejs/node/blob/fddc701d3c0eb4520f2af570876cc987ae6b4ba2/src/inspector_agent.cc#L802-L805
It seems that it only prints if no parent
python spawning seems like a parent ? I'll have to check more

FuPeiJiang: Stderr doesn’t mean error,
Status code > 0 means error

> Yes, programs can output to standard error and still return a zero exit status. You will find many POSIX utilities specifications which state that "The standard error shall be used only for diagnostic messages."

https://unix.stackexchange.com/a/594670

this is the correct handling

If any program wants to indicate error, just set status code

FuPeiJiang:
the tests aren't even re-ran after force-push, this broke 2 tests

FuPeiJiang:
it's checking if the last line of stderr === 'gyp info ok'
but now it's === 'Waiting for the debugger to disconnect...'

FuPeiJiang:
nevermind, no breaking changes,
these 2 tests were always meant to fail in JavaScript Debug Terminal
and they were always meant to succeed in Not JavaScript Debug Terminal

FuPeiJiang:
I feel that checking the last line of something isn't a foolproof way of checking for success, maybe exit code ? I'll have to check the intention

gyp/pylib/gyp/input.py
```diff
@@ -961,33 +961,26 @@ def ExpandVariables(input, phase, variables, build_file):
...
-                        # Simulate check_call behavior, since check_call only exists
-                        # in python 2.5 and later.
+                    if result.returncode > 0:
                         raise GypError(
                             "Call to '%s' returned exit status %d while in %s."
-                            % (contents, p.returncode, build_file)
+                            % (contents, result.returncode, build_file)
                         )
-                    replacement = p_stdout.rstrip()
+                    replacement = result.stdout.decode("utf-8").rstrip()
```
cclauss, line 983: Add `text = True` to the `run()` call and then...
```suggestion
                    replacement = result.stdout.rstrip()
```

gyp/pylib/gyp/input.py
```diff
@@ -961,33 +961,26 @@ def ExpandVariables(input, phase, variables, build_file):
...
+                        # stderr will be printed no matter what
+                        result = subprocess.run(
                             contents,
-                            shell=use_shell,
                             stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE,
-                            stdin=subprocess.PIPE,
+                            shell=use_shell,
                             cwd=build_file_dir,
+                            check=False
```
cclauss, line 970:
```suggestion
                            check=False,
                            text=True
```

gyp/pylib/gyp/input.py
```diff
@@ -961,33 +961,26 @@ def ExpandVariables(input, phase, variables, build_file):
                     # Fix up command with platform specific workarounds.
                     contents = FixupPlatformCommand(contents)
                     try:
-                        p = subprocess.Popen(
+                        # stderr will be printed no matter what
+                        result = subprocess.run(
                             contents,
-                            shell=use_shell,
                             stdout=subprocess.PIPE,
```
cclauss, line 967:
```suggestion
```

FuPeiJiang: `stdout=subprocess.PIPE,` is needed for `result.stdout` to exist

cclauss: Only if `text = True` is absent?

FuPeiJiang: test this:

Fails:
```python
import subprocess
result = subprocess.run(
    'node -p "3245"',
    check=False,
    text=True
)
print(result.stdout) # None
```

Succeeds:
```python
import subprocess
result = subprocess.run(
    'node -p "3245"',
    stdout=subprocess.PIPE,
    check=False,
    text=True
)
print(result.stdout) # 3245
```

cclauss: Perfect.  Thanks for testing that out.  I think this is looking good.

FuPeiJiang:
**fails python 3.6 tests**
revert change `text=True`

> Changed in version 3.7: **Added the text parameter**, as a more understandable alias of universal_newlines. Added the capture_output parameter.

<https://docs.python.org/3/library/subprocess.html#:~:text=Changed%20in%20version%203.7%3A%20Added%20the%20text%20parameter%2C%20as%20a%20more%20understandable%20alias%20of%20universal_newlines.%20Added%20the%20capture_output%20parameter.>

cclauss (approved):
LGTM (Once all tests are passing ;-) )

cclauss:
Yes. Py36 is end-of-life https://devguide.python.org/versions so we should drop support for it but let's move this forward as is.

rvagg:
this shouldn't have been merged here but over in gyp-next: https://github.com/nodejs/gyp-next/blob/main/pylib/gyp/input.py

I believe it's going to get overwritten with a upgrade. @FuPeiJiang would you mind doing an equivalent PR over there please?

PR #2721, opened 2022-08-22
fix: create Python symlink only during builds, and clean it up after

### LINKED ISSUE #2713: `node_gyp_bins` causes build failures in downstream packages
* **Node Version**: Node 16.13.0 and NPM 8.1.0
* **Node-Gyp Version**: >= v9.0.0
* **Platform**: Mac & Linux
* **Module**: electron-packager & electron-rebuild

Hey folks! I'm a maintainer of `electron-packager` and `electron-rebuild`. We're seeing an issue with rebuilding native modules using node-gyp, that we think we introduced by this change: https://github.com/nodejs/node-gyp/commit/b9ddcd5bbd93b05b03674836b6ebdae2c2e74c8c . When trying to rebuild a native module using Electron and node-gyp 9.0.0+, rebuilding will fail for older versions of python or folks using a python version manager with this error:

```
An unhandled rejection has occurred inside Forge:
Error: /var/folders/sp/141453k53mg70282wr9vkn_h0000gp/T/electron-packager/darwin-x64/test-darwin-x64/Electron.app/Contents/Resources/app/node_modules/ffi-napi/build/node_gyp_bins/python3: file links out of the package
```

The link to the specific issue, with more details, is here: https://github.com/electron/electron-rebuild/issues/1024

We can handle this error inside electron-packager by ignoring the directory, but would it be possible to move this symlink directory outside of `node_modules` and into something like a tmp dir? I wasn't sure if the directory being in node_modules may cause other issues with other downstream users/packages, and thought we'd err on the side of reporting it 🙂 If there's a technical reason why it needs to be included in node_modules, please feel free to close this.

Thanks for your time!

pimterry:
##### Checklist

- [ ] tests are included 

Not really sure how to automatically test this fix, since the whole idea is that the symlink is leaves no trace, and I don't think there's any documentation required here.

##### Description of change

Fixes #2713

Previously in b9ddcd5bbd93b05b03674836b6ebdae2c2e74c8c this python symlink was created during configuration, and the symlink persisted indefinitely. This causes problems with many tools (including Electron Builder, Electron Forge, and Apple's `codesign` tool) that do not expect a codebase to ever include symlinks to external absolute paths.

This PR largely reverts that commit, and instead just writes the path to link to into the config, and then creates the symlink only temporarily during the build process, always deleting it afterwards.


apaprocki:
@pimterry @cclauss This is still causing problems for a lot of people. Is this fix going to be merged soon?

DeeDeeG (approved):
I happen to like this PR, since it resolves a user concern for a feature I wrote in https://github.com/nodejs/node-gyp/pull/2362.

The issue is apparently big enough to prevent people from publishing their app to app stores if `node-gyp` has been used to build code anywhere in their app.

The approach taken is one of the ones I suggested in my original PR.

<details>

With this PR, node-gyp records the validated python path (from `lib/find-python.js`) in the generated `config.gypi`, and reads that at build time to symlink to the validated python bin for the build. This means we can delete the symlink when the build finishes, rather than having to leave it to accommodate users who manually run `node-gyp configure` _then separately_ `node-gyp build`.

</details>

I'd like to see this merged, personally speaking.

lib/create-config-gypi.js
```diff
@@ -75,6 +75,9 @@ async function getCurrentConfigGypi ({ gyp, nodeDir, vsInfo }) {
   // set the node development directory
   variables.nodedir = nodeDir
 
+  // set the configured Python path
+  variables.python = python
```
DeeDeeG, line 79: My only concern is, if there are any existing packages which end up setting a variable `variables.python` in their `confi.gypi`?

Maybe this should be something more specific where you'd really have to try to collide with our namespace, like `variables.node_gyp.python_path` or `variables.node_gyp_python_path`?

(I'm admittedly not a heavy `node-gyp` or `gyp` user, so maybe someone would know if that's a realistic concern or not?)

jagthedrummer:
Any change of getting this merged?

nipunn1313 (approved):
looks good! Nice job w/ the node14 compat and isolating the symlinking behavior to lib/build.js (with appropriate cleanup).

If possible, a nicer solution would be to directly use the `config.variables.python` as needed rather than symlinks to the build bins dir - but that feels much larger scope than this fix. This fix feels like an improvement.

Regarding testing, maybe a few manual tests to show it works would mitigate concern - making sure it works on node 12/14/16/18 and using `find . -type l python3` to make sure nothing ends up there.

RCH010 (approved):

cclauss:
Please rebase to fix git conflicts.

pimterry:
> Please rebase to fix git conflicts.

@cclauss done - the build is just waiting approval

cclauss:
@StefanStojanovic @lukekarrys Your reviews, please.

pimterry:
The visual-studio error (a test after-all timeout) that's failing here seems ignorable too - that's been failing in other unrelated changes already, including this docs-only PR: https://github.com/nodejs/node-gyp/actions/runs/5412919694/jobs/10135897829

willm:
Also keen for this to get merged. FYI the python lint error that is failing has been fixed upstream, can this be rebased? Maybe it has more chance of getting merged if the status checks pass.

cclauss (approved):
RSLGTM

cclauss:
The failing Visual Studio tests are fixed elsewhere.

brendonjohn:
Thank you @cclauss 🙇

PR #2746, opened 2022-10-07
feat: remove support for VS2015 in Node.js >=19

targos:
Refs: https://github.com/nodejs/node/issues/37472

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -347,6 +347,11 @@ VisualStudioFinder.prototype = {
 
   // Find an installation of Visual Studio 2015 to use
   findVisualStudio2015: function findVisualStudio2015 (cb) {
+    if (this.nodeSemver.major >= 19) {
+      this.addLog(
+        'not looking for VS2015 as it is only supported up to Node.js 18')
+      return cb(null)
+    }
     return this.findOldVS({
       version: '14.0',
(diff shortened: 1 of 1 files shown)


targos:
/cc @nodejs/build @nodejs/npm @nodejs/releasers

We will need to coordinate an npm release with this change, its inclusion in v19.0.0, and adaptations to the Node.js CI.

targos:
I think this is something we should do regardless of issues with V8, to prepare users for Node.js 20. Visual Studio 2015 will be out of support on [Oct 14, 2025](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2015), before the 20.x EoL.

sxa (approved):

Trott (approved):

richardlau (approved):
LGTM but I wonder if the commit message should mark this as a breaking change (we're using conventional commits for release-please).

targos:
I thought about it but since it only applies to an unreleased version of node, I thought it could not really break anyone

targos:
But feel free to land as breaking, I have no strong opinion on that

richardlau:
Ah good point about it referring to an unreleased version. I'm okay with it landing just as a feat (i.e. Node.js 19 support).

gengjiawen (approved):

PR #2772, opened 2022-12-09
chore: skip test when python version is unsupported on platform

lukekarrys:
##### Checklist

##### Description of change

`python@3.6` is no longer supported on `ubuntu-latest` (which is now `22.04`). See this recent [workflow run](https://github.com/nodejs/node-gyp/actions/runs/3661197360/jobs/6189168960#step:4:12) to see the failure that occurs when the matrix tries to test that combination.

This change makes CI skip all subsequent steps when the `setup-python` step fails.

One downside to this is that the GitHub UI will still show ✅ for the workflow run instead of indicating that it was skipped.

##### Alternative options

Instead of this, `3.6` could be dropped from the CI matrix, but I don't know enough about python usage to have an opinion on if that is a good idea.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,28 +25,34 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - name: Use Python ${{ matrix.python }}
+        id: setup-python
+        continue-on-error: true
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
         env:
           PYTHON_VERSION: ${{ matrix.python }}  # Why do this?
       - name: Install Dependencies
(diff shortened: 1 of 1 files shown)


cclauss (changes requested):
Python 3.6 is not supported anywhere...  https://devguide.python.org/versions

My vote would be to ___drop Python 3.6___ and not make the changes in this PR. Allowing failure and then testing for success sounds dicey and it would be easy to have logic errors that hide important failures.

lukekarrys:
TIL about `matrix.exclude` which would be a much better idea than the `continue-on-error` strategy used here.

But if 3.6 is no longer supported than I agree the best idea would be to drop it from the test matrix.

lukekarrys:
Closing in favor of #2774

PR #2810, opened 2023-03-06
Fix: find visual studio arm64 support

StefanStojanovic:
##### Checklist

##### Description of change
This enables running find visual studio script on Windows on ARM64.

Find-VisualStudio.cs uses certain COM classes and interfaces to fetch information about installed Visual Studio versions. As it turns out, these components are not registered correctly for ARM64 when installed, thus they cannot be used from the ARM64 process. Because of that, when searching for Visual Studio installations on Windows on ARM64, PowerShell from SysWOW64 is used, since then Find-VisualStudio.cs can create instances of COM class.

This issue was noticed when running Node.js native test suites on Windows on ARM. The change proposed here fixes this issue as seen from this run https://ci.nodejs.org/job/node-test-binary-windows-native-suites/17121/

The same problem was previously reported in other repositories, eg. https://github.com/microsoft/vs-setup-samples/issues/15 and https://github.com/rust-lang/rust/issues/83043

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -126,7 +126,19 @@ VisualStudioFinder.prototype = {
   // Invoke the PowerShell script to get information about Visual Studio 2017
   // or newer installations
   findVisualStudio2017OrNewer: function findVisualStudio2017OrNewer (cb) {
-    var ps = path.join(process.env.SystemRoot, 'System32',
+    // 32-bit PowerShell (available in SysWOW64) is needed on ARM64 because of
+    // the COM classes and interfaces used by Find-VisualStudio.cs, which are
+    // not registered for being used from ARM64 processes.
+    //
+    // This fix is required for the clean installation of Windows 11. Old
+    // versions (e.g. Windows 10 upgraded to 11) work without it.
+    //
+    // Since VS v17.4 and later is natively supported on ARM64 it seems to
+    // register the ARM64 COM components correctly, but VS 2019 and 2017 will
+    // still have problems being detected. Once support for VS 2017 and 2019
+    // is dropped, this change can be reverted.
+    var systemDirectory = process.arch === 'arm64' ? 'SysWOW64' : 'System32'
+    var ps = path.join(process.env.SystemRoot, systemDirectory,
       'WindowsPowerShell', 'v1.0', 'powershell.exe')
     var csFile = path.join(__dirname, 'Find-VisualStudio.cs')
     var psArgs = [


anonrig (approved):

anonrig:
cc @nodejs/gyp

lib/find-visualstudio.js
```diff
@@ -126,7 +126,11 @@ VisualStudioFinder.prototype = {
   // Invoke the PowerShell script to get information about Visual Studio 2017
   // or newer installations
   findVisualStudio2017OrNewer: function findVisualStudio2017OrNewer (cb) {
-    var ps = path.join(process.env.SystemRoot, 'System32',
+    // SysWOW64 PowerShell is needed on ARM64 because of the COM classes and
+    // interfaces used by Find-VisualStudio.cs, which are not registered for
+    // being used from ARM64 processes.
+    var systemDirectory = process.arch === 'arm64' ? 'SysWOW64' : 'System32'
```
gengjiawen, line 132: It used to work for arm64. Should some compatibility need to do here ? Like `SysWOW64` available in all windows arm64 distribution.

gengjiawen: Also from what I tested, currently works

I am on windows virtual machine with latest visual studio
```console
PS C:\Users\jiawengeng\node-gyp\lib> node -p process.arch 
arm64
PS C:\Users\jiawengeng\node-gyp\lib> node .\find-visualstudio.js
info find VS using VS2022 (17.5.33424.131) found at:
info find VS "C:\Program Files\Microsoft Visual Studio\2022\Enterprise"
info find VS run with --verbose for detailed information
```

StefanStojanovic: > It used to work for arm64.

That makes sense to me. I ran the following command from bash on two ARM64 machines:
* Command:
```
file /c/Windows/*/WindowsPowerShell/v1.0/powershell.exe
```
* Machine 1: Originally installed Windows 10, was upgraded to Windows 11
```
/c/Windows/SysArm32/WindowsPowerShell/v1.0/powershell.exe: PE32 executable (console) Intel 80386, for MS Windows
/c/Windows/SysWOW64/WindowsPowerShell/v1.0/powershell.exe: PE32 executable (console) Intel 80386, for MS Windows
/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe: PE32 executable (console) Intel 80386, for MS Windows
```
* Machine 2: Originally installed Windows 11
```
/c/Windows/SysArm32/WindowsPowerShell/v1.0/powershell.exe: PE32 executable (console) ARMv7 Thumb, for MS Windows
/c/Windows/SysWOW64/WindowsPowerShell/v1.0/powershell.exe: PE32 executable (console) Intel 80386, for MS Windows
/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe: PE32+ executable (console) Aarch64, for MS Windows
```
Can you please run the same command and check what the output will be on your machine? Only `Intel 80386` should be able to find Visual Studio, so as expected the fix proposed in this PR is not required on machine 1 but it is for machine 2.

> Should some compatibility need to do here ? Like SysWOW64 available in all windows arm64 distribution.

After you do a clean install, SysWOW64 is present on Windows on ARM.

> Also from what I tested, currently works
> 
> I am on windows virtual machine with latest visual studio

This is in line with https://github.com/nodejs/node/pull/46995#issuecomment-1458285152. Since VS v17.4 and later is natively supported on ARM64 I assume it registers the ARM64 COM components correctly, but VS 2019 and 2017 will still have problems being detected.

anonrig: Can you add a comment to the code too? It would be beneficial in the long run to have this knowledge on the code as well.

StefanStojanovic: I've updated the comment in the code with the most important info from my previous comment.

StefanStojanovic: @gengjiawen is there something else I can do to help this move forward?

cclauss:
@nodejs/platform-windows-arm

lib/find-visualstudio.js
gengjiawen: I am Neutral on this since I am skeptical on vs support on arm in older versions based on my experience (Rarely fix bugs and lack support from official). I am more inclined drop VS2019 on arm instead patch on it. 

@rvagg Second Opinion.

PR #2822, opened 2023-04-01
doc: Update README.md

beliaev-maksim:
Update installation instruction

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Install tools and configuration manually:
    * Install Visual C++ Build Environment: [Visual Studio Build Tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools)
    (using "Visual C++ build tools" workload) or [Visual Studio Community](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community)
    (using the "Desktop development with C++" workload)
-   * Launch cmd, `npm config set msvs_version <VS Studio version>`, e.g. `npm config set msvs_version 2022`
+   * Launch cmd, `npm config set msvs_version 2022` (or other supported version year)
 
    If the above steps didn't work for you, please visit [Microsoft's Node.js Guidelines for Windows](https://github.com/Microsoft/nodejs-guidelines/blob/master/windows-environment.md#compiling-native-addon-modules) for additional tips.
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,6 @@ Install tools and configuration manually:
    * Install Visual C++ Build Environment: [Visual Studio Build Tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools)
    (using "Visual C++ build tools" workload) or [Visual Studio Community](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community)
    (using the "Desktop development with C++" workload)
-   * Launch cmd, `npm config set msvs_version 2022` (or other supported version year)
 
    If the above steps didn't work for you, please visit [Microsoft's Node.js Guidelines for Windows](https://github.com/Microsoft/nodejs-guidelines/blob/master/windows-environment.md#compiling-native-addon-modules) for additional tips.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ Install tools and configuration manually:
    * Install Visual C++ Build Environment: [Visual Studio Build Tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools)
    (using "Visual C++ build tools" workload) or [Visual Studio Community](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community)
    (using the "Desktop development with C++" workload)
-   * Launch cmd, `npm config set msvs_version 2017`
+   * Launch cmd, `npm config set msvs_version <VS Studio version>`, e.g. `npm config set msvs_version 2022`
```
cclauss, line 56:
```suggestion
   * Launch cmd, `npm config set msvs_version 2022` (or other supported version year)
```

README.md
```diff
@@ -53,7 +53,7 @@ Install tools and configuration manually:
    * Install Visual C++ Build Environment: [Visual Studio Build Tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools)
    (using "Visual C++ build tools" workload) or [Visual Studio Community](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community)
    (using the "Desktop development with C++" workload)
-   * Launch cmd, `npm config set msvs_version 2017`
```
gengjiawen, line 56: This is actually optional, I am inclined to delete this setup. @cclauss @rvagg Second opinion ?

cclauss: I am not a Windows user.

gengjiawen: @beliaev-maksim Can you delete this line ? I think for most user, this is not necessary.

beliaev-maksim: @gengjiawen @cclauss
Maybe we completely remove the part with visual studio?
To my knowledge new versions of Node come directly with additional developer tools (although, you have to check it while install it)

That should provide better integration
What do you think?

gengjiawen: > Maybe we completely remove the part with visual studio?

Not a good idea, many people use visual studio directly. They have theire own cpp config.

gengjiawen (approved):

gengjiawen:
thx for the contribution

karlhorky:
> @gengjiawen in [comment 1161506873](https://github.com/nodejs/node-gyp/pull/2822#discussion_r1161506873): Can you delete this line ? I think for most user, this is not necessary.

@gengjiawen some of our students with Windows machines have problems with running this step, so it's interesting to us that you removed it with the explanation that it's not necessary.

Their problems look like [this](https://github.com/upleveled/system-setup/issues/31):

```
PS C:\Windows\system32> npm config set msvs_version 2017
npm ERR! `msvs_version` is not a valid npm option

npm ERR! A complete log of this run can be found in: C:\Users\khayo\AppData\Local\npm-cache\_logs\2023-08-22T12_12_13_960Z-debug-0.log
```

Can you expand on the explanation why it's not necessary for most users? Which users would it be necessary for?

Our [Windows System Setup Guide](https://github.com/upleveled/system-setup/blob/main/windows.md#:~:text=choco%20install%20python%20visualstudio2017%2Dworkload%2Dvctools%20%2Dy) suggests to install the Chocolatey packages [`visualstudio2017-workload-vctools`](https://community.chocolatey.org/packages/visualstudio2017-workload-vctools) and [`python`](https://community.chocolatey.org/packages/python) and then configure `msvs_version`, which leads to this failure.

If you have any suggestions or remarks, we're happy to also do the setup a different way - but it should be able to be performed via the command line, no GUI tools or downloads from websites.

gengjiawen:
> Can you expand on the explanation why it's not necessary for most users? 

Gyp can automatically choose msvs version.

> Which users would it be necessary for?

You have a project only can compile in vs2017, but tbh MS don't maintain old visual studio.


My general advice: Install `visual studio` and python on windows is only needed for node-gyp.

karlhorky:
Ok great, then if I'm understanding correctly, @ProchaLu will try to upgrade to [`visualstudio2022-workload-vctools`](https://community.chocolatey.org/packages/visualstudio2022-workload-vctools) and remove the `npm config set msvs_version ...` command to see if this works, thanks!

- https://github.com/upleveled/system-setup/issues/31#issuecomment-1689952461

karlhorky:
Ok great, then if I'm understanding correctly, @ProchaLu will try to upgrade to [`visualstudio2022-workload-vctools`](https://community.chocolatey.org/packages/visualstudio2022-workload-vctools) and remove the `npm config set msvs_version ...` command to see if this works, thanks!

- https://github.com/upleveled/system-setup/issues/31#issuecomment-1689952461

PR #2837, opened 2023-05-04
Using mocha for running some tests

StefanStojanovic:
##### Checklist
- [x] tests are included 

##### Description of change
Constant GitHub Action failures started after landing https://github.com/nodejs/node-gyp/commit/02480f6b6894a6bb81dd3e1365bc4df6d5336f7c. One way to fix this is reverting that commit as proposed in https://github.com/nodejs/node-gyp/pull/2849. This PR uses a different approach by keeping the `make-fetch-happen` version and making changes in testing modules.

The first thing updated is the `tap` module, from its legacy version 12.7.0 to the latest version 16.3.4. This change fixed the failing tests, but afterward, `test-addon.js` started failing randomly in GitHub Actions. Rerunning failed jobs would sometimes fix that, but in some cases, mostly on Windows, they stood broken after many reruns. There was no consistency in this and no pattern was to be found. On the other hand, running the `npm test` locally always passed for me and I tested with various node/python combinations.

Since I couldn't reproduce any issues locally, I decided to try changing the testing framework (at least for the failing tests). That brings me to my second change - using `mocha` to run `test-addon.js` which had to be adapted to the new framework. After doing this everything passed in the GitHub Actions without any issues.

##### Other ideas
At this time, the solution isn't pretty but it unblocks the CI, which was my main goal. Potentially, all tests can be migrated to `mocha` eventually, thus `tap` can be removed completely then. Another idea (proposed in https://github.com/nodejs/node-gyp/pull/2851) is to simply use `mocha` to run tests we already have (eg. `mocha test/test-*`) instead of `tap`, and keep tests written in `tap`.


StefanStojanovic:
Closing in favor of https://github.com/nodejs/node-gyp/pull/2851

Issue #2848, opened 2023-05-17
subdependency regression breaks node-gyp 9


legobeat:
<details><summary>Verbose output (from npm or node-gyp):</summary>

[CI run on current `main`](https://github.com/nodejs/node-gyp/actions/runs/4811204563/jobs/8564881101#step:9:143):

```
# Subtest: test/test-create-config-gypi.js
    # Subtest: config.gypi with no options
        1..2
        ok 1 - should be equal
        ok 2 - should be equal
    ok 1 - config.gypi with no options # time=22.437ms
    
    # Subtest: config.gypi with --debug
        1..1
        ok 1 - should be equal
    ok 2 - config.gypi with --debug # time=3.625ms
    
    # Subtest: config.gypi with custom options
        1..1
        ok 1 - should be equal
    ok 3 - config.gypi with custom options # time=2.86ms
    
    # Subtest: config.gypi with nodedir
        1..1
        ok 1 - should be equal
    ok 4 - config.gypi with nodedir # time=6.656ms
    
    # Subtest: config.gypi with --force-process-config
        1..1
        ok 1 - should be equal
    ok 5 - config.gypi with --force-process-config # time=3.603ms
    
    # Subtest: config.gypi parsing
        1..1
        ok 1 - should be equivalent
    ok 6 - config.gypi parsing # time=1.559ms
    
    1..6
    # time=67.566ms
ok 3 - test/test-create-config-gypi.js # time=603.309ms

/home/runner/work/node-gyp/node-gyp/node_modules/path-scurry/node_modules/lru-cache/dist/cjs/index.js:52
            this.signal.onabort?.(reason);
                                ^

SyntaxError: Invalid or unexpected token
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
# Subtest: test/test-download.js
    1..0 # no tests found
not ok 4 - test/test-download.js # time=831.278ms
  ---
  timeout: 600000
  file: test/test-download.js
  childId: 3
  command: /opt/hostedtoolcache/node/14.21.3/x64/bin/node
  args:
    - '-r'
    - /home/runner/work/node-gyp/node-gyp/node_modules/esm/esm.js
    - test/test-download.js
  stdio:
    - 0
    - pipe
    - 2
  cwd: /home/runner/work/node-gyp/node-gyp
  exitCode: 1
  ...

```

</details>

The incompatible syntax got introduced in `lru-cache` update from 8 to 9:
 
[...]

PR #2827, opened 2023-04-08
feat!: update `engines.node` to `^14.17.0 || ^16.13.0 || >=18.0.0`

benjaminpjones:
Closes #2753 

##### Checklist

- [x] `npm install && npm test` passes [^1]
- [ ] ~~tests are included~~ (N/A) 
- [ ] ~~documentation is changed or added~~ (N/A)
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change

This PR drops support for node 12 and adds node 18.

The change was taken directly from #2770 (I haven't even modified the commit message [^2]), but I am breaking it out for convenience of review.  The updated engines was listed as a blocker for #2796, so the goal is to unblock that fix.

##### Footnotes

[^1]: One test was skipped: `test/test-addon.js ... python console app can't encode non-ascii character.`

[^2]: I noticed previous PRs tagged engine updates under "deps" rather than "feat", but since no one called it out in the original PR, I didn't change it.

PR #2796, opened 2023-02-10
update make-fetch-happen to 11.0.3

ravindraP20:
Closes #2795
Blocked by #2770

http-cache-semantics 4.1.0 is vulnerable

https://www.cve.org/CVERecord?id=CVE-2022-25881

##### Checklist

- [x] `npm install && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change
http-cache-semantics 4.1.0 is vulnerable

https://www.cve.org/CVERecord?id=CVE-2022-25881

https://github.com/npm/make-fetch-happen/pull/210

PR #2770, opened 2022-12-09
Sync deps and engines with npm

lukekarrys:
##### Checklist

- [x] `npm install && npm test` passes
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change

The main purpose of this PR is to update dependencies in order to dedupe as many deps as possible within `npm`. As can be seen [in the `npm/cli` source](https://github.com/npm/cli/tree/7018b3d46e10ea4d9d81a478dbdf114b6505ed36/node_modules/node-gyp/node_modules) there are a number of dependencies that end up nested within `node-gyp/node_modules` due to conflicting versions.

To accomplish this a number of other changes were made:

- [`fe5b13a`](https://github.com/nodejs/node-gyp/pull/2770/commits/fe5b13a) **BREAKING CHANGE** update `engines.node` to `^14.17.0 || ^16.13.0 || >=18.0.0` 
- [`a729c15`](https://github.com/nodejs/node-gyp/pull/2770/commits/a729c15) Update `standard` to work with newer JS syntax and fix new lint errors
- [`9ccc66d`](https://github.com/nodejs/node-gyp/pull/2770/commits/9ccc66d) The @npm/cli-team team will be deprecating `npmlog` soon, so this PR drops that dependency in favor of `proc-log` + an otherwise dependency-less custom logger that includes the subset of functionality from `npmlog` used by `node-gyp`
- [`d485f1b`](https://github.com/nodejs/node-gyp/pull/2770/commits/85b642b) Promisifies the `build` command since `which@3.0.0` is promise only
- [`02fc75d`](https://github.com/nodejs/node-gyp/pull/2770/commits/e33ee18) Promisifies the `remove` and `clean` commands after dropping `rimraf` in favor of native recursive rm with `fs.promises.rm`

##### Commits

Each commit in the PR is a separate unit of work and I'd be happy to break it up if this is too difficult to review as is.

##### Reference issues/PRs

After doing the above, I looked through the list of open issues and PRs and identified the following that would either be closed or superseded by this PR.

Closes #2753
Closes #2647
Closes #2358
Closes #2357
Closes #2756
Closes #2379
Closes #2225

PR #2849, opened 2023-05-17
Revert "update make-fetch-happen to 11.0.3 (#2796)"

legobeat:
This reverts commit 02480f6b6894a6bb81dd3e1365bc4df6d5336f7c, as it introduced breaking changes blocking a release of node-gyp 9 from a clean `main`.

`make-fetch-happen` would have to be updated to a (hypothetical future) v10.x fix, or replaced with something else.

##### Checklist

- [x] tests are included 

##### Description of change

This reverts commit 02480f6b6894a6bb81dd3e1365bc4df6d5336f7c, thereby rolling back dependency `make-fetch-happen` from `^11.0.3` to `^10.0.3`.

The upgrade is breaking for node-fetch users as it has transitive
dependencies with syntax incompatible with supported Node.js versions.

Related:
- https://github.com/nodejs/node-gyp/pull/2770
- https://github.com/nodejs/node-gyp/pull/2837
- https://github.com/nodejs/node-gyp/pull/2816
- https://github.com/nodejs/node-gyp/issues/2848
- https://github.com/nodejs/node-gyp/pull/2827
- https://github.com/nodejs/node-gyp/pull/2796

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "exponential-backoff": "^3.1.1",
     "glob": "^8.0.3",
     "graceful-fs": "^4.2.6",
-    "make-fetch-happen": "^11.0.3",
+    "make-fetch-happen": "^10.0.3",
     "nopt": "^7.0.0",
     "proc-log": "^3.0.0",
     "semver": "^7.3.5",


rvagg:
This seems reasonable given our "engines" claim (which is different to our CI versions, which isn't great). But doesn't this take us back to a version with a security vulnerability tag?

Comments @benjaminpjones, @lukekarrys, @cclauss?

rvagg:
A mild dilemma, but IMO we should ditch `engines` entirely (in a semver-major probably), I'm not sure why we're still doing that other than to telegraph our contract, which we're not exactly sticking to! It might have been to sync with npm originally. I'd rather we ditch it and test our contract in CI than in package.json.

I think `^12.18` would be ok to switch, then we get to pull in the `@nodejs/undici` team whenever we have a problem as a bonus.

lukekarrys:
One more point on `engines`:

Anything wider than what `npm/cli` implements (eg `^12.18`) would allow `npm@9` to update to that upcoming version of `node-gyp`.

However as #2770 shows `node-gyp` uses a number of `npm` dependencies and we've implemented the same `enginges` range across all our supported packages. So landing #2770 would be problematic under that new `engines` range.

lukekarrys:
Now that #2770 has landed, we should figure out how to have multiple release lines. I plan on making a PR setting up tooling similar to how the npm CLI does it which supports multiple release lines.

Once that is complete, this should land on v9 as 9.4.1 and v10 will be the breaking engines changes.

cclauss:
Please resolve the git conflict.

cclauss:
I am not qualified to review this.  Is there a Node.js contributor that can?

lukekarrys:
I just landed https://github.com/nodejs/node-gyp/pull/2917 on the `release/v9` branch, so I think this is safe to close for now.

Issue #2847, opened 2023-05-15
Random HTTP 500 statuses or ECONNRESET when node-gyp fetches dependencies


rchennafi:
This is occurring across multiple node versions, architectures and system types. Our build system regularly runs node gyp in its build for our NPM packages across every version of nodejs and about 5% of builds are subject to a 500 status failure or an ECONNRESET when setting up node-gyp.

This strikes me as an issue with the server which hosts the node gyp files. Is there possibly a way around downloading these node-gyp dependencies each time? Or maybe something that can be done to fix the instability there?

These builds all took place inside of Amazon server instances.

Here is an instance on node 14 with node gyp 7.1.2 on a Windows Server 2019 machine.
```
04:08:09  gyp info ok 
04:08:13  Now using node v14.21.2 (64-bit)
04:08:19  gyp info it worked if it ends with ok
04:08:19  gyp info using node-gyp@7.1.2
04:08:19  gyp info using node@14.21.2 | win32 | x64
04:08:19  gyp info find Python using Python version 3.8.5 found at "C:\Python38\python3.exe"
04:08:23  gyp http GET https://nodejs.org/download/release/v14.21.2/node-v14.21.2-headers.tar.gz
04:08:23  gyp http 200 https://nodejs.org/download/release/v14.21.2/node-v14.21.2-headers.tar.gz
04:08:25  gyp http GET https://nodejs.org/download/release/v14.21.2/SHASUMS256.txt
04:08:25  gyp http GET https://nodejs.org/download/release/v14.21.2/win-x86/node.lib
04:08:25  gyp http GET https://nodejs.org/download/release/v14.21.2/win-x64/node.lib
04:08:25  gyp http GET https://nodejs.org/download/release/v14.21.2/win-arm64/node.lib
04:08:25  gyp http 200 https://nodejs.org/download/release/v14.21.2/SHASUMS256.txt
04:08:26  gyp http 500 https://nodejs.org/download/release/v14.21.2/win-arm64/node.lib
04:08:26  gyp WARN install got an error, rolling back install
04:08:26  gyp ERR! configure error 
04:08:26  gyp ERR! stack Error: 500 status code downloading arm64 node.lib
04:08:26  gyp ERR! stack     at Request.<anonymous> (C:\nvm\v14.21.2\node_modules\node-gyp\lib\install.js:335:22)
04:08:26  gyp ERR! stack     at Request.emit (events.js:412:35)
[...]

PR #2857, opened 2023-05-26
win,install: only download target_arch node.lib

StefanStojanovic:
##### Checklist

- [x] tests are included

##### Description of change

Instead of downloading node.lib for all architectures, just download the one that will be needed. `install.js` changed to enable downloading just node.lib for node versions that already have tarball downloaded and extracted. Not fetching lib now fails the installation. Increased installVersion because of the changes.

A note to reviewers - although it seems there are many changes, most of them are just indentation. Also @dsanders11 I'd like to ask you for a review, since your change for `install.js` landed a few days ago, and I worked on those parts of the code too.

Refs: https://github.com/nodejs/node-gyp/issues/2847

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -22,6 +22,10 @@ const streamPipeline = util.promisify(stream.pipeline)
 
 async function install (fs, gyp, argv) {
   const release = processRelease(argv, gyp, process.version, process.release)
+  // Detecting target_arch based on logic from create-cnfig-gyp.js. Used on Windows only.
+  const arch = win ? (gyp.opts.target_arch || gyp.opts.arch || process.arch || 'ia32') : ''
+  // Used to prevent downloading tarball if only new node.lib is required on Windows.
+  let shouldDownloadTarball = true
 
   // Determine which node dev files version we are installing
   log.verbose('install', 'input version string %j', release.version)
@@ -92,6 +96,26 @@ async function install (fs, gyp, argv) {
       }
     }
     log.verbose('install', 'version is good')
+    if (win) {
+      log.verbose('on Windows; need to check node.lib')
+      const nodeLibPath = path.resolve(devDir, arch, 'node.lib')
+      try {
+        await fs.promises.stat(nodeLibPath)
+      } catch (err) {
+        if (err.code === 'ENOENT') {
+          log.verbose('install', `version not already installed for ${arch}, continuing with install`, release.version)
+          try {
+            shouldDownloadTarball = false
+            return await go()
+          } catch (err) {
+            return rollback(err)
+          }
+        } else if (err.code === 'EACCES') {
+          return eaccesFallback(err)
+        }
+        throw err
+      }
+    }
   } else {
     try {
       return await go()
@@ -179,6 +203,7 @@ async function install (fs, gyp, argv) {
     }
 
     // download the tarball and extract!
+    // Ommited on Windows if only new node.lib is required
 
     // on Windows there can be file errors from tar if parallel installs
     // are happening (not uncommon with multiple native modules) so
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "gyp"
   ],
   "version": "9.3.1",
-  "installVersion": 10,
```
StefanStojanovic, line 15: A question for people with more experience in `node-gyp`: Since a PR that changed `installVersion` from 9 to 10 landed yesterday in https://github.com/nodejs/node-gyp/commit/aaa117c514430aa2c1e568b95df1b6ed1c1fd3b6 do I need to increase it to 11 now, or not, because there was no release with `installVersion: 10`?

lib/install.js
```diff
@@ -92,6 +96,26 @@ async function install (fs, gyp, argv) {
...
     }
     log.verbose('install', 'version is good')
+    if (win) {
+      log.verbose('on Windows; need to check node.lib')
+      const nodeLibPath = path.resolve(devDir, arch, 'node.lib')
+      try {
+        await fs.promises.stat(nodeLibPath)
+      } catch (err) {
+        if (err.code === 'ENOENT') {
+          log.verbose('install', `version not already installed for ${arch}, continuing with install`, release.version)
```
dsanders11, line 106:
```suggestion
          log.verbose('install', `node.lib not already installed for ${arch}, continuing with install`)
```

lib/install.js
```diff
@@ -22,6 +22,10 @@ const streamPipeline = util.promisify(stream.pipeline)
 
 async function install (fs, gyp, argv) {
   const release = processRelease(argv, gyp, process.version, process.release)
+  // Detecting target_arch based on logic from create-cnfig-gyp.js. Used on Windows only.
```
dsanders11, line 25:
```suggestion
  // Detecting target_arch based on logic from create-config-gypi.js. Used on Windows only.
```

dsanders11 (commented):
Without going too deep on it, seems reasonable.

For more context, what's the main motivation for this change? Saving disk space, overall efficiency, fixing a corner case?

This does create a small behavior change - while before you could do `node-gyp install` to "pre-warm" the cache and then do subsequent builds using that cache, say offline, now a build with a specific `--target` value might require a download which would fail without network connection. Not sure that's necessarily a problem, just noting the change in behavior for end users.

StefanStojanovic:
> For more context, what's the main motivation for this change? Saving disk space, overall efficiency, fixing a corner case?

The main goal is to minimize the number of requests made to the download server, thus decreasing the network traffic. There are some random issues that occur (eg. https://github.com/nodejs/node-gyp/issues/2847) that we'd like to get rid of.

> This does create a small behavior change - while before you could do `node-gyp install` to "pre-warm" the cache and then do subsequent builds using that cache, say offline, now a build with a specific `--target` value might require a download which would fail without network connection. Not sure that's necessarily a problem, just noting the change in behavior for end users.

This is a good point. Thanks for bringing it up. I agree this changes the way `node-gyp install` works, but it is still possible to pass `target_arch` to `node-gyp install` and get node.lib you need (even all 3 of them, but you'll need to run it 3 times). Because this changes how the install command works by default, I've added a `semver-major` label to it.

StefanStojanovic:
Can someone from @nodejs/node-gyp please take a look? Please note that the failed GitHub Actions are not caused by this change, it's a general problem that `node-gyp` started having after updating `make-fetch-happen`.

StefanStojanovic:
After rebasing to the latest main (with the fix for GitHub Actions), all the tests are passing. @nodejs/node-gyp Is there something else I should do here?

package.json
dsanders11: > A question for people with more experience in `node-gyp`

I'm not more experienced, but my 2 cents: it's probably unnecessary for end users (since as you stated, no release went out), but AFAIK there's no harm in bumping it again, and it will prevent a situation where anyone who tested `main` (so folks working on `node-gyp` like you and I, or maybe even someone some where consuming `node-gyp@main` for whatever reason) and got their cache populated for `installVersion: 10` won't end up in a broken situation.

So to me bumping it to 11 seems reasonable to prevent any problems.

dsanders11 (approved):
Overall LGTM, although I have not tested it locally. Also AFAIK my check mark is not green here but happy to give it anyway. 😄

cclauss (approved):
RSLGTM

dsanders11:
@StefanStojanovic, just realized this could cause problems due to this `installVersion` check allowing newer versions without reinstalling:

https://github.com/nodejs/node-gyp/blob/33391db3a0008eff8408890da6ab232f2f90fcab/lib/install.js#L90

If a user were to do something like `npx node-gyp install`, `npx` would use `node-gyp@9.4.0` (current release) and as a result would not download `node.lib` for all architectures. If the user then does `npm install` with a different `target_arch`, their npm-bundled version of `node-gyp` will (likely) be older, it would see that the `installVersion` is newer than its version and consider it "good", but then fail when it can't find `node.lib` for the `target_arch`. 

Seems like the check I've linked above should be `!=`, not `<`. Otherwise changes requiring bumping `installVersion` have to be backwards-compatible due to the aforementioned scenario, which doesn't seem reasonable. If the `installVersion` do not match it seems reasonable to force a reinstall.

Although that does mean that you could get fighting versions since `npx node-gyp` does not use the npm-bundled version. 🤔 That behavior already exists though if your npm-bundled version of `node-gyp` has a lower `installVersion` since `npx node-gyp` would reinstall it.

PR #2868, opened 2023-06-09
test: remove deprecated Node.js and Python

StefanStojanovic:
##### Description of change
Since Node.js v14 has reached its EOL, this PR removes it from GitHub Actions for testing.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,8 +19,8 @@ jobs:
       fail-fast: false
       max-parallel: 15
       matrix:
-        node: [14.x, 16.x, 18.x]
-        python: ["3.7", "3.9", "3.11"]
+        node: [16.x, 18.x, 20.x]
+        python: ["3.9", "3.11"]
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       max-parallel: 15
       matrix:
-        node: [14.x, 16.x, 18.x]
+        node: [16.x, 18.x]
```
cclauss, line 22: Why not...
```suggestion
        node: [16.x, 18.x, 20x]
```

.github/workflows/tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
...
         python: ["3.7", "3.9", "3.11"]
```
cclauss, line 23: Python 3.7 is also EOL this month...  What about just `python: [3.8, 3.11]`?
* https://devguide.python.org/versions/

cclauss (approved):

mhdawson (approved):
LGTM

PR #2886, opened 2023-07-17
fix: a Python lint error by using an f-string

cclauss:
##### Checklist

- [x] `npm install && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [ ] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

PR #2887, opened 2023-07-19
test: increase mocha timeout

StefanStojanovic:
##### Checklist

##### Description of change

Since migrating from Tap to Mocha it can be observed that sometimes tests will fail on Windows because of the "after all" hook is not executed within 2 seconds. This change increases the default timeout affecting that hook in order to avoid failures due to this hook timing out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,6 @@
   },
   "scripts": {
     "lint": "standard */*.js test/**/*.js",
-    "test": "npm run lint && mocha --reporter=test/reporter.js test/test-download.js test/test-*"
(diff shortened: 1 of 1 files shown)


cclauss:
The Python lint error is fixed in #2886

cclauss (approved):
LGTM

cclauss:
This seems to cause errors on our GitHub Actions...
> Starting test 'parallel installs (tarball)'
Test 'parallel installs (tarball)' failed in 599997 ms with Error: Timeout of 600000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/node-gyp/node-gyp/test/test-install.js)

Issue #2873, opened 2023-06-15
9.4.0 should've been semver-major


blattersturm:
We're stuck running on Node.js 12.x due to [node-ffi-napi](https://github.com/node-ffi-napi/node-ffi-napi/issues) being nondeterministically [broken](https://github.com/node-ffi-napi/node-ffi-napi/issues/97) on any release above 14.x.

In this project, we install `node-gyp@9`, but since 9.4.0 bumped to `make-fetch-happen@11` (via 02480f6b6894a6bb81dd3e1365bc4df6d5336f7c) which has a semver-major change that propagates downward, installing fails as such:

```
error make-fetch-happen@11.1.1: The engine "node" is incompatible with this module. Expected version "^14.17.0 || ^16.13.0 || >=18.0.0". Got "12.22.12"
error Found incompatible module.
```

From the [make-fetch-happen changelog](https://github.com/npm/make-fetch-happen/blob/v11.0.3/CHANGELOG.md#1100-2022-10-13):

> ### ⚠️ BREAKING CHANGES
>
> * `make-fetch-happen` is now compatible with the following semver range for node: `^14.17.0 || ^16.13.0 || >=18.0.0`

Another failing in this amazing 'semver' system? Probably.

lukekarrys:
`v9.4.1` has been released which reverts the dependency changes that caused this issue. Those changes will be released again as `v10.0.0` with the changelog outlining the newly supported Node versions.

PR #2922, opened 2023-10-27
chore: add check engines script to CI

lukekarrys:
This will test against regressions in (sub-)dependency engine changes and guard against breaking changes like #2848 #2873.

Errors will look like this:
```
Error: The following production dependencies are not compatible with `engines.node: ^14.17.0 || ^16.13.0 || >=18.0.0` found in `/home/runner/work/node-gyp/node-gyp/package.json`:
make-fetch-happen@13.0.0:
  engines.node: ^16.14.0 || >=18.0.0
  location: node_modules/make-fetch-happen
which@4.0.0:
  engines.node: ^16.13.0 || >=18.0.0
  location: node_modules/which
```

### DIFF
--- a/.github/scripts/check-engines.js
+++ b/.github/scripts/check-engines.js
@@ -1,0 +1,41 @@
+const { join } = require('path')
+const semver = require('semver')
+const Arborist = require('@npmcli/arborist')
+
+const run = async (path, useEngines) => {
+  const pkgPath = join(path, 'package.json')
+  const pkg = require(pkgPath)
+
+  const engines = useEngines || pkg.engines.node
+
+  const arb = new Arborist({ path })
+  const tree = await arb.loadActual({ forceActual: true })
+  const deps = await tree.querySelectorAll(`#${pkg.name} > .prod:attr(engines, [node])`)
+
+  const invalid = []
+  for (const dep of deps) {
+    const depEngines = dep.target.package.engines.node
+    if (!semver.subset(engines, depEngines)) {
+      invalid.push({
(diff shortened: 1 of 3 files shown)


DeeDeeG (approved):
Adding this check seems like a good idea! (Given how it impacted users last time dependencies were bumped outside of the "advertised" supported versions of Node.)

I haven't read through all the code or anything, but I would approve of doing this, generally speaking!

I tried moving it to its own workflow file on my fork, seems to work just fine: https://github.com/DeeDeeG/node-gyp/commit/7564e32d754c4c0957dd738b42e6d28b2ce226c5

Feel free to copy that if wanting it in its own separate GitHub Actions workflow file (separate `.yml` CI file).

lukekarrys:
@DeeDeeG The comment I left in the code was too vague, sorry about that. I meant I would like to move the whole thing to its own GitHub action to be used like `uses: pkgjs/check-engines`. 

I use this same action other places and I'd love to find a good org for to be used ecosystem-wide. I talked a bit to some folks at the `pkgjs` org and I think that might be a good place for it.

DeeDeeG:
You might have seen this already, but the updated `engines:` field got removed in one of the force pushes, and some just-merged dependency bumps are failing the "Check Engines" job in CI without that updated `engines:` value.

NodeJS 14 is still in there, and the 16 is still `^16.13.0` instead of `^16.14.0`.

lukekarrys:
I was having some trouble getting the tests passing in CI so I decided to separate out all the commits into separate PRs. It did end up showing that the `check-engines` job will work though! 😄

PR #2946, opened 2023-11-09
deps: env-paths@3.0.0

lukekarrys:


lukekarrys:
This requires `env-paths` to be imported using `await import()` so I took the opportunity to refactor the bin file to use more `async/await`.

This led me to realize that some of this file is untested so I wanted to hold off landing this until I could ensure appropriate coverage of this file.

lukekarrys:
Closing in favor of #3235 since we can now land this dep with no refactoring thanks to require(esm)!

PR #2957, opened 2023-12-23
Detect visual studio installation using VSSetup module 

jarig:
##### Checklist

- [X] tests are included 

##### Description of change

Detect visual studio installation using VSSetup module and Get-VSSetupInstance method as a first attempt, this approach works even in systems with Constrained language mode of the powershell. 
Original type-base detection requires `FullLanguage` mode, which quite often not available on the enrolled devices.

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -54,6 +54,7 @@ class VisualStudioFinder {
     }
 
     const checks = [
+      () => this.findVisualStudio2017OrNewerUsingSetupModule(),
       () => this.findVisualStudio2017OrNewer(),
       () => this.findVisualStudio2015(),
       () => this.findVisualStudio2013()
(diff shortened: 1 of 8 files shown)


jarig:
Verified tests on both windows and linux now
image
image

jarig:
Sry last linting issue also fixed. Btw mind to add `.eslintrc` into the project? VSCode not scanning automatically it looks

jarig:
@cclauss any other activities or blockers to be done for the PR merge?

test/test-find-visualstudio.js
```diff
@@ -171,7 +189,7 @@ describe('find-visualstudio', function () {
     finder.parseData(null, '[]', '', (info) => {
       assert.ok(/find .* Visual Studio/i.test(finder.errorLog[0]), 'expect error')
       assert.ok(!info, 'no data')
-    })
+    }, { checkIsArray: true })
```
StefanStojanovic, line 192: Input parameters order is bad, the callback will be passed as options. Also, seems this test is outdated as the callback is never called in general (no `parseData` parameter for any type of callback).

test/test-find-visualstudio.js
```diff
@@ -189,7 +207,7 @@ describe('find-visualstudio', function () {
       assert.ok(/unknown version/i.test(finder.errorLog[0]), 'expect error')
       assert.ok(/find .* Visual Studio/i.test(finder.errorLog[1]), 'expect error')
       assert.ok(!info, 'no data')
-    })
+    }, { checkIsArray: true })
```
StefanStojanovic, line 210: Input parameters order is bad, the callback will be passed as options. Also, seems this test is outdated as the callback is never called in general (no `parseData` parameter for any type of callback).

StefanStojanovic (commented):
The PR looks good overall. I have [another PR](https://github.com/nodejs/node-gyp/pull/2959) that works on the same parts of the code, so this one should land first as my changes should be in the new `findVisualStudio2017OrNewerUsingSetupModule` function as well. I can rebase my changes on top of this one after it lands.

I've noticed something with the tests here though. There are no `VSSetup` files in fixtures for all of the fixtures we had before. The problem with it is that in our tests we will simulate the VSSetup function to return null when it should return something (eg. having 2019 installed). I'm aware this would be a tedious task to make all those fixtures and I'm not sure it's worth the effort, but just wanted to bring it up in case it comes up later.

lib/find-visualstudio.js
```diff
@@ -113,6 +114,48 @@ class VisualStudioFinder {
     throw new Error('Could not find any Visual Studio installation to use')
   }
 
+  async findVisualStudio2017OrNewerUsingSetupModule (vcInstallDir) {
```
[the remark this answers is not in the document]
jarig: Not required in this case, made it for any potential future parametrization or re-use. 
But can remove to avoid potential confusion and for consistency with the current approach.

StefanStojanovic:
@jarig my [PR](https://github.com/nodejs/node-gyp/pull/2959) is ready for landing. Do you have time to address the feedback I gave? If you are busy or unavailable, I can take over and make the required changes.

jarig:
> @jarig my [PR](https://github.com/nodejs/node-gyp/pull/2959) is ready for landing. Do you have time to address the feedback I gave? If you are busy or unavailable, I can take over and make the required changes.

I'll try to make and finalize changes this week

lib/find-visualstudio.js
```diff
@@ -128,24 +175,35 @@ class VisualStudioFinder {
...
-        'could not use PowerShell to find Visual Studio 2017 or newer, try re-running with \'--loglevel silly\' for more details')
+        `could not use PowerShell to find Visual Studio 2017 or newer, try re-running with '--loglevel silly' for more details. \n
+        Failure details: ${failureDetails}`)
       return null
     }
 
     if (err) {
       this.log.silly('PS err = %j', err && (err.stack || err))
-      return failPowershell()
+      return failPowershell(`${err}`.substring(0, 40))
```
StefanStojanovic, line 206: Just out of curiosity, why cut at 40?

lib/find-visualstudio.js
```diff
@@ -157,11 +215,16 @@ class VisualStudioFinder {
       return failPowershell()
```
StefanStojanovic, line 215: Nitpicking. Just a thought, maybe adding some comment would be good.

StefanStojanovic (approved):
The changes LGTM. Thanks for addressing my comments @jarig. @lukekarrys (since you are a reviewer) if this LGTY too we can go ahead and merge this and I'll rebase and adapt my changes in https://github.com/nodejs/node-gyp/pull/2959 accordingly.

lib/find-visualstudio.js
jarig: Just to give a possible hint where the failure is and then look for more details in the log or `--verbose` output.
Didn't want to put too large number which can clog the output. But it can be something around 70-90 too (usual line length)

StefanStojanovic:
I plan to land this tomorrow. If there are any concerns, please state them today. cc @lukekarrys

Mapuppy09:
Good!

PR #2959, opened 2024-01-15
win: update supported vs versions

StefanStojanovic:
##### Checklist

##### Description of change
Since Node.js v22 will use C++20 standard, VS2017 can no longer be used for compiling native add-ons. This is already changed in [the docs](https://github.com/nodejs/build/blob/main/doc/windows-visualstudio-supported-versions.md#for-building-nodejs-addons). This PR drop VS2017 support for Node.js v22 and above for node-gyp.

**Note:** Node-gyp with this change should be released and consumed by Node for the v22 RC to avoid potential problems on Windows.

cc @targos, @nodejs/node-gyp

Refs: https://github.com/nodejs/build/pull/3603
Refs: https://github.com/nodejs/node/pull/45427

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -54,8 +54,10 @@ class VisualStudioFinder {
     }
 
     const checks = [
-      () => this.findVisualStudio2017OrNewerUsingSetupModule(),
-      () => this.findVisualStudio2017OrNewer(),
+      () => this.findVisualStudio2019OrNewerUsingSetupModule(),
+      () => this.findVisualStudio2019OrNewer(),
+      () => this.findVisualStudio2017UsingSetupModule(),
+      () => this.findVisualStudio2017(),
       () => this.findVisualStudio2015(),
       () => this.findVisualStudio2013()
     ]
@@ -114,7 +116,20 @@ class VisualStudioFinder {
     throw new Error('Could not find any Visual Studio installation to use')
   }
 
-  async findVisualStudio2017OrNewerUsingSetupModule () {
+  async findVisualStudio2019OrNewerUsingSetupModule () {
+    return this.findNewVSUsingSetupModule([2019, 2022])
+  }
+
+  async findVisualStudio2017UsingSetupModule () {
+    if (this.nodeSemver.major >= 22) {
+      this.addLog(
+        'not looking for VS2017 as it is only supported up to Node.js 21')
+      return null
+    }
+    return this.findNewVSUsingSetupModule([2017])
+  }
+
+  async findNewVSUsingSetupModule (supportedYears) {
     const ps = path.join(process.env.SystemRoot, 'System32',
       'WindowsPowerShell', 'v1.0', 'powershell.exe')
     const vcInstallDir = this.envVcInstallDir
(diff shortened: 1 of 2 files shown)


lib/find-visualstudio.js
```diff
@@ -113,9 +114,25 @@ class VisualStudioFinder {
...
+        'not looking for VS2017 as it is only supported up to Node.js 21')
+      return null
+    }
+    return this.findNewVS()
+  }
+
   // Invoke the PowerShell script to get information about Visual Studio 2017
   // or newer installations
-  async findVisualStudio2017OrNewer () {
+  async findNewVS () {
```
targos, line 135: The body of this function isn't changed. Wouldn't it still find VS2017 if that version is installed?

StefanStojanovic: This is just the implementation used for finding both VS2017 and later (2019/2022). It is called a few lines above, but there is a guard to fail for VS2017 if Node.js is v22+. This is [the code](https://github.com/JaneaSystems/node-gyp/blob/f38e957b6ee890540cd089575cb215e1b90c2fb1/lib/find-visualstudio.js#L119-L131) to do that,

StefanStojanovic: Oh that, yes, you're right! My bad, I'll fix it.

StefanStojanovic: @targos I've updated the code to address the issue you've found. Once it is approved, I'll just squash it and force-push. For now, it is separated for easier review.

lib/find-visualstudio.js
```diff
@@ -113,9 +114,25 @@ class VisualStudioFinder {
...
   }
 
+  // Invoke the PowerShell script to get information about Visual Studio 2019
+  // or newer installations
+  async findVisualStudio2019OrNewer () {
+    return this.findNewVS([2019, 2022])
+  }
+
+  // Invoke the PowerShell script to get information about Visual Studio 2017
+  async findVisualStudio2017 () {
```
gengjiawen, line 124: This function should be removed. Also this should be a breaking change. commit. msg should be like 
```
win!: update supported vs versions
```

StefanStojanovic: I've changed the commit message. Should I also add the `semver-major` label?

Regarding the removal of `findVisualStudio2017`, why do you think it should be removed? I've added it in this PR so that VS2017 can be prohibited on Node v22+, while VS2019 and VS2022 will be enabled. On previous Node versions all 3 VS versions should work, like they are working in the current `main` branch.

targos (commented):
LGTM, but I don't have authority on this project.

lib/find-visualstudio.js
lukekarrys: I don't follow the reasoning why this would be a breaking change. @gengjiawen Can you elaborate on this?

lukekarrys (approved):
LGTM. I would like others' thoughts on if this is a breaking change as referenced here https://github.com/nodejs/node-gyp/pull/2959#discussion_r1452877626.

I don't think it is, in which case I can squash and merge with the PR title as the commit so the automated release does not tag this as a semver major.

gengjiawen:
> LGTM. I would like others' thoughts on if this is a breaking change as referenced here https://github.com/nodejs/node-gyp/pull/2959#discussion_r1452877626.

this mainly follow Node.js main repo for compiler change.

lukekarrys:
@StefanStojanovic Do you think this should be merged while #2957 is awaiting updates to your feedback?

StefanStojanovic:
> @StefanStojanovic Do you think this should be merged while #2957 is awaiting updates to your feedback?

This can be merged. The thing is, if that happens, the function other PR adds (`findVisualStudio2017OrNewerUsingSetupModule`) should also be split into 2 parts: 2017 and 2019+. I can do that work in that PR directly or as additional PR after that one lands.

Just a note, before merging I should squash the fixup commit and probably change `win!` to `win` at the beginning of the first commit message. I'll do that a bit later during the day

StefanStojanovic:
@lukekarrys I've made the changes preparing this for merge. As I mentioned earlier, I have no problem making adjustments in https://github.com/nodejs/node-gyp/pull/2957 (or in a separate PR) to get it in line with this.

StefanStojanovic:
I've rebased this PR on top of the main branch, which now has the changes from https://github.com/nodejs/node-gyp/pull/2957. @lukekarrys, since you've already approved my previous changes, I'd like to ask you to review the new ones and if it LGTY, land this.

StefanStojanovic:
A reminder to @nodejs/node-gyp about this PR. I've adjusted it after https://github.com/nodejs/node-gyp/pull/2957 landed, and as far as I can tell, it should be ready for merging.

StefanStojanovic:
Since this is already approved and no there were no objections, I plan to land this tomorrow. If there are any complaints please state them until then.

lukekarrys:
Sorry for the delay on review @StefanStojanovic and thank you for landing this.

I will work on getting this published this week.

PR #2964, opened 2024-01-19
src: add support for locally installed headers

mhdawson:
##### Checklist

- [x] tests are included 

##### Description of change

Some linux distros allow headers to be installed through tools like rpm. If the runtime sets
process.config.variables.use_prefix_to_find_headers, look for matching headers based on the directory set for the prefix in process.config.variables.prefix

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -39,7 +39,7 @@ async function configure (gyp, argv) {
       const prefix = process.config.variables.node_prefix
       let availVersion
       try {
-        const nodeVersionH = fsSync.readFileSync(path.join(prefix,
+        const nodeVersionH = readFileSync(path.join(prefix,
           'include', 'node', 'node_version.h'), { encoding: 'utf8' })
         const major = nodeVersionH.match(majorRe)[1]
         const minor = nodeVersionH.match(minorRe)[1]
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -32,11 +32,11 @@ async function configure (gyp, argv) {
     // 'python' should be set by now
     process.env.PYTHON = python
 
-    const prefix = process.config.variables.node_prefix
     if (!gyp.opts.nodedir &&
         process.config.variables.use_prefix_to_find_headers) {
       // check if the headers can be found using the prefix specified
       // at build time. Use them if they match the version expected
+      const prefix = process.config.variables.node_prefix
       let availVersion
       try {
         const nodeVersionH = fsSync.readFileSync(path.join(prefix,
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const { promises: fs, readFileSync } = require('graceful-fs')
-const fsSync = require('graceful-fs')
 const path = require('path')
 const log = require('./log')
 const os = require('os')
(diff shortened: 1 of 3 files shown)


mhdawson:
Not sure if this should be added to doc, if so just let me know and I'll look for the right place.

mhdawson:
Upstream PR in Node.js to allow setting use_prefix_to_find_headers - https://github.com/nodejs/node/pull/51525

mhdawson:
Note that the addition to [test/test-configure-python.js](https://github.com/nodejs/node-gyp/pull/2964/files#diff-310cfc6e538a2710f5919f4fc313baafcf5a8b2df6a50caf6cab53c17b346ca0) was needed so that the tests would run ok when  applied to other streams, and does not adversely affect testing in head

mhdawson:
From the test failure on windows, I do see the test is too linux specific will take a look.

mhdawson:
Guess it was more than just the path, spinning up a windows machine to investigate on.

mhdawson:
Ok, windows all fixed  up :)

lib/configure.js
```diff
@@ -27,6 +32,28 @@ async function configure (gyp, argv) {
     // 'python' should be set by now
     process.env.PYTHON = python
 
+    const prefix = process.config.variables.node_prefix
```
richardlau, line 35: `prefix` is only used inside the following `if` conditional so could be moved there.

mhdawson: @richardlau thanks, will move

mhdawson: moved

lib/configure.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 const { promises: fs } = require('graceful-fs')
```
lukekarrys, line 3:
```suggestion
const { promises: fs, readFileSync } = require('graceful-fs')
```

lib/configure.js
```diff
@@ -27,6 +32,28 @@ async function configure (gyp, argv) {
...
     process.env.PYTHON = python
 
+    if (!gyp.opts.nodedir &&
+        process.config.variables.use_prefix_to_find_headers) {
+      // check if the headers can be found using the prefix specified
+      // at build time. Use them if they match the version expected
+      const prefix = process.config.variables.node_prefix
+      let availVersion
+      try {
+        const nodeVersionH = fsSync.readFileSync(path.join(prefix,
```
lukekarrys, line 42:
```suggestion
        const nodeVersionH = readFileSync(path.join(prefix,
```

lukekarrys (approved):
LGTM! One minor suggestion to only expose `readFileSync`.

lib/configure.js
lukekarrys: I find this easier to refactor in the future especially since we have to mock the individual methods for tests.

lib/configure.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
-const { promises: fs } = require('graceful-fs')
+const { promises: fs, readFileSync } = require('graceful-fs')
+const fsSync = require('graceful-fs')
```
lukekarrys, line 4:
```suggestion
```

Issue #2829, opened 2023-04-11
Python path cannot handle Unicode characters like ü


xHyroM:
Please look thru your error log for the string `gyp info using node-gyp@` and if the version number is less than the [current release of node-gyp](https://github.com/nodejs/node-gyp/releases) then __please upgrade__ using the instructions at https://github.com/nodejs/node-gyp/blob/master/docs/Updating-npm-bundled-node-gyp.md and try your command again.

Requests for help with [`node-sass` are very common](https://github.com/nodejs/node-gyp/issues?q=label%3A%22Node+Sass+--%3E+Dart+Sass%22). Please be aware that this package is deprecated, you should seek alternatives and avoid opening new issues about it here.

* **Node Version**:  node: v19.3.1 npm: 9.5.1
* **Platform**: Microsoft Windows 11 Home Insider Preview Version 22H2 (OS Build 23424.1000)
* **Compiler**:  command not recognized
* **Module**:  better-sqlite3

<details><summary>Verbose output (from npm or node-gyp):</summary>

npm i better-sqlite3 --python="C:\Users\Jozef Steinhübl\AppData\Local\Programs\Python\Python311\python.exe" --verbose
```
0 verbose cli C:\Program Files\nodejs\node.exe C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js
1 info using npm@9.5.1
2 info using node@v19.8.1
3 timing npm:load:whichnode Completed in 2ms
4 timing config:load:defaults Completed in 1ms
5 timing config:load:file:C:\Program Files\nodejs\node_modules\npm\npmrc Completed in 2ms
6 timing config:load:builtin Completed in 2ms
7 timing config:load:cli Completed in 1ms
8 timing config:load:env Completed in 1ms
9 timing config:load:project Completed in 0ms
10 timing config:load:file:C:\Users\Jozef Steinhübl\.npmrc Completed in 1ms
11 timing config:load:user Completed in 1ms
12 timing config:load:file:C:\Users\Jozef Steinhübl\AppData\Roaming\npm\etc\npmrc Completed in 0ms
13 timing config:load:global Completed in 0ms
14 timing config:load:setEnvs Completed in 1ms
15 timing config:load Completed in 7ms
16 timing npm:load:configload Completed in 7ms
17 timing npm:load:mkdirpcache Completed in 0ms
18 timing npm:load:mkdirplogs Completed in 0ms
19 verbose title npm i better-sqlite3
[...]

PR #2987, opened 2024-03-05
lib: print Python executable path using UTF-8

huseyinacacak-janea:
##### Checklist

##### Description of change
The Python executable path may have non-ASCII characters, which can make the print function fail if the environment encoding is different. This PR fixes this issue by using `stdout.buffer`, which can be used with UTF-8 encoding for the output, regardless of the environment encoding.

Fixes: https://github.com/nodejs/node-gyp/issues/2829

### DIFF
--- a/lib/find-python.js
+++ b/lib/find-python.js
@@ -41,7 +41,7 @@ class PythonFinder {
   static findPython = (...args) => new PythonFinder(...args).findPython()
 
   log = log.withPrefix('find Python')
-  argsExecutable = ['-c', 'import sys; print(sys.executable);']
+  argsExecutable = ['-c', 'import sys; sys.stdout.buffer.write(sys.executable.encode(\'utf-8\'));']
   argsVersion = ['-c', 'import sys; print("%s.%s.%s" % sys.version_info[:3]);']
   semverRange = '>=3.6.0'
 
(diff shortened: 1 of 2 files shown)


cclauss:
Please create a test or tests that fails with the old code and pass with the new code so we will be alerted if there are future regressions.

huseyinacacak-janea:
@cclauss thanks for reviewing. I've added a test.

lukekarrys (approved):

lukekarrys:
Linting is failing in CI which is unfortunately preventing the tests from being run.

I did check this out and run it locally on macOS and they were failing. It could be because the test assumes that the result of `PythonFinder.findPython(null)` will be a single file which can be copied, but this is not the case at least on my machine.

cclauss:
The linter errors are caused by a new version of `ruff` and are fixed in nodejs/gyp-next#226

huseyinacacak-janea:
@lukekarrys thank you for your feedback. I've updated the test to fix the failure on MacOS.
There are still linter errors because of `ruff`. I can re-run them once `ruff` fix lands in `node-gyp`.

lukekarrys:
@huseyinacacak-janea I think linting should now be fixed in main. Can you try updating your branch now?

huseyinacacak-janea:
I rebased the branch to get the `ruff` fix. However, as I deleted the branch accidentally while pushing, I've opened a new PR with the same changes: https://github.com/nodejs/node-gyp/pull/2995

PR #2990, opened 2024-03-07
feat(gyp): update gyp to v0.16.2

lukekarrys:


lukekarrys:
I believe the lint failing here will be resolved by https://github.com/nodejs/gyp-next/pull/226.

I will wait for the next release and replace this PR when it is available.

PR #3036, opened 2024-05-29
win: allow VCINSTALLDIR to specify a portable instance

legendecas:
##### Checklist

- [x] tests are included 
- [ ] ~~documentation is changed or added~~

##### Description of change

When node-gyp is running in a developer command prompt from a portable enterprise VS instance, allow node-gyp to use the specified instance instead of reporting that the instance found from the VSInstaller doesn't match the environ `VCINSTALLDIR`. Tools like CMake can use a portable VS instance as well: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6651.


StefanStojanovic (commented):
This looks reasonable. A few things are assumed, so they are a weak spot, but I guess there is not much we can do about it?

legendecas:
@StefanStojanovic thanks for the review. While trusting what the VS developer command prompt sets in the environment variables, the `VisualStudioFinder` also checks if the needed commands exist, like in https://github.com/nodejs/node-gyp/blob/main/lib/find-visualstudio.js#L380-L381.

StefanStojanovic (approved):

PR #3049, opened 2024-07-19
Update tar and glob

lukekarrys:
- **deps: tar@7.4.0**
- **deps: glob@11.0.0**


lukekarrys:
Neither of these deps are compatible with the current engines (`^16.14.0 || >=18.0.0`) and will need to wait until the next major version.

PR #3065, opened 2024-08-28
install: don't preserve owner when extracting files

behrmann:
##### Checklist

- [ ] tests are included 

##### Description of change

tar defaults to `true` for `preserveOwner` when run as root. 

In a user namespace one can be root, but not have the capability to chown files, e.g. because of a seccomp filter. 

It doesn't make sense to preserve the owner anyway, root or not, since the owner in a tar is chosen by the side providing the tar file, so the user name might not be present in the system.  Also, all other files written, would also have a different owner than what was extracted from the tar file.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -212,6 +212,7 @@ async function install (gyp, argv) {
             file: tarPath,
             strip: 1,
             filter: isValid,
+            preserveOwner: false,
             onwarn,
             cwd: tarExtractDir
           })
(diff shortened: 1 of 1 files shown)


behrmann:
My bad, the sandbox that made this necessary actually didn't catch the chown calls, so this explains why the errors from chown bubbled up. That being said, even if gyp could untar things with their uid from the tar, it shouldn't since it doesn't control the input and the uid might not exist on the system. Chowning to an effectively random number doesn't make a lot of sense.

PR #3102, opened 2024-12-02
Update engines and dependencies

lukekarrys:
This PR updates the engines from `^16.14.0 || >=18.0.0` to `^18.17.0 || >=20.5.0` and updates all dependencies that were behind on major versions due them not supporting Node 16.

The new engines range was chosen to align with the `npm` dependencies that were updated such as [`make-fetch-happen`](https://github.com/npm/make-fetch-happen/blob/ecd8f33c411096e8f79ae6f7043edab11ab29289/package.json#L57).

This PR also closes #2663 as it updates `make-fetch-happen` to a version with the necessary fix.

PR #3103, opened 2024-12-03
chore: migrate from standard to neostandard

lukekarrys:
This is built on #3102.

After updating those dependencies, `standard` was the only dep bringing in anything deprecated. I read https://github.com/standard/standard/issues/1948#issuecomment-2138078249 and opted to migrate to `neostandard` following the instructions here: https://github.com/neostandard/neostandard?tab=readme-ov-file#migrate-from-standard

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,0 +1,3 @@
+'use strict'
+
+module.exports = require('neostandard')({})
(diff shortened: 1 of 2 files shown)


legendecas:
LGTM once rebased

PR #3105, opened 2024-12-03
chore: add gyp-next updater

legendecas:
Add automated updater for gyp-next.

Example: https://github.com/legendecas/node-gyp/pull/1

- [x] Add nodejs-github-bot token

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -9,13 +9,12 @@ jobs:
   release-please:
     outputs:
       release_created:  ${{ steps.release.outputs.release_created }}
-    permissions:
-      contents: write # to create release commit (googleapis/release-please-action)
-      pull-requests: write # to create release PR (googleapis/release-please-action)
     runs-on: ubuntu-latest
     steps:
       - uses: googleapis/release-please-action@v4
         id: release
+        with:
+          token: ${{ secrets.GH_USER_TOKEN }}
           # Standard Conventional Commits: `feat` and `fix`
           # node-gyp subdirectories: `bin`, `gyp`, `lib`, `src`, `test`
           # node-gyp subcommands: `build`, `clean`, `configure`, `install`, `list`, `rebuild`, `remove`
--- a/.github/workflows/update-gyp-next.yml
+++ b/.github/workflows/update-gyp-next.yml
@@ -14,7 +14,6 @@ env:
 
 jobs:
   update-gyp-next:
-    # if: github.repository == 'nodejs/node' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/update-gyp-next.yml
+++ b/.github/workflows/update-gyp-next.yml
@@ -1,0 +1,52 @@
+name: Update gyp-next
+
+on:
+  schedule:
+    # Run once a week at 12:00 AM UTC on Sunday.
+    - cron: 0 0 * * *
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  NODE_VERSION: lts/*
+
+jobs:
+  update-gyp-next:
+    # if: github.repository == 'nodejs/node' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - uses: actions/github-script@v7
+        id: get-gyp-next-version
+        with:
+          script: |
+            const result = await github.rest.repos.getLatestRelease({
+              owner: 'nodejs',
+              repo: 'gyp-next',
+            });
+            return result.data.tag_name
+          result-encoding: string
+
+      - name: Update gyp-next
+        run: |
+          python update-gyp.py --no-commit ${{ steps.get-gyp-next-version.outputs.result }}
+
+      - name: Open or update PR for the gyp-next update
+        uses: gr2m/create-or-update-pull-request-action@v1
+        with:
+          branch: actions/update-gyp-next
+          author: Node.js GitHub Bot <github-bot@iojs.org>
+          title: 'feat: update gyp-next to ${{ steps.get-gyp-next-version.outputs.result }}'
+          commit-message: 'feat: update gyp-next to ${{ steps.get-gyp-next-version.outputs.result }}'
+          update-pull-request-title-and-body: true
+          body: >
+            This is an automated update of the gyp-next to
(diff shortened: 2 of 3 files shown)


.github/workflows/update-gyp-next.yml
```diff
@@ -0,0 +1,52 @@
...
+            });
+            return result.data.tag_name
+          result-encoding: string
+
+      - name: Update gyp-next
+        run: |
+          python update-gyp.py --no-commit ${{ steps.get-gyp-next-version.outputs.result }}
+
+      - name: Open or update PR for the gyp-next update
+        uses: gr2m/create-or-update-pull-request-action@v1
```
cclauss, line 41: Do you have experience with this one?  https://github.com/peter-evans/create-pull-request from GitHub staff is broadly  used but if you are confident in this one, great.

update-gyp.py
```diff
@@ -60,5 +64,6 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
             os.path.join(unzip_target, os.listdir(unzip_target)[0]), CHECKOUT_GYP_PATH
         )
 
-subprocess.check_output(["git", "add", "gyp"], cwd=CHECKOUT_PATH)
-subprocess.check_output(["git", "commit", "-m", "feat(gyp): update gyp to " + args.tag])
+if not args.no_commit:
+  subprocess.check_output(["git", "add", "gyp"], cwd=CHECKOUT_PATH)
+  subprocess.check_output(["git", "commit", "-m", "feat(gyp): update gyp to " + args.tag])
```
cclauss, line 69:
```suggestion
  subprocess.check_output([
      "git", "commit", "-m", "feat(gyp): update gyp to " + args.tag
  ])
```

cclauss (approved):
Nice!

.github/workflows/update-gyp-next.yml
legendecas: This is used in the nodejs/node repo: https://github.com/nodejs/node/blob/main/.github/workflows/update-openssl.yml#L53

legendecas: GitHub should make one of these official :)

.github/workflows/update-gyp-next.yml
```diff
@@ -0,0 +1,52 @@
...
+
+permissions:
+  contents: read
+
+env:
+  NODE_VERSION: lts/*
+
+jobs:
+  update-gyp-next:
+    # if: github.repository == 'nodejs/node' || github.event_name == 'workflow_dispatch'
```
lukekarrys, line 17:
```suggestion
```

.github/workflows/update-gyp-next.yml
```diff
@@ -0,0 +1,52 @@
...
+          branch: actions/update-gyp-next
+          author: Node.js GitHub Bot <github-bot@iojs.org>
+          title: 'feat: update gyp-next to ${{ steps.get-gyp-next-version.outputs.result }}'
+          commit-message: 'feat: update gyp-next to ${{ steps.get-gyp-next-version.outputs.result }}'
+          update-pull-request-title-and-body: true
+          body: >
+            This is an automated update of the gyp-next to
+            https://github.com/nodejs/gyp-next/releases/tag/${{ steps.get-gyp-next-version.outputs.result }}.
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_USER_TOKEN }}
```
lukekarrys, line 52:
```suggestion
          GITHUB_TOKEN: ${{ secrets.GYP_NEXT_UPDATE_TOKEN }}
```

Giving this token a more specific name will help in the future if it needs to be rotated, revoked, etc.

.github/workflows/release-please.yml
```diff
@@ -9,13 +9,12 @@ jobs:
...
       release_created:  ${{ steps.release.outputs.release_created }}
-    permissions:
-      contents: write # to create release commit (googleapis/release-please-action)
-      pull-requests: write # to create release PR (googleapis/release-please-action)
     runs-on: ubuntu-latest
     steps:
       - uses: googleapis/release-please-action@v4
         id: release
+        with:
+          token: ${{ secrets.GH_USER_TOKEN }}
```
lukekarrys, line 17:
```suggestion
```

I think we should continue to use the workflow permissions model to lessen the need for long lived tokens. Obviously this PR will need one to update `gyp-next` but I think it's better to create a single token whose only responsibility is creating the `gyp-next` PR.

lukekarrys (changes requested):
This PR is a great idea and will help a lot so I'm +1 on landing it.

The only change I'm requesting is that to use a more specific GitHub token that is only used for updating `gyp-next`.

legendecas:
@lukekarrys the new token permissions are tracked at https://github.com/nodejs/admin/pull/935. It has the same permission needed for release-please and gyp-next updater as:

```
permissions:
  contents: write
  pull-requests: write
```

The reason to use a token for release-please is that the token GitHub actions created by default will not trigger GitHub actions:

>   When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.
>   https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow

So we need a token of @nodejs-github-bot, as requested at https://github.com/nodejs/admin/pull/935.

lukekarrys (approved):

lukekarrys:
@legendecas Good call, I forgot about that restriction. I withdraw my previous suggestions and 👍

lukekarrys:
@legendecas What do you think about the suggestions in the unresolved comment threads?

legendecas:
yes, I will update the PR according to the suggestions above once the token is setup. Thanks for catching them!

legendecas:
Token is created, this is ready for review!

update-gyp.py
```diff
@@ -66,4 +66,6 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
 
 if not args.no_commit:
   subprocess.check_output(["git", "add", "gyp"], cwd=CHECKOUT_PATH)
-  subprocess.check_output(["git", "commit", "-m", "feat(gyp): update gyp to " + args.tag])
+  subprocess.check_output([
+    "git", "commit", "-m", "feat(gyp): update gyp to " + args.tag
```
cclauss, line 70:
```suggestion
    "git", "commit", "-m", f"feat(gyp): update gyp to {args.tag}"
```

legendecas:
Generated PR https://github.com/nodejs/node-gyp/pull/3122

Issue #3095, opened 2024-11-21
Parallel node-gyp builds failing due to concurrent file access (Windows only)


MRayermannMSFT:
Hi! I'm looking for help solving an issue we're starting to see more often on Windows, when `npm install` builds our native node modules. 

The issue is that because `npm install` runs native module builds in parallel, and each needs exclusive access to some files, they end up failing due to not being able to get that exclusive access. We believe we're seeing this issue more often because we recently standardized on using `node-addon-api` for all five of our native modules. Thus, we have five parallel builds, all trying to access a lot of the same files. I've provided a series of example errors below (if someone wants full logs, I can provide those as well).

To be clear, I **don't** think this is necessarialy node-gyp's fault, however it may be part of the solution. For now, I am here to see if anyone in the node-gyp community has any ideas/thoughts/suggestions on how to proceed next. So far here's some ideas I've had:

1. Use `--install-strategy=nested` to de-deduplicate `node-addon-api`, thus reducing (maybe even eliminating) the chance of conflicting file access. Our project is ultimately bundled, so this will probably work out ok, but there are likely going to be other complications. Would be nice if we could only de-deduplicate `node-addon-api`.
2. Combine some of the five native modules, thus reducing the total number. This won't fix the problem, but it'll reduce the chance of it happening.
3. Solicit `npm` to not do parallel native node module installs.
4. Add a retry mechanism to node-gyp.
5. Prebuild the binaries for as many of our native node_modules as we can, thus avoiding having to rebuild them for the main project.
6. Write our own build-time code to de-deduplicate `node-addon-api`.
7. Use `--ignore-scripts`, and then use `npm rebuild` for any package which has an installScript: true in package-lock. **Currently using this.**
8. Use `--foreground-scripts`. **Need to test this.**

<details><p>

```
The specified task executable "CL.exe" could not be run. System.IO.IOException: The process cannot access the file 'PATH_TO_NODE_MODULES\node-addon-api\Release\obj\nothing\nothing.tlog\CL.command.1.tlog' because it is being used by another process. [PATH_TO_NODE_MODULES\node-addon-api\nothing.vcxproj]
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath) [PATH_TO_NODE_MODULES\node-addon-api\nothing.vcxproj]
[...]

PR #3112, opened 2024-12-29
fix: disable msbuild.exe nodeReuse

tmm1:
##### Checklist

- [ ] tests are included 

##### Description of change

On windows, one may experience errors due to concurrent node-gyp calls (see #3095)

When retrying after such a failure, retries are observed to fail repeatedly with strange
EBUSY and EPERM errors. This turns out to be due to a caching feature where msbuild.exe
will continue to run in the background (for either 15s or 15m depending on VS version),
and keep open file handles to files and directories that the user may be trying to delete
(for example with a new call to `npm ci`).

This behavior is [well](https://stackoverflow.com/questions/51076195/why-is-msbuild-generating-tlog-files-into-cmakefiles-compileridc-and-how-do-i) [documented](https://github.com/dotnet/msbuild/issues/1709), as is the [recommended workaround](https://learn.microsoft.com/en-us/azure/devops/pipelines/troubleshooting/troubleshooting?view=azure-devops#msbuild-and-nodereusefalse) implemented here.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -141,6 +141,8 @@ async function build (gyp, argv) {
     if (msvs) {
       // Turn off the Microsoft logo on Windows
       argv.push('/nologo')
+      // No lingering msbuild processes and open file handles
+      argv.push('/nodeReuse:false')
     }
 
     // Specify the build type, Release by default


StefanStojanovic:
SGTM if the GH actions pass.

StefanStojanovic (approved):

lukekarrys (approved):

PR #3116, opened 2025-01-06
Test on GitHub Actions `windows-2025` image

cclauss:
https://learn.microsoft.com/en-us/windows-server/get-started/windows-server-release-info
https://learn.microsoft.com/en-us/windows-server/get-started/whats-new-windows-server-2025
* actions/runner-images#11228

Windows 2025 currently ships with Visual Studio 2022.

`vswhere -latest -property installationPath`
```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise
```

##### Checklist

- [ ] tests are included 

### DIFF
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -385,7 +385,7 @@ class VisualStudioFinder {
       ret.versionYear = 2019
       return ret
     }
-if (ret.versionMajor === 17) {
+    if (ret.versionMajor === 17) {
       ret.versionYear = 2022
       return ret
     }
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -119,7 +119,7 @@ class VisualStudioFinder {
   }
 
   async findVisualStudio2019OrNewerFromSpecifiedLocation () {
-    return this.findVSFromSpecifiedLocation([2019, 2022, 2025])
+    return this.findVSFromSpecifiedLocation([2019, 2022])
   }
 
   async findVisualStudio2017FromSpecifiedLocation () {
@@ -162,7 +162,7 @@ class VisualStudioFinder {
   }
 
   async findVisualStudio2019OrNewerUsingSetupModule () {
-    return this.findNewVSUsingSetupModule([2019, 2022, 2025])
+    return this.findNewVSUsingSetupModule([2019, 2022])
   }
 
   async findVisualStudio2017UsingSetupModule () {
@@ -223,7 +223,7 @@ class VisualStudioFinder {
   // Invoke the PowerShell script to get information about Visual Studio 2019
   // or newer installations
   async findVisualStudio2019OrNewer () {
-    return this.findNewVS([2019, 2022, 2025])
+    return this.findNewVS([2019, 2022])
   }
 
   // Invoke the PowerShell script to get information about Visual Studio 2017
--- a/lib/find-visualstudio.js
+++ b/lib/find-visualstudio.js
@@ -385,12 +385,9 @@ class VisualStudioFinder {
       ret.versionYear = 2019
       return ret
     }
-    if (ret.versionMajor === 17) {
+    // Fix this when Visual Studio 2025 ships
+    if ([17, 18].includes(ret.versionMajor)) {
       ret.versionYear = 2022
-      return ret
-    }
-    if (ret.versionMajor === 18) {
-      ret.versionYear = 2025
       return ret
     }
     this.log.silly('- unsupported version:', ret.versionMajor)
@@ -461,7 +458,7 @@ class VisualStudioFinder {
     } else if (versionYear === 2022) {
       return 'v143'
     } else if (versionYear === 2025) {
-      return 'v144'
+      return 'v143'  // Fix this when Visual Studio 2025 ships
     }
     this.log.silly('- invalid versionYear:', versionYear)
     return null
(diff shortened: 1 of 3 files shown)


lib/find-visualstudio.js
```diff
@@ -389,6 +389,10 @@ class VisualStudioFinder {
       ret.versionYear = 2022
       return ret
     }
+    if (ret.versionMajor === 18) {
+      ret.versionYear = 2025
```
cclauss, line 393: Is this correct?!?

lib/find-visualstudio.js
```diff
@@ -456,6 +460,8 @@ class VisualStudioFinder {
       return 'v142'
     } else if (versionYear === 2022) {
       return 'v143'
+    } else if (versionYear === 2025) {
+      return 'v144'
```
cclauss, line 464: Is this correct?!?

lib/find-visualstudio.js
legendecas: Is Visual Studio 2025 published yet? I don't think VS version is aligned with Windows version. AFAICT, VS 2025 should be around November of 2025 to December of 2025.

cclauss: You are correct.  Windows 2025 currently ships with Visual Studio 2022.

`vswhere -latest -property installationPath`
```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise
```

lib/find-visualstudio.js
```diff
@@ -456,6 +457,8 @@ class VisualStudioFinder {
       return 'v142'
     } else if (versionYear === 2022) {
       return 'v143'
+    } else if (versionYear === 2025) {
+      return 'v143'  // Fix this when Visual Studio 2025 ships
```
legendecas, line 461: Is these VS 2025 changes required now? Shall we add these when VS 2025 is actually released?

cclauss: That is probably a question for @StefanStojanovic.  I am not a Windows user but if Windows 2025 is available on GitHub Action, it will not be long before new issues are opened about using it.  The great news is that it seems to work out-of-the box and in some tests is quite a bit faster.  The only incompatibility appears in our integration tests because we are still many months away from having a Visual Studio 2025.

StefanStojanovic: The way I see it, Windows and VS versions are not connected. We are combining them for our tests (Win2019 + VS2019 and Win2022 + VS2022), but that is not mandatory.

@cclauss can you please explain this part of the comment: _**These tests will fail on Windows 2025 without these mods because Windows 2025 currently ships with Visual Studio 2022.**_ Does that mean that only changing the `yml` files results in test failures?

cclauss: If you make the changes to the YAML files but do not make the changes to `lib/find-visualstudio.js` then the Windows 2025 tests will fail.

cclauss: > Windows and VS versions are not connected.

My sense is that `lib/find-visualstudio.js` assumes that they are connected.

lib/find-visualstudio.js
```diff
@@ -385,7 +385,8 @@ class VisualStudioFinder {
       ret.versionYear = 2019
       return ret
     }
-    if (ret.versionMajor === 17) {
+    // Fix this when Visual Studio 2025 ships
+    if ([17, 18].includes(ret.versionMajor)) {
```
cclauss, line 389:
```suggestion
if (ret.versionMajor === 17) {
```

lib/find-visualstudio.js
cclauss, line 461:
```suggestion
```

lib/find-visualstudio.js
```diff
@@ -385,7 +385,7 @@ class VisualStudioFinder {
       ret.versionYear = 2019
       return ret
     }
-    if (ret.versionMajor === 17) {
+if (ret.versionMajor === 17) {
```
cclauss, line 388:
```suggestion
    if (ret.versionMajor === 17) {
```

lib/find-visualstudio.js
```diff
@@ -119,7 +119,7 @@ class VisualStudioFinder {
   }
 
   async findVisualStudio2019OrNewerFromSpecifiedLocation () {
-    return this.findVSFromSpecifiedLocation([2019, 2022])
+    return this.findVSFromSpecifiedLocation([2019, 2022, 2025])
```
cclauss, line 122:
```suggestion
    return this.findVSFromSpecifiedLocation([2019, 2022])
```

lib/find-visualstudio.js
```diff
@@ -162,7 +162,7 @@ class VisualStudioFinder {
   }
 
   async findVisualStudio2019OrNewerUsingSetupModule () {
-    return this.findNewVSUsingSetupModule([2019, 2022])
+    return this.findNewVSUsingSetupModule([2019, 2022, 2025])
```
cclauss, line 165:
```suggestion
    return this.findNewVSUsingSetupModule([2019, 2022])
```

lib/find-visualstudio.js
```diff
@@ -223,7 +223,7 @@ class VisualStudioFinder {
   // Invoke the PowerShell script to get information about Visual Studio 2019
   // or newer installations
   async findVisualStudio2019OrNewer () {
-    return this.findNewVS([2019, 2022])
+    return this.findNewVS([2019, 2022, 2025])
```
cclauss, line 226:
```suggestion
    return this.findNewVS([2019, 2022])
```

lib/find-visualstudio.js
cclauss: Nice!  Thanks @legendecas you are correct.

Issue #3126, opened 2025-02-12
Build failure on Windows in `node-gyp@11.1.0` due to `GetModuleHandle` LPCWSTR conversion issue


BohdanVV:
## Description

After upgrading to `node-gyp@11.1.0`, I encountered a build failure on Windows due to an incorrect function call in `win_delay_load_hook.cc`. The issue is caused by `GetModuleHandle`, which expects a wide string (`LPCWSTR`) but is receiving a narrow string (`char*`).

This happens on both local hardware and GitHub runners.

## Error Message

> C:\Users\myuser\AppData\Local\Yarn\Data\global\node_modules\node-gyp\src\win_delay_load_hook.cc(32,36): error C2664: 'HMODULE GetModuleHandleW(LPCWSTR)': cannot convert argument 1 from 'const char [12]' to 'LPCWSTR'

## Additional Context  
This issue appears to be introduced by changes in **#2834**, which were merged and released as part of `node-gyp@11.1.0`

Detailed logs (GitHub Actions run): [logs from the failed build](https://github.com/51Degrees/device-detection-node/actions/runs/13276740370/job/37067647818#step:5:338)

## Affected Versions
- **node-gyp:** `11.1.0`
- **Node.js:** Any, I've tested w/ `18.x`, `20.x`, `22.x`
- **OS:** Windows 10 (x64), Windows GitHub runners

## Proposed Fix  
I'm not a C++ expert, but I attempted a modification in `win_delay_load_hook.cc` to explicitly use `GetModuleHandleA` (ANSI) instead of `GetModuleHandle`, which defaults to `GetModuleHandleW` (wide-character version).

**Change this:**
```cpp
m = GetModuleHandle("libnode.dll");
```
to this:
```cpp
m = GetModuleHandleA("libnode.dll");
```

If this sounds like a reasonable solution, I can open a PR.

Would appreciate any comments from the maintainers.

@zombieyang

PETEPEtrek:
Happened to me when I tried to build vscode , thanks!

cclauss:
Would it be possible to create a GitHub Action that fails because of this issue?

ScubaDrew:
Use `GetModuleHandleA` fix for me too. Thanks.

ymeine:
[...]

PR #2834, opened 2023-04-26
try libnode.dll first in load_exe_hook

zombieyang:
##### Checklist

- [x] `npm install && npm test` passes
- [ ] tests are included 
- [x] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change

It is a following PR of https://github.com/nodejs/node/issues/47682.

I've tried to rename my executable to `node.exe` but the error is still there.
That's because the node.js's symbol is in `libnode.dll` but not in my executable. So, `GetModuleHandle(NULL)` will not work.

However, I think we can try to load `libnode.dll` first. If failed then fallback to try the executable. That will solve the problem and keep the original path work either.
Maybe it is looks like a edge case, but as Node.js has a [embedding tutorial](https://nodejs.org/dist/latest-v18.x/docs/api/embedding.html) in the doc, I think it's worth to do this.

Or please tell me if there is a more simple way to solve my problem.

------
Test is not included because I need to provide a huge nodedll binary.

Thank you, I love node.js.

PR #3139, opened 2025-03-05
win,src: fix GetModuleHandle for libnode.dll

StefanStojanovic:
Based on the description in the linked issue, this should fix the problem. There is no strong reason for making a change like this (and for example not using `GetModuleHandleA` or `L"libnode.dll"`). This is the only `GetModuleHandle` call in the entire project, so if someone thinks other approaches are better, let me know and I'll change it.

Fixes: https://github.com/nodejs/node-gyp/issues/3126
Refs: https://github.com/nodejs/node-gyp/pull/2834

### DIFF
--- a/src/win_delay_load_hook.cc
+++ b/src/win_delay_load_hook.cc
@@ -29,7 +29,7 @@ static FARPROC WINAPI load_exe_hook(unsigned int event, DelayLoadInfo* info) {
     return NULL;
 
   // try for libnode.dll to compat node.js that using 'vcbuild.bat dll'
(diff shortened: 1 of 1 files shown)


StefanStojanovic:
I've tried few different variations of a fix, and even running the action without any changes. Windows test jobs always to fail. Because of that, I do not think failures are caused by my changes.

lukekarrys (approved):

lukekarrys:
I think the failing Windows tests in this PR were fixed by https://github.com/nodejs/node-gyp/pull/3145. I think this can land.

Issue #3144, opened 2025-03-19
Wasm tests on Windows are failing on gyp-next and this repo


cclauss:
I don't have a good solution to fix the wasm test right now. I think we should disable it before we have a proper fix.

/cc author of the tests, @toyobayashi I'd appreciate it if you could help us fixing the test, thanks!

_Originally posted by @legendecas in https://github.com/nodejs/node-gyp/issues/3143#issuecomment-2736314227_

* nodejs/gyp-next#287
* https://github.com/nodejs/gyp-next/pull/286#issuecomment-2733199010

@nodejs/wasi @nodejs/wasi

toyobayashi:
The quickest solution is to install wasi-sdk in CI and then modify the path of clang.exe in the test code. If you agree to introduce additional dependencies, I can try to fix it. If you don't want to introduce additional dependencies, I think it's okay to skip the wasm test. I will open issue again if I encounter problems in the future. @legendecas  @cclauss

cclauss:
Great.  Please try first on the gyp-next repo and then we can upgrade this repo from that one.

legendecas:
The [test](https://github.com/nodejs/node-gyp/blob/main/test/test-windows-make.js#L77) is in this repo, not in gyp-next (gyp-next is failing because we run node-gyp tests in it).

cclauss:
Confusing relationship between these two repos but OK.

PR #3145, opened 2025-03-19
test: fix wasm test on Windows

toyobayashi:
fixes https://github.com/nodejs/node-gyp/issues/3144

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,6 +109,10 @@ jobs:
             node: 23.x
     name: ${{ matrix.os }} - ${{ matrix.python }} - ${{ matrix.node }}
     runs-on: ${{ matrix.os }}
+    env:
+      WASI_VERSION: '25'
+      WASI_VERSION_FULL: '25.0'
+      WASI_SDK_PATH: 'wasi-sdk-25.0'
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


test/test-windows-make.js
```diff
@@ -72,14 +74,15 @@ function quote (path) {
...
   }
+  return path
 }
 
 describe('windows-cross-compile', function () {
   it('build simple node-api addon', async function () {
     if (process.platform !== 'win32') {
       return this.skip('This test is only for windows')
     }
-    const env = getEnv('win-clang')
```
legendecas, line 82: Can we skip the test if wasi sdk is not installed? Otherwise we will need to install wasi-sdk for gyp-next repo as well.

toyobayashi: It will skip if `WASI_SDK_PATH` not exist, so gyp-next no need to change.

https://github.com/nodejs/node-gyp/blob/c0b57d2883dd3eabc0886b1fa0597e9b722a81bb/test/test-windows-make.js#L54
https://github.com/nodejs/node-gyp/blob/c0b57d2883dd3eabc0886b1fa0597e9b722a81bb/test/test-windows-make.js#L85-L88

.github/workflows/tests.yml
```diff
@@ -123,6 +125,13 @@ jobs:
         env:
           PYTHON_VERSION: ${{ matrix.python }}  # Why do this?
       - uses: seanmiddleditch/gha-setup-ninja@v6
+      - name: Install wasi-sdk (Windows)
+        shell: pwsh
+        if: ${{ startsWith(matrix.os, 'windows') }}
```
cclauss, line 130:
```suggestion
        if: runner.os == 'Windows'
```
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#detecting-the-operating-system

test/test-windows-make.js
```diff
@@ -72,14 +74,15 @@ function quote (path) {
   if (path.includes(' ')) {
     return `"${path}"`
   }
+  return path
 }
 
 describe('windows-cross-compile', function () {
   it('build simple node-api addon', async function () {
     if (process.platform !== 'win32') {
       return this.skip('This test is only for windows')
```
cclauss, line 83:
```suggestion
      return this.skip('This test is only for Windows')
```

cclauss:
`Tests / Python Lint` failing can be safely ignored.

legendecas (approved):
Thank you

PR #3159, opened 2025-05-02
feat(test): adopt `node:test`

avivkeller:
##### Checklist

- [x] tests are included 

##### Description of change

Since v18, Node.js provides the `node:test` module. Recently, we switched https://nodejs.org to use it, and I believe it would be ideal for other repositories in the organization to follow-suit, after-all, it's our test runner, we should use it.


avivkeller:
🤔 On my local windows machine, all tests pass

avivkeller:
Just need to fix the globs

avivkeller:
https://github.com/nodejs/node-gyp/actions/runs/14804181563/job/41569437842?pr=3159#step:11:256

legendecas:
Why rename the test directory?

avivkeller:
> Why rename the test directory?

Because in v20, the test runner doesn't have glob support, and will (by default) run every file in the "test" directory. We only want it to run files prefixed with "test-", so the directory was renamed. 

I can probably undo that with a `run()` test API call instead.

cclauss:
Git conflicts.

oga35767-eng (approved):

oga35767-eng:
👍

Issue #3192, opened 2025-08-10
Node-gyp config instructions are going to stop working soon


CamJN:
Running `npm i` today with config set per https://www.npmjs.com/package/node-gyp#user-content-configuration, you will get the following warning:

`npm warn Unknown user config "devdir". This will stop working in the next major version of npm.`

Googling will lead you to https://github.com/npm/cli/issues/8153 where the npm folks are very adamant that they do not want global node package installation related configuration in your .npmrc anymore, and will break it on purpose soon.

Switching to an env var, the only other method of configuring node-gyp in the docs gives a similar warning:

`npm warn Unknown env config "devdir". This will stop working in the next major version of npm.`

So node-gyp needs a new place for config that npm cannot break.
* npm/cli#8153

lukekarrys:
Closing as a duplicate of ~~#3156~~ #3163

Issue #3163, opened 2025-05-19
Alternative environment variable due to warning on `npm_config_nodedir`


jellelicht:
See title.
Somewhere since `npm@11.x.y`, `npm` now warns on seeing 'unknown' configuration options.
Setting `nodedir` in your npmrc (or setting the environment variable `npm_config_nodedir`) gives us a warning that `nodedir` is not supported, and will be an error Real Soon (tm):

```
 npm warn Unknown user config "nodedir". This will stop working in the next major version of npm.
```

Provided I want to set up the nodedir for scripts that end up calling scripts than end up calling scripts than run `npm` [^1], how will I be able to set up a simple environment variable that allow node-gyp to find the relevant node files in the future when the current approach will no longer work?

[^1]: So I don't have control over the actual invocation of the `npm`, and therefore can't simply pass an option.

lukekarrys:
This is a good question and one we will need to figure out how to handle in `node-gyp`. I'm going to close this as a duplicate of #3156 so we can consolidate future discussions to a single issue.

CamJN:
This needs to be reopened as #3156 is closed now and #3196 did not address this.

lukekarrys:
Here's how you would accomplish this now that #3196 is merged. I'm running a local copy of node-gyp and grepping the output on stderr to confirm that it is picking up the new environment variable.

```bash
> export npm_package_config_node_gyp_nodedir=/MY_NODEDIR

> ./bin/node-gyp.js configure --loglevel=verbose 2>&1 | grep 'get node dir'
gyp verb get node dir compiling against specified --nodedir dev files: /MY_NODEDIR
```

Also I've confirmed that exporting this environment variable does not show any warnings from `npm`:

```bash
> export npm_package_config_node_gyp_nodedir=/MY_NODEDIR

> npm --version
11.5.2

> export npm_config_nodedir=/FROM_NPM

> npm --version
npm warn Unknown env config "nodedir". This will stop working in the next major version of npm.
11.5.2
```

@CamJN Can you write an example for your use-case showing what is not achievable with this new functionality? I can't figure out what I might be missing.

CamJN:
Ah my bad, I misread. I set nodedir instead of devdir and thought the env vars weren't working.

lukekarrys:
[...]

PR #3196, opened 2025-08-14
feat: support reading config from package.json

### LINKED ISSUE #3156: Unknown env config "nodegit-binary-host-mirror".  This will stop working in the next major version of npm.
Hi,

I'm using [nodegit](https://github.com/nodegit/nodegit), which uses node-gyp to include the platform specific binaries.  In my organization, I have in my .npmrc, an override for `nodegit_binary_host_mirror` to set the location to point to my custom repository.  Using npm v11.3.0, I started getting `Unknown env config "nodegit-binary-host-mirror".  This will stop working in the next major version of npm.`  My understanding is this is a feature of node-gyp to allow for overriding the binary location.  Is there an alternative being discussed?

Thanks,
Rob

lukekarrys:
`npm@11` has deprecated setting arbitrary config values (https://github.com/npm/cli/issues/8153), which was the previously recommended place for users to set node-gyp related config. The new recommendation from npm is to use the `config` object in package.json for this (https://github.com/npm/cli/issues/8153#issuecomment-2718937461). I think it makes sense to follow npm's recommendation here.

When a user sets a value in `package.json#config` it will be set in the environment by npm with the prefix `npm_package_config_<KEY>`.

I think it makes sense to allow prescribe that they keys be prefixed with `node_gyp_`. So as an example:

**package.json**
```json
{
  "config": {
    "node_gyp_devdir": "/tmp/.gyp"
  }
}
```

Then in `node-gyp` we can read the env var `npm_package_config_node_gyp_devdir`. If a user wants to set a global value, they can set that env var on their system.

Fixes #3156

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,9 +235,24 @@ Some additional resources for Node.js native addons and writing `gyp` configurat
 
 ## Configuration
 
+### package.json
+
+Use the `config` object in your package.json with each key in the form `node_gyp_OPTION_NAME`. Any of the command
+options listed above can be set (dashes in option names should be replaced by underscores).
+
+For example, to set `devdir` equal to `/tmp/.gyp`, your package.json would contain this:
+
+```json
+{
+  "config": {
+    "node_gyp_devdir": "/tmp/.gyp"
+  }
+}
+```
+
 ### Environment variables
 
-Use the form `npm_config_OPTION_NAME` for any of the command options listed
+Use the form `npm_package_config_node_gyp_OPTION_NAME` for any of the command options listed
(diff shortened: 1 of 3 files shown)


lukekarrys:
TODO:

- [x] add tests

legendecas (approved):

lukekarrys:
This PR should have been linked to fixing #3163 and #3192 (not #3156). Not sure how to update this in the changelog now that this has been merged though.

wraithgar:
I'm updating `node-gyp` in npm right now so the next version of `v11` will have support for this!

PR #3216, opened 2025-09-25
fix(ci): Test on Python 3.14 release candidate 3 on Linux and macOS

cclauss:
Python v3.14 -- October 7th
* https://www.python.org/download/pre-releases
* https://www.python.org/downloads/release/python-3140rc3

Node.js using Python 3.14 on Windows is blocked by:
* nodejs/node#59983

---
Also, upgrade from deprecated `macos-13` to `macos-15-intel`.

The currently available GitHub Actions macOS runners are:
| macOS Version | runner.arch |
|---------------|-------------|
| macos-13 | X64 |
| macos-14 | ARM64 |
| macos-15 | ARM64 |
| macos-15-intel | X64 |
| macos-26 | ARM64 |
| macos-latest | ARM64 |
* Let's prepare for actions/runner-images#13046

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -93,18 +93,22 @@ jobs:
       fail-fast: false
       max-parallel: 11
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
-        python: ["3.9", "3.11", "3.13"]
+        # Windows on Python 3.14 is blocked by nodejs/node#59983
+        os: [macos-latest, ubuntu-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -123,6 +127,7 @@ jobs:
         uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python }}
+          allow-prereleases: true
```
legendecas, line 130: Should we wait until Oct 7 for the official release?

PR #3233, opened 2025-11-07
chore: update devDependencies

lukekarrys:


lukekarrys:
Failing test is unrelated to the devDeps change and is a flaky test due to timeouts being hit. I should increase the timeout in another commit though.

PR #3255, opened 2025-12-22
fix: cpu concurrency detection on some platforms

### LINKED ISSUE #3191: Build fails due to `cpus().length === 0`
I want to install `isolated-vm`, but when I try to install it, it wants to build. When it tries that, it fails, because `make` requires a positive integer argument.

```
make: the '-j' option requires a positive integer argument
```
https://github.com/nodejs/node-gyp/blob/02f747f13b8fea73416ca8e302801f4a3f20a5f6/lib/build.js#L181

Maybe it would be a good fallback to use `1` in case the length is `0`?

```javascript
require('os').cpus().length || 1
```

<details>
<summary>Additional config</summary>

To get this error on an android device using termux, you have to install `nodejs`, `ndk-sysroot`, `make`, `clang`, `binutils`. Also, a `~/.gyp/include.gypi` file has to be created with the following contents:

```javascript
{
  'variables': {
    'android_ndk_path': ''
  }
}
```
</details>

thunder-coding:
`os.availableParallelism()` should produce more accurate results of how
much parallelism should be used then `os.cpus().length`. The only
breaking change is the newer API requires node versions >= v18.x which
is fine as node-gyp only aims to support latest and LTS releases. And
the oldest LTS release still supported as of this commit is v22.

Fixes #3191

This should improve Android support.

##### Checklist

- [ ] tests are included 

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -163,7 +163,7 @@ async function build (gyp, argv) {
         if (!isNaN(j) && j > 0) {
           argv.push('/m:' + j)
         } else if (jobs.toUpperCase() === 'MAX') {
-          argv.push('/m:' + require('os').cpus().length)
+          argv.push('/m:' + require('os').availableParallelism())
         }
       }
     } else {
@@ -178,7 +178,7 @@ async function build (gyp, argv) {
           argv.push(j)
         } else if (jobs.toUpperCase() === 'MAX') {
           argv.push('--jobs')
-          argv.push(require('os').cpus().length)
+          argv.push(require('os').availableParallelism())
         }
       }
     }


cclauss:
Are there docs that explain how the CPU count will change on Android?

thunder-coding:
No, no straightforward docs, but this should give you an explaination.

From [API docs for os.cpus()](https://nodejs.org/docs/latest/api/os.html#oscpus)

> os.cpus().length should not be used to calculate the amount of parallelism available to an application. Use [os.availableParallelism()](https://nodejs.org/docs/latest/api/os.html#osavailableparallelism) for this purpose.

Looking at source tree of Node.js:

- [`os.cpus()` relies on exposed `getCPUs()` internal binding](https://github.com/nodejs/node/blob/9120924de1878e61d65eaf2d5d0d27cbe129224e/lib/os.js#L143)
- [Native C++ code for `getCPUs()` is implemented in another C++ routine](https://github.com/nodejs/node/blob/9120924de1878e61d65eaf2d5d0d27cbe129224e/src/node_os.cc#L437)
- [This internally relies on libuv's `uv_get_cpus()`](https://github.com/nodejs/node/blob/9120924de1878e61d65eaf2d5d0d27cbe129224e/src/node_os.cc#L115)
- [Implementation of libuv's `uv_get_cpus()` reveals that it will error out if `/proc/stat` is not readible. In case of Android `/proc/stat` is restricted in newer Android versions](https://github.com/nodejs/node/blob/9120924de1878e61d65eaf2d5d0d27cbe129224e/deps/uv/src/unix/linux.c#L1789)

So on newer Android versions `os.cpus()` returns an empty array. `os.cpus().length` evaluates to `0`. `os.getParallelism()` on the other hand gives an accurate answer.

I believe that the docs specifying that `os.cpus()` not being a reliable way to get number of parallelism is a enough answer for this change. `os.getParallelism()` does solve this problem effectively

If you want I can include these details in the commit message as well (I didn't as I thought it would make the commit message bit too long)

cclauss (approved):

Issue #3264, opened 2026-01-09
Outdated Python 3.14 manual install instructions on Windows


MikeMcC399:
## Situation

The [README > On Windows](https://github.com/nodejs/node-gyp/blob/main/README.md#on-windows) section says:

>   * Install the current [version of Python](https://devguide.python.org/versions/) from the [Microsoft Store](https://apps.microsoft.com/store/search?publisher=Python+Software+Foundation).

The current version of Python (highest version not pre-release) is Python 3.14. This was released on  2025-10-07.

The [Microsoft Store](https://apps.microsoft.com/store/search?publisher=Python+Software+Foundation) offers:

- Python 3.12
- Python 3.13
- Python Install Manager

<img width="948" height="906" alt="Image" src="https://github.com/user-attachments/assets/3563f6d0-ddaa-44ea-98ef-c6a49ab9b4bd" />

Note that Python 3.14 is not offered in a stand-alone package, instead the Python Install Manager notes in the "What's new in this version" section "This update bundles 3.14.2 as the default version."

### Python 3.13

The [Python 3.13 documentation](https://docs.python.org/3.13/using/windows.html) says:

> The [Microsoft Store package](https://docs.python.org/3.13/using/windows.html#windows-store) is a simple installation of Python that is suitable for running scripts and packages, and using IDLE or other development environments. It requires Windows 10 and above, but can be safely installed without corrupting other programs. It also provides many convenient commands for launching Python and its tools.

### Python 3.14

[Python 3.14 documentation](https://docs.python.org/3/using/windows.html) says:

> To obtain Python from the CPython team, use the Python Install Manager. This is a standalone tool that makes Python available as global commands on your Windows machine, integrates with the system, and supports updates over time. You can download the Python Install Manager from [python.org/downloads](https://www.python.org/downloads/) or through the [Microsoft Store app](https://apps.microsoft.com/detail/9NQ7512CXL7T).

## Suggestion

Modify the node-gyp instructions to correspond to the Python 3.14 documentation.

I suggest to use the link https://docs.python.org/3/using/windows.html to refer to the live Python 3.x (latest) documentation.

MikeMcC399:
- PR #3265 submitted to resolve this issue

PR #3265, opened 2026-01-09
readme: update Python manual install instructions for Windows

MikeMcC399:
- closes https://github.com/nodejs/node-gyp/issues/3264

## Checklist

### Situation

Python instructions to install on Windows changed with the release of Python 3.14 in October 2025.

[Python 3.14 documentation](https://docs.python.org/3/using/windows.html) says:

> To obtain Python from the CPython team, use the Python Install Manager. This is a standalone tool that makes Python available as global commands on your Windows machine, integrates with the system, and supports updates over time. You can download the Python Install Manager from [python.org/downloads](https://www.python.org/downloads/) or through the [Microsoft Store app](https://apps.microsoft.com/detail/9NQ7512CXL7T).

Python 3.14 is not available as a separate package in the [Microsoft Store](https://apps.microsoft.com/store/search?publisher=Python+Software+Foundation), unlike Python 3.12 and 3.13.

The Python Install Manager can be used to install 3.14 and other versions.

### Description of change

In the [README > On Windows](https://github.com/nodejs/node-gyp/blob/main/README.md#on-windows) section change the manual install instructions to:

> Follow the instructions in [Using Python on Windows](https://docs.python.org/3/using/windows.html) to install the current [version of Python](https://www.python.org/downloads/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,8 @@ choco install python visualstudio2022-workload-vctools -y
 
 Or install and configure Python and Visual Studio tools manually:
 
-  * Install the current [version of Python](https://devguide.python.org/versions/) from the
-  [Microsoft Store](https://apps.microsoft.com/store/search?publisher=Python+Software+Foundation).
+  * Follow the instructions in [Using Python on Windows](https://docs.python.org/3/using/windows.html) to install
+    the current [version of Python](https://www.python.org/downloads/).
 
    * Install Visual C++ Build Environment: For Visual Studio 2019 or later, use the `Desktop development with C++` workload from [Visual Studio Community](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=Community).  For a version older than Visual Studio 2019, install [Visual Studio Build Tools](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools) with the `Visual C++ build tools` option.
 


MikeMcC399:
Preview

https://github.com/MikeMcC399/node-gyp/blob/python-install-windows-doc/README.md#on-windows

cclauss:
Alternatively we could point to the ___Windows install manager___ at:
* https://www.python.org/downloads/windows

MikeMcC399:
@cclauss 

> Alternatively we could point to the _**Windows install manager**_ at:
> 
> * https://www.python.org/downloads/windows

Pointing to the documentation https://docs.python.org/3/using/windows.html gives more comprehensive information, including different choices of where to get the Python install manager for Windows and how to use the Python install manager.

When I started to migrate to Python 3.14 I found that I needed this documentation.

If you want to point instead to the download page, https://www.python.org/downloads (one level higher) presents simpler information. The contents depends on the operating system where the browser is running and from Windows it looks like this:

image

this is easier on the eyes compared to https://www.python.org/downloads/windows

https://www.python.org/downloads/windows

image

---

Another possibility would be to use https://www.python.org/downloads and include a link to https://docs.python.org/3/using/windows.html

That may be the best way.

Would be great to hear also from @StefanStojanovic about his thoughts!

MikeMcC399:
My original (short) suggestion, which delegates the instructions to the Python docs, is:

> Follow the instructions in [Using Python on Windows](https://docs.python.org/3/using/windows.html) to install the current [version of Python](https://www.python.org/downloads/).

An alternate longer suggestion, which links directly to the download location options would be:

> To install the current version of Python use the Python Install Manager, which can be installed from the [Microsoft Store app](https://apps.microsoft.com/detail/9NQ7512CXL7T) or downloaded and installed from [python/downloads](https://www.python.org/downloads/). Refer to the documentation [Using Python on Windows](https://docs.python.org/3/using/windows.html) for Python Install Manager usage instructions.

I prefer the shorter version.

StefanStojanovic:
Generally speaking, I'm more in favor of shorter docs, as long as it gives all the info you need.

MikeMcC399:
Is `readme:` all right for the commit message? I took it from [release-please-config.json](https://github.com/nodejs/node-gyp/blob/main/release-please-config.json) or would `doc:` or `docs:` be preferred?

MikeMcC399:
@StefanStojanovic 

Thanks for your feedback! Are you (or @cclauss) ready to approve this PR? If not, could you advise what I should change?

PR #3170, opened 2025-06-22
fix: use temp dir for tar extraction on all platforms

tmm1:
fixes #3165

##### Checklist

- [ ] `npm install && npm run lint && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [ ] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

PR #3301, opened 2026-04-12
fix: better handle concurrent builds

### LINKED ISSUE #3165: Header extraction is unreliable and results in truncated files
I frequently see corrupted node/electron headers. Sounds similar to #1054, but I'm using `npm`. And I see this on an isolate github actions runner, so it doesn't seem like a concurrency issue either?

```
npm error gyp info using node-gyp@10.1.0
npm error gyp info using node@20.18.2 | linux | x64
npm error gyp info find Python using Python version 3.12.3 found at "/usr/bin/python3"
npm error gyp http GET https://electronjs-mirror.anysphere.workers.dev/headers/v34.5.1/node-v34.5.1-headers.tar.gz
npm error gyp http 200 https://electronjs-mirror.anysphere.workers.dev/headers/v34.5.1/node-v34.5.1-headers.tar.gz
npm error gyp http GET https://electronjs-mirror.anysphere.workers.dev/headers/v34.5.1/SHASUMS[25](https://github.com/anysphere/everysphere/actions/runs/15405526830/job/43347404983#step:14:26)6.txt
npm error gyp http 200 https://electronjs-mirror.anysphere.workers.dev/headers/v34.5.1/SHASUMS256.txt
...
npm error   File "/home/runner/.cache/node-gyp/34.5.1/include/node/common.gypi", line 177
npm error     'msvs_settings': {
npm error                      ^
npm error SyntaxError: '{' was never closed
```

more examples:

```
npm error /home/runner/.cache/node-gyp/34.3.4/include/node/v8-object.h:5: error: unterminated #ifndef
npm error     5 | #ifndef INCLUDE_V8_OBJECT_H_
```

```
npm error gyp: /Users/ec2-user/Library/Caches/node-gyp/20.18.2/common.gypi not found while reading includes of binding.gyp while trying to load binding.gyp
```

```
npm error gyp: /home/AzDevOps/.cache/node-gyp/30.5.1/common.gypi not found while reading includes of binding.gyp while trying to load binding.gyp
```

We see this on both linux and macos, but interestingly never on windows. Perhaps due to this?

https://github.com/nodejs/node-gyp/blob/7d883b5cf4c26e76065201f85b0be36d5ebdcc0e/lib/install.js#L203-L206

I would propose making the separate directory download step the default on all platforms.

dsanders11:
##### Checklist

- [x] tests are included 

##### Description of change

Fixes #3165, which unfortunately was not actually fixed by #3170. I'm not sure #3170 had any effect on the issue, as my original changes in #2846 which were extended to all platforms in #3710 were targeting a specific issue on Windows, and did not aim to fix (or know of) the issue in #3165.

The second commit on this branch adds a test which repros the issue being reported in #3165, namely that header files can be seen as partially written when multiple builds are happening concurrently. A failing test run before the other commits on this branch [can be seen here](https://github.com/dsanders11/node-gyp/actions/runs/24318080696).

The remainder of the commits refactor `copyDirectory` to fix the issue, fix another concurrent build issue I came across during testing, and adds more test coverage. This refactor also allows for dropping `exponential-backoff` as a dependency.

### DIFF
--- a/lib/copy-directory.js
+++ b/lib/copy-directory.js
@@ -1,0 +1,43 @@
+'use strict'
+
+const { promises: fs } = require('graceful-fs')
+const crypto = require('crypto')
+const path = require('path')
+
+const { backOff } = require('exponential-backoff')
+
+async function copyDirectory (src, dest, ensure = false) {
+  try {
+    await fs.stat(src)
+  } catch {
+    throw new Error(`Missing source directory for copy: ${src}`)
+  }
+  await fs.mkdir(dest, { recursive: true })
+  const entries = await fs.readdir(src, { withFileTypes: true })
+  for (const entry of entries) {
+    if (entry.isDirectory()) {
+      await copyDirectory(path.join(src, entry.name), path.join(dest, entry.name))
+    } else if (entry.isFile()) {
+      // with parallel installs, copying files may cause file errors on
+      // Windows so use an exponential backoff to resolve collisions
+      await backOff(async () => {
+        try {
+          await fs.copyFile(path.join(src, entry.name), path.join(dest, entry.name))
+        } catch (err) {
+          // if ensure, check if file already exists and that's good enough
+          if (ensure && err.code === 'EBUSY') {
+            try {
+              await fs.stat(path.join(dest, entry.name))
+              return
+            } catch {}
+          }
+          throw err
+        }
+      })
+    } else {
+      throw new Error('Unexpected file directory entry type')
+    }
+  }
+}
+
+module.exports = copyDirectory
--- a/lib/copy-directory.js
+++ b/lib/copy-directory.js
@@ -4,9 +4,9 @@ const { promises: fs } = require('graceful-fs')
 const crypto = require('crypto')
 const path = require('path')
 
-const { backOff } = require('exponential-backoff')
+const RACE_ERRORS = ['ENOTEMPTY', 'EEXIST', 'EBUSY', 'EPERM']
 
-async function copyDirectory (src, dest, ensure = false) {
+async function copyDirectory (src, dest) {
   try {
     await fs.stat(src)
   } catch {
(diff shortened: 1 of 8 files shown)


test/test-copy-directory.js
```diff
@@ -0,0 +1,80 @@
...
+        polls++
+        if (stat.size !== fileSize) {
+          violations.push({ poll: polls, size: stat.size })
+        }
+      } catch (err) {
+        if (err.code !== 'ENOENT') throw err
+      }
+    }, 0)
+
+    await copyDirectory(srcDir, destDir)
```
legendecas: Would you mind adding a test that `srcDir` is unchanged?

lib/copy-directory.js
```diff
@@ -0,0 +1,58 @@
...
+    // temp path in the dest directory first, then rename within the
+    // same filesystem to keep it atomic.
+    //
+    // When another process wins the race, rename may fail with one of
+    // these codes — all mean the destination was already placed and
+    // are safe to ignore since every process extracts identical content.
+    const srcPath = path.join(src, entry.name)
+    const destPath = path.join(dest, entry.name)
+    try {
+      await fs.rename(srcPath, destPath)
```
legendecas, line 34: This changes the semantics to `moveDirectory` instead of `copyDirectory`.

dsanders11: Happy to rename it and the files. You're right that now it's move, outside of the edge case of cross-device on Windows, which falls back to a copy and move.

test/test-copy-directory.js
dsanders11: I've added a test that `moveDirectory` (renamed it to reflect the changed semantics) retains the contents from `srcDir` moved to `destDir`. Let me know if I misunderstood what you were asking for here and I can add more test coverage.

lib/copy-directory.js
oga35767-eng: 👍

oga35767-eng (approved):

oga35767-eng:
👍

lib/copy-directory.js
```diff
@@ -0,0 +1,58 @@
...
+    } catch (err) {
+      if (RACE_ERRORS.includes(err.code)) {
+        // Another parallel process already placed this entry — ignore
```
legendecas, line 37: I don't think this is a really good idea. This could result in a folder where some of its sub-entries are stale (like on windows it's still been used resulting in EBUSY), but some are new.

A directory already at the destination is not proof the contents match.

Issue #2490, opened 2021-08-30
The generated config.gypi file should inherit from $nodedir/config.gypi instead of process.config


zcbenz:
Currently the `variables` field of the generated `config.gypi` file inherits from `process.config`:

https://github.com/nodejs/node-gyp/blob/5f1a06c50f3b0c3d292f64948f85a004cfcc5c87/lib/configure.js#L101-L103

And the `target_defaults` field does inherit from `process.config` for the reason that the build configurations of the running `node` executable may be different from the target being compiled for:
https://github.com/nodejs/node-gyp/blob/5f1a06c50f3b0c3d292f64948f85a004cfcc5c87/lib/configure.js#L115-L122

However the `variables` field also belongs to build configurations of Node.js, and should be treated the same with `target_defaults`, for example Node.js executable compiled with `v8_enable_pointer_compression=1` can not load modules compiled with `v8_enable_pointer_compression=0`.

This has been causing problems when compiling modules for Electron, because node-gyp is using `process.config.variables` of the running `node` executable for compilation, which are different from the build configurations of Electron.

So the generated `config.gypi` file should not inherit from `process.config` at all, and it should inherit from the `$nodedir/include/node/config.gypi` file instead, which is the actual build configuration of the target.

I can start working out a pull request if the above suggested solution sounds good to the maintainers.

/cc @rvagg

rvagg:
@zcbenz I'm inclined to trust you on this and say go ahead and submit a PR and we'll see if we can figure out what the blast radius of such a change might be. Maybe it's just a semver-major to be safe if we can't find evidence of a major impact.

Any input from other @nodejs/gyp heads would be good head (tbh my mental bandwidth on this is a bit limited at the moment).

PR #3303, opened 2026-04-20
fix: override host-specific variables when reading headers config.gypi

iFwu:
## Background

`getBaseConfigGypi` reads `\$nodedir/include/node/config.gypi` verbatim whenever `--disturl`, `--dist-url`, or `--nodedir` is set (and `--force-process-config` is not). This was introduced in #2497 to let Electron / NW.js builds pick up target build config from custom headers.

The official Node release headers tarball is, however, **a single universal artifact shipped to all platforms**, and its embedded `config.gypi` reflects the build host of the Node release-engineering machine (currently a Linux x64 / GCC box). When a consumer on a different host inherits it verbatim, host-specific fields end up wrong.

## Symptom on macOS arm64

Field-by-field diff between the cached headers' `config.gypi` and `process.config.variables` on darwin/arm64 with Node v24.13.1:

| field | cache | process.config |
|---|---|---|
| host_arch | x64 | arm64 |
| clang | 0 | 1 |
| llvm_version | \"0.0\" | \"16.0\" |
| xcode_version | \<undef\> | \"16.0\" |
| arm_fpu | \<undef\> | \"neon\" |
| gas_version | \"2.38\" | \<undef\> |
| shlib_suffix | \"so.137\" | \"137.dylib\" |

The killer is `clang=0`: Node's `common.gypi` gates `-std=gnu++20` (and several other flags) on `clang==1`. With `clang=0` those branches never fire, C++ compilation falls back to C++03, and any addon using `node-addon-api` immediately fails with:

\`\`\`
node-addon-api/napi.h: error: a space is required between consecutive right angle brackets (use '> >')
node-addon-api/napi.h: error: no template named 'initializer_list' in namespace 'std'
node-addon-api/napi.h: error: unknown type name 'constexpr'
\`\`\`

For comparison, on Linux x64 / Node v24.13.1 the same diff shows **all 7 host-specific fields are identical** between cache and process.config (the cache literally is a Linux x64 build farm fingerprint), so the fix in this PR is a strict no-op there.

## Minimal repro (no Electron required)

\`\`\`bash
mkdir /tmp/repro && cd /tmp/repro
cat > package.json <<'EOF'
{\"name\":\"r\",\"dependencies\":{\"node-addon-api\":\"^8.0.0\"}}
EOF
cat > binding.gyp <<'EOF'
{\"targets\":[{\"target_name\":\"r\",\"sources\":[\"r.cc\"],
  \"include_dirs\":[\"<!@(node -p \\\"require('node-addon-api').include\\\")\"],
  \"dependencies\":[\"<!(node -p \\\"require('node-addon-api').gyp\\\")\"]}]}
EOF
echo '#include <napi.h>' > r.cc
npm install --ignore-scripts

# Works:
npx node-gyp rebuild

# Fails (anywhere disturl is set):
npm_config_disturl=https://npmmirror.com/mirrors/node npx node-gyp rebuild
\`\`\`

## Affected scope

Anyone with \`disturl\` set in \`~/.npmrc\`, project \`.npmrc\`, or as \`npm_config_disturl\` env var. Common cases:

- China mirrors: \`npmmirror.com\`, \`taobao\`
- Self-hosted: Nexus, JFrog, Verdaccio with \`disturl\` rewriting
- Electron projects via \`electron-rebuild\` (the symptom most users hit; reported separately as electron/rebuild#1209)

## Why \`--force-process-config\` is not a real fix

- It throws away the *target build config* that the headers tarball legitimately carries (the entire point of #2497 / #2490).
- For Electron native modules it produces a `node_module_version` mismatch — Electron then refuses to load the addon. See pulsar-edit/ppm#162 (2025-12, reverting `--force-process-config` after exactly this).
- Electron's [own blog post](https://github.com/electron/website/pull/143#discussion_r749594721) restricts the flag to \"old Electron (<=13, <14.2.0, <15.3.0)\".

## Fix

Override only the host-specific variables from `process.config.variables` after parsing the headers' `config.gypi`:

\`\`\`js
const HOST_SPECIFIC_VARIABLES = [
  'host_arch',
  'clang',
  'llvm_version',
  'xcode_version',
  'arm_fpu',
  'gas_version',
  'shlib_suffix'
]
\`\`\`

PR #2497's intent is preserved — target build config (v8 features, bundled vs shared deps, `node_module_version`, etc.) still comes from the headers tarball; only host (build machine) fields are corrected.

## Verification

- [x] All existing tests pass (`npm test`: 108 passing, 6 pending — unchanged from main).
- [x] Two new tests added in `test/test-create-config-gypi.js` covering the override behaviour and `--force-process-config` back-compat, with a new fixture `test/fixtures/nodedir-mismatched-host/` that mimics a Linux x64 build farm headers tarball.
- [x] No-op verified on Linux x64 (all 7 host fields already match cache / process.config).
- [x] Necessary verified on macOS arm64 (all 7 fields mismatched).
- [x] `npm run lint` clean for changed files.

## Open questions for maintainers

1. Is the `HOST_SPECIFIC_VARIABLES` set complete? It was derived from a full diff on macOS arm64; would appreciate a sanity check on Windows arm64 / FreeBSD / AIX.
2. Should an upstream fix in Node release engineering (ship platform-specific headers, or strip host-specific keys before packaging the universal tarball) take priority? That would fix every node-gyp version retroactively, but turnaround is much slower.

## Related

- #2497 / #2490 — original introduction of the code path
- electron/rebuild#1209 — first user report (filed in the wrong repo, no traction)
- pulsar-edit/ppm#162 — `--force-process-config` side-effect on Electron module_version

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lib/create-config-gypi.js
+++ b/lib/create-config-gypi.js
@@ -15,6 +15,40 @@ function parseConfigGypi (config) {
   return JSON.parse(config)
 }
 
+// Variables that describe the build host (the machine running node-gyp), not
+// the target Node binary. The official Node release headers tarball is a
+// single universal artifact whose embedded config.gypi reflects the build
+// farm host (currently Linux x64 / GCC), so when a consumer on a different
+// platform inherits it verbatim via --disturl/--nodedir, these fields end up
+// wrong. The most visible symptom on macOS arm64 is `clang=0` silently
+// dropping the `clang==1` branches in common.gypi (e.g. `-std=gnu++20`),
+// breaking node-addon-api compilation. See PR #2497 for the original code
+// path and electron/rebuild#1209 for the first user report.
+const HOST_SPECIFIC_VARIABLES = [
+  'host_arch',
+  'clang',
+  'llvm_version',
+  'xcode_version',
+  'arm_fpu',
+  'gas_version',
+  'shlib_suffix'
+]
+
+function overrideHostSpecificVariables (config) {
+  if (!config || !config.variables || !process.config || !process.config.variables) {
+    return config
+  }
+  for (const key of HOST_SPECIFIC_VARIABLES) {
+    const value = process.config.variables[key]
+    if (value !== undefined) {
+      config.variables[key] = value
+    } else {
+      delete config.variables[key]
+    }
+  }
+  return config
+}
+
 async function getBaseConfigGypi ({ gyp, nodeDir }) {
   // try reading $nodeDir/include/node/config.gypi first when:
   // 1. --dist-url or --nodedir is specified
@@ -25,7 +59,7 @@ async function getBaseConfigGypi ({ gyp, nodeDir }) {
     try {
       const baseConfigGypiPath = path.resolve(nodeDir, 'include/node/config.gypi')
       const baseConfigGypi = await fs.readFile(baseConfigGypiPath)
-      return parseConfigGypi(baseConfigGypi.toString())
+      return overrideHostSpecificVariables(parseConfigGypi(baseConfigGypi.toString()))
     } catch (err) {
       log.warn('read config.gypi', err.message)
     }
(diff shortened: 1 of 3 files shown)


oga35767-eng (approved):

oga35767-eng:
👍

PR #3345, opened 2026-07-09
test: fix undici 8.7.0 compatibility

legendecas:
##### Checklist

- [x] `npm install && npm run lint && npm test` passes
- [x] tests are included 
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change
Previously undici ProxyAgent/EnvHttpProxyAgent tunneled all proxied requests via CONNECT. As of nodejs/undici#5116 (v8.7.0), plain HTTP requests are forwarded to the proxy in absolute-form (a normal request), and only HTTPS still uses a CONNECT tunnel. The node-gyp test proxy only listened for the connect event, so the forwarded HTTP request got no response and the test hung to the 30s timeout.

PR #3346, opened 2026-07-12
fix: handle double quotes inside single-quoted GYP strings

### LINKED ISSUE #3333: parseConfigGypi fails on GYP files whose single-quoted strings contain double-quote characters (e.g. canvas, 'OS=="win"' conditions)
## Description

`parseConfigGypi` in `lib/create-config-gypi.js` converts single-quoted GYP strings to JSON by doing a bare global substitution:

```js
config = config.replace(/'/g, '"')
```

This breaks whenever a single-quoted string itself contains double-quote characters, which is ubiquitous in GYP condition strings like:

```gyp
['OS=="win"', { … }]
```

After the substitution that becomes:

```
["OS==""win"", { … }]
```

which is invalid JSON and causes `JSON.parse` to throw:

```
SyntaxError: Expected ',' or ']' after array element in JSON at position 31 (line 3 column 12)
```

## Affected versions

Reproduced on **v12.3.0** and **v13.0.0** (latest). The three-step implementation of `parseConfigGypi` has been unchanged since the function was introduced.

## Reproduction

```js
const { parseConfigGypi } = require('node-gyp/lib/create-config-gypi')
// Any real-world binding.gyp that uses condition strings — e.g. canvas@2.11.2
const raw = require('fs').readFileSync('/path/to/canvas/binding.gyp', 'utf8')
parseConfigGypi(raw) // → SyntaxError
```

`canvas@2.11.2` is a good, publicly available repro: install it with `--ignore-scripts`, then call `parseConfigGypi` on its `binding.gyp`.

## Root cause

Step 3 of `parseConfigGypi` is a naive global `'` → `"` replacement. It needs to be a proper tokenizer that converts single-quoted strings to double-quoted JSON strings while **escaping any double-quote characters found inside those strings**.

## Suggested fix

Replace the global regex with a character-by-character pass:

```js
function parseConfigGypi(raw) {
  // 1. Strip # comments
  let src = raw.replace(/#[^\n]*/g, '')
  // 2. Join Python-style multiline string continuations
  src = src.replace(/'\\?\n\s*'/g, '')

  // 3. Tokenise: convert single-quoted strings to double-quoted JSON strings,
  //    escaping any embedded double-quote characters.
  let out = ''
  let i = 0
  while (i < src.length) {
    const ch = src[i]
    if (ch === "'") {
      let s = '"'
      i++
      while (i < src.length && src[i] !== "'") {
        if (src[i] === '"') {
          s += '\\"'
        } else if (src[i] === '\\') {
          s += src[i] + (src[i + 1] || '')
          i++
        } else {
          s += src[i]
        }
        i++
      }
      s += '"'
      out += s
      i++ // skip closing `'`
    } else if (ch === '"') {
      // Copy already-valid double-quoted strings verbatim
      out += ch
      i++
      while (i < src.length && src[i] !== '"') {
        if (src[i] === '\\') { out += src[i] + (src[i + 1] || ''); i += 2 }
        else { out += src[i]; i++ }
      }
      out += src[i] || ''
      i++
    } else {
      out += ch
      i++
    }
  }

  // 4. Strip trailing commas (also invalid in JSON, common in GYP)
  out = out.replace(/,(\s*[}\]])/g, '$1')
  return JSON.parse(out)
}
```

bitpshr:
Fixes #3333.

`parseConfigGypi()` converted single quotes to double quotes with a bare `config.replace(/'/g, '"')`, which produced invalid JSON whenever a single-quoted string contained `"` (e.g. GYP condition strings like `'OS=="win"'`, used by native addons such as canvas). It now converts each single-quoted string individually and escapes the double quotes inside it. Added a regression test.

### DIFF
--- a/lib/create-config-gypi.js
+++ b/lib/create-config-gypi.js
@@ -10,8 +10,13 @@ function parseConfigGypi (config) {
   config = config.replace(/#.*/g, '')
   // 2. join multiline strings
   config = config.replace(/'$\s+'/mg, '')
-  // 3. normalize string literals from ' into "
-  config = config.replace(/'/g, '"')
+  // 3. normalize string literals from ' into ", escaping any double-quote
+  // characters that appear inside a single-quoted string (e.g. GYP condition
+  // strings such as 'OS=="win"') so the result is still valid JSON
+  config = config.replace(/'(?:[^'\\]|\\.)*'/g, (str) => {
+    const inner = str.slice(1, -1).replace(/\\'/g, "'").replace(/"/g, '\\"')
+    return `"${inner}"`
+  })
   return JSON.parse(config)
 }
 
--- a/test/test-create-config-gypi.js
+++ b/test/test-create-config-gypi.js
@@ -76,4 +76,10 @@ describe('create-config-gypi', function () {
     const config = parseConfigGypi(str)
     assert.deepStrictEqual(config, { variables: { multiline: 'AB' } })
   })
+
+  it('config.gypi parsing with double quotes inside single-quoted strings', function () {
+    const str = '{\'variables\': {\'cond\': \'OS=="win"\', \'flags\': \'-DFOO="x"\'}}'
+    const config = parseConfigGypi(str)
+    assert.deepStrictEqual(config, { variables: { cond: 'OS=="win"', flags: '-DFOO="x"' } })
+  })
 })


bitpshr:
Rebased this onto main to pick up the undici 8.7.0 fix from #3345, which looks like what was causing the Windows + Node 26 `test-download` timeouts here (this PR itself is just GYP string escaping, unrelated to downloads). The create-config-gypi tests pass locally.

@legendecas the CI run is sitting in `action_required` since it's a fork PR, so it needs an approval to actually run. Whenever you get a chance.

legendecas (changes requested):
In real world, `config.gypi` is actually a valid JSON since https://github.com/nodejs/node/pull/60794, see: https://github.com/nodejs/node/blob/db3a8d866a8b71a295e2f0ea7176811c3ac918fb/configure.py#L2918-L2920

This `parseConfigGypi` method is a limited subset dedicated for `config.gypi`. I'd prefer to have a real-world reproduction, for example, `node-gyp rebuild` with a `binding.gyp`.

bitpshr:
You're right that config.gypi is valid JSON since #60794, so current headers parse fine. The catch is `parseConfigGypi` runs on `include/node/config.gypi` via `--nodedir`/`--dist-url`, and every node released before #60794 ships the old single-quoted format, which can contain embedded double quotes (exactly what #60794 fixed upstream, per #60703's temporal/rust flags). The original issue's repro was misleading since it fed a `binding.gyp`, but the real input hits it too. A minimal old-format config.gypi reproduces it:

```
'variables': { 'rust_flags': '--cfg feature="temporal"' }
```

Without the change, `parseConfigGypi` produces `"rust_flags": "--cfg feature="temporal""` and `JSON.parse` throws; with it, it parses.

That said, it's a narrow case, and since config.gypi is JSON going forward you may prefer to simplify `parseConfigGypi` rather than harden the legacy path. Happy to go either way, or to close this if you'd rather not carry it.

legendecas:
I'm inclined that the way forward with `config.gypi` parsing is simplify it, and eventually just parse it as a JSON when legacy Node.js versions phases out.

bitpshr:
Sounds good @legendecas. Closing!

PR #3352, opened 2026-08-05
fix: normalize OS to linux when building addons on Android

zetzet7298:
## Problem

When node itself is an Android build (e.g. Termux on Android), `process.config.variables.OS` is `"android"`. node-gyp copies that into the generated `build/config.gypi`, so gyp evaluates `OS == "android"` conditions in node's `common.gypi` as true — including branches that only exist for NDK cross-compiles, e.g.:

```
['OS == "android"', {
  'cflags': [ '-fPIC', '-I<(android_ndk_path)/sources/android/cpufeatures' ],
```

`android_ndk_path` is only defined when building node itself with the NDK, so any local addon build on Termux fails with:

```
gyp: Undefined variable android_ndk_path in binding.gyp while trying to load binding.gyp
```

This hits every npm package with a native addon (node-pty, etc.) installed on Termux.

## Root cause

node-gyp's gyp fork already maps `android` → `linux` in `make.py` ("Keep this legacy behavior for now"), but `create-config-gypi.js` copies `OS: "android"` from `process.config` into `config.gypi` first, which overrides that mapping. A local addon build on Android is a **native** build, not an NDK cross-compile — Android is the host, not the target.

## Fix

In `getCurrentConfigGypi`, normalize `variables.OS` to `"linux"` when `process.platform === "android"` and OS was inherited from `process.config`. Custom headers (`--nodedir` / `--dist-url`) are honored: those may intentionally target Android via the NDK.

## Validation

- `test/test-create-config-gypi.js`: added coverage for the Android normalization path and for `--nodedir` preserving custom config
- Verified on a real Termux device (Android 16, node 26.1.0): `node-pty` now builds and loads; previously failed with the `android_ndk_path` error

### DIFF
--- a/lib/create-config-gypi.js
+++ b/lib/create-config-gypi.js
@@ -47,6 +47,17 @@ async function getCurrentConfigGypi ({ gyp, nodeDir, vsInfo, python }) {
   const defaults = config.target_defaults
   const variables = config.variables
 
+  // When node itself is an Android build (e.g. Termux), process.config
+  // reports OS "android". gyp then enables Android-only branches (like the
+  // one referencing android_ndk_path in node's common.gypi) that only exist
+  // for NDK cross-compiles. Building a local addon on Android is a native
+  // build, not an NDK one, so normalize OS to linux. Honor custom headers
+  // (--nodedir/--dist-url): those may intentionally target Android via NDK.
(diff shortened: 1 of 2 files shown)


oga35767-eng (approved):

oga35767-eng:
👍

PR #3133, opened 2025-02-18
chore: switch to tinyglobby

benmccann:
##### Checklist

- [x] `npm install && npm run lint && npm test` passes
- [x] tests are included 
- [x] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change

https://npmgraph.js.org/?q=glob - 26 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies

PR #3353, opened 2026-08-09
refactor: replace tar extraction with modern-tar

ayuhito:
##### Checklist

- [ ] tests are included 

##### Description of change

Replaces `tar` with [`modern-tar`](https://www.npmjs.com/package/modern-tar). Cuts dependencies from 6 --> 1. 4MB on disk --> 95KB on disk.

image

_Ref: https://npmgraph.js.org/?q=node-gyp#select=tar%407.5.22_

Similar to the motivation of #3133. Thanks for everyone's time!

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -1,12 +1,13 @@
 'use strict'
 
-const { createWriteStream, promises: fs } = require('graceful-fs')
+const { createReadStream, createWriteStream, promises: fs } = require('graceful-fs')
 const os = require('os')
 const { backOff } = require('exponential-backoff')
-const tar = require('tar')
+const { unpackTar } = require('modern-tar/fs')
 const path = require('path')
 const { Transform, promises: { pipeline } } = require('stream')
 const crypto = require('crypto')
+const { createGunzip } = require('zlib')
 const log = require('./log')
 const semver = require('semver')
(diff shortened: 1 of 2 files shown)


lib/install.js
```diff
@@ -192,10 +192,11 @@ async function install (gyp, argv) {
       return isValid
     }
 
-    function onwarn (code, message) {
-      extractErrors = true
-      log.error('error while extracting tarball', code, message)
-    }
+    const extract = () => unpackTar(tarExtractDir, {
+      strict: true,
```
ayuhito, line 196: This is the only section I'm somewhat unsure about if it may regress. I don't think the differences are large, but I do want to point it out.

The previous behavior was very strict, relying on the `onwarn` handler to throw. `strict: true` is the same thing, but there still may be minor differences on what our definitions of strict are. e.g. `node-tar` accepts an archive with a single EOF zero block, while `modern-tar` strict mode rejects it as truncated.

This is fine for the normal path from Node.js sources, but might affect `--tarball` and `--dist-url`? How do we feel about that?

oga35767-eng: 👍

oga35767-eng (approved):

oga35767-eng:
👍

PR #3356, opened 2026-08-13
fix: print actionable hints when gyp fails on Android with missing NDK

### LINKED ISSUE #3355: gyp: 'Undefined variable android_ndk_path' when building native modules on Android (Termux) without NDK
### Describe the bug

Building any native module with node-gyp on Android (Termux) fails with an unhelpful gyp error, even though a working toolchain (clang) is installed:

```
gyp: Undefined variable android_ndk_path in binding.gyp while trying to load binding.gyp
```

Environment: Termux on Android 15 (aarch64), Node v26.4.0, npm 12.0.2, clang 21, **no NDK** (Android NDK is not practical to install on Termux; the system provides its own bionic toolchain).

### Root cause chain

1. Termux's python3 reports `sys.platform == "android"`, so gyp selects the `android` flavor.
2. node-gyp does not use the headers shipped with the Node package (`/usr/include/node/common.gypi`, which Termux's packaging already patches for this exact issue); instead it downloads the official Node headers from nodejs.org into `~/.cache/node-gyp/<version>/`.
3. The official `common.gypi` has an `OS == "android"` branch that references `<(android_ndk_path)/sources/android/cpufeatures`. On a plain Android system with no NDK, `android_ndk_path` is undefined and gyp aborts with the message above — it reads as an internal gyp bug rather than a missing-NDK situation.

### Expected behavior

- Either node-gyp should prefer locally installed Node headers when present (Termux's `/usr/include/node` is a valid, already-patched header set), or
- on Android without an NDK, emit an actionable error (e.g. "building native modules on Android requires the Android NDK, or use the system toolchain via `--nodedir=/usr/include/node`") instead of `Undefined variable android_ndk_path`.

### Workaround (used successfully)

Patch `~/.cache/node-gyp/26.4.0/include/node/common.gypi`: in the `OS == "android"` branch, replace `-I<(android_ndk_path)/sources/android/cpufeatures` with the system include dir. After that, node-pty and other native modules compile and link fine against the Termux bionic toolchain.

This affects every Termux user who builds native npm modules (node-pty, sharp, koffi, etc.).

lesliechowsh:
Fixes #3355

When a native module build fails on Android (e.g. Termux) with "gyp: Undefined variable android_ndk_path in binding.gyp", the message looks like an internal gyp bug and nothing tells the user what to do. The variable comes from the android branch of the official Node headers' `common.gypi` (downloaded via `lib/process-release.js`); it is only defined when Node itself is built with an NDK.

This PR captures gyp's stderr in `lib/configure.js` where gyp is spawned (still written through to the terminal byte-for-byte) and, when the failure matches an `android_ndk_*` variable or happens while loading `common.gypi` (see `isNdkUndefinedVariableHint`), prints actionable hints: use an NDK, or point node-gyp at system-toolchain headers via `--nodedir` / a Node build with `--use-prefix-to-find-headers` (termux/termux-packages#30974).

A binding.gyp referencing its own undefined variable is a project bug and does not trigger the hint, so users are not sent down the wrong path. Only the Android failure path is affected.

### DIFF
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -311,9 +311,37 @@ async function configure (gyp, argv) {
     process.env.PYTHONPATH = pypath.join(win ? ';' : ':')
 
     await new Promise((resolve, reject) => {
-      const cp = gyp.spawn(python, argv)
+      // Capture gyp's stderr (still forwarded to the terminal) so we can
+      // recognise known failure modes and print actionable hints.
+      const cp = gyp.spawn(python, argv, { stdio: ['inherit', 'inherit', 'pipe'] })
+      let gypStderr = ''
+      if (cp.stderr) {
+        cp.stderr.on('data', (chunk) => {
+          gypStderr += chunk
+          process.stderr.write(chunk)
+        })
+      }
       cp.on('exit', (code) => {
         if (code !== 0) {
+          // Only hint when the undefined variable comes from the Node/NDK
+          // header toolchain (android_ndk_* variables, or any variable while
+          // loading common.gypi), not when a binding.gyp itself references an
+          // undefined variable — that is a project bug and this hint would
+          // send the user down the wrong path.
+          const ndkHint = process.platform === 'android' &&
+            isNdkUndefinedVariableHint(gypStderr)
+          if (ndkHint) {
+            log.error('gyp', [
+              'The Node headers downloaded for this target reference NDK',
+              'variables, but no NDK is installed. On Android (e.g. Termux)',
+              'you usually want to compile against the system toolchain instead:',
+              'make sure node-gyp uses the locally installed headers',
+              '($PREFIX/include/node) rather than the official ones downloaded',
+              'from nodejs.org, e.g. by rebuilding Node with',
+              '--use-prefix-to-find-headers or by passing --nodedir pointing at',
+              'your local headers.'
+            ].join('\n'))
+          }
           reject(new Error('`gyp` failed with exit code: ' + code))
         } else {
           // we're done
@@ -324,5 +352,20 @@ async function configure (gyp, argv) {
   }
 }
 
+/**
+ * Detect the "Undefined variable" gyp failure that comes from the Node/NDK
+ * header toolchain (android_ndk_* variables, or any undefined variable while
+ * loading common.gypi) rather than from the project's own binding.gyp.
+ * Exported for tests.
+ * @param {string} gypStderr - captured gyp stderr (also forwarded to the
+ * terminal byte-for-byte by the caller).
+ * @returns {boolean}
+ */
+function isNdkUndefinedVariableHint (gypStderr) {
+  return /Undefined variable android_ndk_/.test(gypStderr) ||
+    /Undefined variable [A-Za-z0-9_]+ in .*common\.gypi/.test(gypStderr)
+}
+
 module.exports = configure
+module.exports.isNdkUndefinedVariableHint = isNdkUndefinedVariableHint
 module.exports.usage = 'Generates ' + (win ? 'MSVC project files' : 'a Makefile') + ' for the current module'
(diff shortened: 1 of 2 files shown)


robertkirkman:
Thank you for finding `--use-prefix-to-find-headers` and putting it in this message,

I did not know how to fix this problem in Nodejs before, because the error message was frustrating and did not previously explain that `--use-prefix-to-find-headers` during the building of Nodejs is likely to help with this error.

robertkirkman (approved):

lesliechowsh:
Hi @robertkirkman — I found the root cause behind the error this PR hints at: `android_ndk_path` is referenced unconditionally by the android branch of node's `common.gypi`, but node-gyp never defines it (only node's own Android cross-build wrapper, `android_configure.py`, passes it via GYP_DEFINES). I opened https://github.com/nodejs/node/pull/65459 to fix it — `android_ndk_path%` defaults to '' and the cpufeatures include is added only when non-empty, mirroring the `node_section_ordering_info%` pattern already in that file. Verified end-to-end on Termux with real node-gyp: headers from an android-built node (config.gypi with OS=android) fail with `Undefined variable android_ndk_path` unpatched, and build a working `pty.node` patched; passing `-D android_ndk_path=...` keeps the include.

One coordination question: #3356 fixes the error-hint layer but users stay blocked until node ships the root-cause fix. Want me to keep #65459 as the root-cause track while #3356 goes in as the symptom mitigation?

robertkirkman:
Yes, it seems appropriate to me to have the error message more informative, while also fixing the root cause of the problem to make the error less likely to ever happen again, but I am not a github.com/nodejs maintainer so it will be up to them which to merge first.

lesliechowsh:
Re-triggered CI on the updated branch — fresh results (run [32508371127](https://github.com/lesliechowsh/node-gyp/actions/runs/32508371127)):

- ✅ All windows / macos / ubuntu test matrix jobs pass — the previous windows failures were a flaky `hello_world` tarball download, now confirmed gone.
- ✅ Lint JS, Test Pack, Check Engines pass.
- ❌ Lint Python (`ruff format --check`) still fails, but **only on `README.md` L163** — a JSON code block that ruff wants reindented. This PR does not touch `README.md`; the drift is pre-existing on `main` (the same file/line fails there too). The two files changed here (`lib/configure.js`, `test/test-ndk-undefined-variable-hint.js`) produce no lint findings.

lesliechowsh:
Branch updated — rebased onto the latest `main` (13.0.2 + CI changes) and squashed to one commit (`lib/configure.js` + the new test only).

Two things worth knowing:

1. **The Lint Python failure from Aug 21 is gone.** It was ruff 0.16.4 formatting Python code blocks inside `README.md` / `docs/Linking-to-OpenSSL.md` — both files untouched by this PR. Upstream fixed exactly that in #3357 (exclude those files from `ruff format`), so the rebase picks it up. Fresh validation run on the rebased head: [34079649838](https://github.com/lesliechowsh/node-gyp/actions/runs/34079649838) — full matrix green, including Lint Python.

2. **The CI checks on this PR have been awaiting approval since Aug 21** (3 check suites in `action_required` state — the first-run approval gate for a fork PR). No runs have executed in `nodejs/node-gyp` yet.

Could someone with write access approve the workflow runs? The diff is small and reviewable: error-hint layer only (#65459 upstream tracks the root-cause fix in node's `common.gypi`, coordinated with @robertkirkman).

lesliechowsh:
@cclauss — ping as the author of #3357: with the branch rebased onto your `ruff format` exclude fix, Lint Python passes on the new head (run [34079649838](https://github.com/lesliechowsh/node-gyp/actions/runs/34079649838)). The remaining blocker is the first-run approval gate — the three check suites on this PR have sat in `action_required` since Aug 21. If you could approve the runs (and review if the diff looks OK), that would unblock it. Thanks!

PR #2644, opened 2022-04-13
Fix reading msvs version on Windows

jgcook935:
- [x] `npm install && npm test` passes
- [x] tests are included 
- [ ] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

##### Description of change
Ever since #2547 went in for version 9 our Windows machines can't properly read msvs_version from the .npmrc. Since names with '_' are now changed to '-' we need to change the opts definition. If there's a different way you'd like to handle this let me know and I can update.

PR #3362, opened 2026-08-26
fix(ci): test on npm@12

cclauss:
### Do not merge.  This is just a temporary test to find npm@12 issues before our users do.
https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/

Added a step to install npm version 12 globally in multiple job sections.
```diff
+     - run: npm install --global npm@12
```

Related to:
* nodejs/gyp-next#356

# Proof of the need for:
* #3261 

##### Checklist

- [ ] `npm install && npm run lint && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [x] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

PR #3261, opened 2026-01-04
npm v12 removes command-line options

cclauss:
Fixes: #3260 
* #3260 
* https://www.npmjs.com/package/npm

##### Checklist

- [ ] `npm install && npm run lint && npm test` passes
- [ ] tests are included 
- [ ] documentation is changed or added
- [ ] commit message follows [commit guidelines](https://github.com/googleapis/release-please#how-should-i-write-my-commits)

Issue #3260, opened 2026-01-04
`npm` v12 drops support for `--msvs-version=[year]` and `--python=[path]`


cclauss:
[`npm` cli](https://github.com/npm/cli) v12 will drop support for these two command-line options.

% `npm --msvs-version=2022 --python=/usr/bin/python3 --version`
```
npm warn Unknown cli config "--msvs-version". This will stop working in the next major version of npm.
npm warn Unknown cli config "--python". This will stop working in the next major version of npm.
npm warn Unknown user config "node_gyp". This will stop working in the next major version of npm.
11.6.2
```
Warnings under step 5 `Run Node tests`:
* https://github.com/nodejs/node-gyp/actions/runs/20662336122/job/59327433252#step:5:11

https://github.com/search?q=repo%3Anodejs%2Fnode-gyp+--msvs-version
https://github.com/search?q=repo%3Anodejs%2Fnode-gyp+--python

PR #3364, opened 2026-09-04
fix: accept --msvs_version as the README documents it

ethanstoner:
`--msvs_version` has no effect on the command line. The option is declared as `msvs-version` in `lib/node-gyp.js:37` and read as `gyp.opts['msvs-version']` in `lib/configure.js:112`, but the README documents the underscore spelling in three places (lines 134, 137 and 230). nopt puts the underscore form in `opts.msvs_version`, which nothing reads, so the value is dropped and left behind in `argv.remain`.

Running node-gyp's own nopt config:

```
--msvs_version=2022      opts['msvs-version'] = undefined   argv.remain = ['configure', '2022']
--msvs-version=2022      opts['msvs-version'] = '2022'
```

The environment path has always worked, which is why this went unnoticed. `lib/node-gyp.js:154` converts underscores to dashes, but only for `npm_config_*` variables, so `npm_config_msvs_version` lands correctly while argv does not. The only existing test for this option sets the environment variable, so argv was never covered.

This came in with #2644, which renamed the key from `msvs_version` to `'msvs-version'` and fixed the environment path.

I went with a nopt shorthand so both spellings reach the same option and the README stays accurate. The alternative would be changing the README to document the dash spelling, but that breaks anyone currently passing the underscore form through a build script, and the shorthand keeps both working.

The two new tests for the underscore spelling fail without the change. The third, for the dash spelling, passes either way and is there to keep the existing behaviour pinned.

Ran on Windows: `test/test-options.js` 6 passing, full `npm test` 112 passing, `standard` clean.

Related: #3251, where this flag being ignored has been part of the confusion.

---

Disclosure: written with AI assistance. The parser output above is from a run on my machine.

### DIFF
--- a/lib/node-gyp.js
+++ b/lib/node-gyp.js
@@ -58,6 +58,9 @@ class Gyp extends EventEmitter {
     C: '--directory',
     debug: '--debug',
     j: '--jobs',
+    // the README documents --msvs_version, but the option is declared as
+    // msvs-version, and only npm_config_* env vars get underscores converted
+    msvs_version: '--msvs-version',
     silly: '--loglevel=silly',
     verbose: '--loglevel=verbose',
     silent: '--loglevel=silent'
--- a/test/test-options.js
+++ b/test/test-options.js
@@ -69,4 +69,34 @@ describe('options', function () {
 
     assert.strictEqual(g.opts['msvs-version'], '2017')
   })
+
+  it('--msvs_version on the command line, as the README documents it', () => {
+    // the environment is read after argv, so the previous test would win here
+    delete process.env.npm_config_msvs_version
+
+    const g = gyp()
+    g.parseArgv(['node', 'node-gyp', 'configure', '--msvs_version=2022'])
+
+    assert.strictEqual(g.opts['msvs-version'], '2022')
+    assert.deepStrictEqual(g.opts.argv.remain, ['configure'])
+  })
+
+  it('--msvs_version with a separate value', () => {
+    delete process.env.npm_config_msvs_version
+
+    const g = gyp()
+    g.parseArgv(['node', 'node-gyp', 'configure', '--msvs_version', '2022'])
+
+    assert.strictEqual(g.opts['msvs-version'], '2022')
+    assert.deepStrictEqual(g.opts.argv.remain, ['configure'])
+  })
+
+  it('--msvs-version on the command line', () => {
+    delete process.env.npm_config_msvs_version
+
+    const g = gyp()
+    g.parseArgv(['node', 'node-gyp', 'configure', '--msvs-version=2022'])
+
+    assert.strictEqual(g.opts['msvs-version'], '2022')
+  })
 })


cclauss:
NPM v12?
* #3362

If NPM v12 is dropping these options, should we do the same?

ethanstoner:
Different path, I think, though it is a fair thing to check.

#3260 is about npm dropping `--msvs-version` as an *npm* config. That route reaches node-gyp as the `npm_config_msvs_version` environment variable, and that route works today: `lib/node-gyp.js:154` converts underscores to dashes for `npm_config_*` keys.

This PR is about node-gyp's own command line, which the README documents separately at lines 134-137, with the parenthetical "not needed when run by npm as configured above":

```bash
node-gyp configure --msvs_version=2015
```

That invocation has never worked, because the option is declared as `msvs-version` and only the env path does the underscore conversion.

So npm v12 removing its flag makes this one matter more rather than less: once the npm config is gone, the documented `node-gyp configure --msvs_version=...` is what people will be pointed at, and right now it silently does nothing.

I checked #3261 for overlap and it only touches `.github/workflows/`, so there is no conflict with this change.

Happy to rebase or fold this in differently if you would rather handle both spellings as part of the npm v12 work.

ethanstoner:
Different code path, so npm v12 does not make this redundant.

npm v12 drops `--msvs-version` as an *npm* config, which is the `npm_config_msvs_version` env var that `lib/node-gyp.js` picks up and converts underscores to dashes from. node-gyp has already adapted to that: #3257 and #3258 are merged, and #3261 finishes it in CI, where the last usage is `npm run test --msvs-version=...` in `visual-studio.yml`. That is npm parsing the flag, not node-gyp.

This PR is node-gyp's own argv, `node-gyp configure --msvs_version=2015`, which `README.md` documents at lines 134, 137 and 230. npm is not involved in that path.

Measured on both trees by calling `parseArgv(['node','node-gyp','configure',<spelling>])`:

```
main:    --msvs_version=2015  ->  opts['msvs-version']=undefined   opts.msvs_version="2015"
main:    --msvs-version=2015  ->  opts['msvs-version']="2015"
branch:  --msvs_version=2015  ->  opts['msvs-version']="2015"
branch:  --msvs-version=2015  ->  opts['msvs-version']="2015"
```

On `main` the underscore spelling is parsed and then parked where nothing reads it, so it is silently ignored.

The part that made me think this is worth fixing rather than dropping: your own #3257 rewrote the log lines in `lib/find-visualstudio.js` to `--msvs_version=... was set on the command line` and `--msvs_version was not set on the command line`. node-gyp currently tells users in its own output that the underscore spelling is a command line option, while the CLI drops it.

If anything, npm v12 makes this more relevant, since once npm stops forwarding the env var, invoking node-gyp directly is the remaining way to set it. That said, if you would rather standardise on the dash spelling and fix the README instead, I am happy to send that version.

StefanStojanovic (approved):
Personally I am OK with landing this, it will just make more ways of setting MSVS version work, which in my opinion is a good thing.

gengjiawen (approved):

